Issue #240, opened 2014-12-14
Read methods on nodes and blocks should follow JavaBean conventions


mojavelinux:
The read (aka "getter") methods on the nodes and blocks should follow JavaBean convensions so they can be more easily used from languages such as Groovy. Currently, AbstractBlock is defined as:

``` java
public interface AbstractBlock {
    String id();
    String title();
    String role();
    String style();
    List<AbstractBlock> blocks();
    Map<String, Object> attributes();
    Object content();
    String nodeName();
    String convert();
    DocumentRuby document();
    String context();
    AbstractBlock delegate();
    List<AbstractBlock> findBy(Map<Object, Object> selector);
}
```

The read methods should be defined as:

``` java
public interface AbstractBlock {
    String getId();
    ...
}
```

I would be okay with having both forms, but we should at least have the "getter" methods.

PR #264, opened 2015-01-01
Aligned AST interfaces with Ruby definitions

robertpanzer:
Wile trying to move the reveal.js slim backend to a Groovy MarkupTemplateEngine based approach I needed to make these changes.
The members that are moved from AbstractBlock to AbstractNode are are also defined on AbstractNode in the Ruby implementation.

Some methods like DocumentRuby.doctitle() or AbstractNode.getAttr() were overloaded to mirror the defaults defined in the Ruby implementation.

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/ast/AbstractNode.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/ast/AbstractNode.java
@@ -7,9 +7,21 @@ public interface AbstractNode {
 
     String id();
     String getNodeName();
+    /**
+     * @deprecated Use {@linkplain #getParent()}  instead.
+     */
     AbstractNode parent();
+    AbstractNode getParent();
+    /**
+     * @deprecated Use {@linkplain #getContext()}  instead.
+     */
     String context();
+    String getContext();
+    /**
+     * @deprecated Use {@linkplain #getDocument()}  instead.
+     */
     DocumentRuby document();
+    DocumentRuby getDocument();
     boolean isInline();
     boolean isBlock();
     Map<String, Object> getAttributes();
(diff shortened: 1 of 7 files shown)


asciidoctorj-core/src/main/java/org/asciidoctor/ast/AbstractNode.java
```diff
@@ -17,6 +24,7 @@
     boolean isRole();
     boolean hasRole(String role);
     String getRole();
+    String role();
```
lordofthejars: With Dan we decided to avoid using the short (none Java Beans) form. In fact there is an issue to change everywhere. For this reason there is the getRole() method. The same should be for context, document, ... you can read more at https://github.com/asciidoctor/asciidoctorj/issues/240 We are going to gradually but if new methods can follow this convention it will be less work to be done then. Of course it is totally fine you didn't notice this change yet :)

mojavelinux:
:+1: 

Let's be sure to deprecate the old methods instead of remove so we don't break the API until we are ready to break the API. I realize there will be some gray area, but best to make a smooth transition as possible.

robertpanzer:
I deprecated the methods that complied to the short naming style and added new getters.

mojavelinux:
:+1:

lordofthejars:
It look awesome :) I can't wait until we release the 1.5.2.1 to merge this change to master branch. @mojavelinux we need to take a decision about what to do with logging redirection. For me we can remove the piece of code, release the 1.5.2.1 and then let's think what to do.

mojavelinux:
I think we should comment out the logging redirection for 1.5.2.1 and
stabilize or refine it for inclusion in 1.5.3.

Before we do the 1.5.2.1 release, we study the diff to make sure everything
going into it matches the expectations of a patch release. We can use a
1_5_2 branch if we need to hold other things back. Another option is to
wait for 1.5.3 core to be ready. Perhaps discuss on the list?

lordofthejars:
Yes on list would be better
El dv., 2 de gen., 2015 a les 20.05 Dan Allen <notifications@github.com> va
escriure:

PR #266, opened 2015-01-10
Added new Java counterparts for List and ListItem nodes.

robertpanzer:
Also provided an implementation for the Inline interface.
Additionally moved Ruby to Java node conversion to separate class org.asciidoctor.ast.NodeFactory.

I tried hard for a DynamicProxy based approach and it worked fine for converters.
But the whole approach got mixed up when TreeConverters are added.
It seems to be impossible or at least very hard to mix and match JRuby type conversion and own DynamicProxies.

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/ast/InlineImpl.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/ast/InlineImpl.java
@@ -1,0 +1,33 @@
+package org.asciidoctor.ast;
+
+import org.jruby.Ruby;
+
+public class InlineImpl extends AbstractNodeImpl implements Inline  {
+
+    protected Inline delegate;
+
+    public InlineImpl(Inline delegate, Ruby ruby) {
+        super(delegate, ruby);
+        this.delegate = delegate;
+    }
+
+    @Override
+    public String render() {
+        return delegate.render();
+    }
+
+    @Override
+    public String convert() {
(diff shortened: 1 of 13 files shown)


lordofthejars:
I have only one word to say. Awesome.

robertpanzer:
Thanks!

lordofthejars:
I have merged, today I will update the code with some additions like an spi for converters.

mojavelinux:
:+1:

asciidoctorj-core/src/main/java/org/asciidoctor/ast/InlineImpl.java
```diff
@@ -0,0 +1,33 @@
...
+
+    protected Inline delegate;
+
+    public InlineImpl(Inline delegate, Ruby ruby) {
+        super(delegate, ruby);
+        this.delegate = delegate;
+    }
+
+    @Override
+    public String render() {
```
mojavelinux: Anywhere the term "render" is used, it should be marked as deprecated. "convert" is the preferred terminology.

robertpanzer: Will do so in upcoming PRs :)

lordofthejars: Be aware that last night I updated the branch with new commits, so rebase first :)

Issue #268, opened 2015-01-12
Instruction for WildFly AS should be updated


lbtc-xxx:
Hello, I got into some problem latest asciidoctorj on WildFly. see:

UPDATE: sorry, wrong URL was pasted. 
http://stackoverflow.com/questions/27904558/asciidoctorj-doesnt-work-on-wildfly-8-2-0-final

mgreau:
Hi,
I've created a new github project named asciidoctor-wildfly-docker in order to validate the required configuration in order to run Asciidoctor into WildFly.
This project gives you the possibility to execute tests into a Docker container in order to validate that Asciidoctor can convert AsciiDoc content to HTML and PDF output. 

I'd love to have your feedback :
https://github.com/mgreau/asciidoctor-wildfly-docker

lordofthejars:
Hi that's great and simplifies a lot starting with the application, only one thing, could you add the Dockerfile you used to create the image? In this way everyone can modify it or see what is doing internally the image.

mgreau:
The Dockerfile is here src/test/resources/wildfly/v8.2/Dockerfile . For now, this is the only way to build create this image since it is not on Docker Hub.
One of the main goal is to execute those tests with severals versions of WildFly and others Applications Server like TomEE :)

mojavelinux:
We've got access to the Asciidoctor organization on DockerHub now! I'll setup the admins later today and notify you when it's ready for push.

PR #269, opened 2015-01-12
fix of issue #268

lbtc-xxx:
fix of issue #268 Instruction for WildFly AS should be updated


robertpanzer:
Thanks a lot for your pull request!
"Unfortunately" the issue is already resolved with #277 that already got merged.

PR #273, opened 2015-01-26
introduced asciidoctorj-common module

ancho:
The asciidoctorj-common module bundles common testclasses to share them between other modules.
Eliminates the duplication of `org.asciidoctor.util.ClasspathResources` and `org.asciidoctor.util.ClasspathHelper`

Submodules can add a testCompile dependency:

```groovy
testCompile project(path: ':asciidoctorj-common', configuration: 'tests')
```


mojavelinux:
I like this change, but I'd like to make one suggestion. Could we try to think of a better name than "common" for the module? Before long, the term "common" comes to mean "miscellaneous" or "trash heap" because it lacks focus. I think this module is really intended to support testing, so perhaps we can name it "asciidoctorj-test-support". I'm open to other names as well. wdyt?

mojavelinux:
I'm also okay with asciidoctorj-test-utilities (or asciidoctorj-test-utils).

mojavelinux:
Btw, thanks for proposing this change!

lordofthejars:
Yes I like the name suggested by Dan so we can put there these classes or
for example if we ever create a JUnit runner or something like this, we
could create that classes here.

El Mon Jan 26 2015 at 9:26:44 PM, Dan Allen (<notifications@github.com>) va
escriure:

ancho:
my pleasure!
Well, the naming thing was just a first shot. I like the name 'asciidoctorj-test-support' better.
I'll change that. We'll see how it feels like ;)
Btw. the nebula-publishing plugin carries a plugin called 'nebula-test-jar' which offers the same functionality.
But there is this compatibilitiy issue with JDK 6 and the nebula-publishing plugin.

mojavelinux:
> the nebula-publishing plugin carries a plugin called 'nebula-test-jar' which offers the same functionality.
But there is this compatibilitiy issue with JDK 6 and the nebula-publishing plugin.

Thanks for that clarification! I was wondering about that. Now I know :)

mojavelinux:
Btw, here's another solution where we avoid another module and instead just reuse the test classes from core.

https://softnoise.wordpress.com/2014/09/07/gradle-sub-project-test-dependencies-in-multi-project-builds/

robertpanzer:
Thanks!
We can build on that for further enhancing the tests!

Squashed and merged into 1.6.0.

Unfortunately I now get an error on publishToMavenLocal:

```
Execution failed for task ':asciidoctorj-test-support:publishMavenNebulaPublicationToMavenLocal'.
> Failed to publish publication 'mavenNebula' to repository 'MavenLocal'
   > Invalid publication 'mavenNebula': artifact file does not exist: 'C:\work\git\asciidoctorj\asciidoctorj-test-support\build\libs\asciidoctorj-test-support-1.5.3-SNAPSHOT.jar'
```

@mojavelinux @ancho Do you have an idea how to exclude a subproject from publishing? Tried different things but did not find a way.

robertpanzer:
Guess I found out how to do it in the root build.gradle: 
```
  if (JavaVersion.current().isJava7Compatible() && !it.name.endsWith('-test-support')) {
    apply from: rootProject.file('gradle/publish.gradle')
  }
```

Is this a reasonable way to do it?

mojavelinux:
That's the only way I know how atm.

robertpanzer:
Thanks!
Committed it like this:
https://github.com/asciidoctor/asciidoctorj/commit/db7138ef28d439ff9d14eb842df8667d3319807d

robertpanzer:
I have a small problem with the current configuration of the test-support project.
I am currently working on an Arquillian integration with Asciidoctor.
Therefore I'd like to add an Arquillian extension to the test-support subproject.
Other projects should use it in their tests and pull in transitive dependencies on Arquillian. 
But it seems that dependencies are not populated to depending projects as the sources are currently test sources and test dependencies are usually not exported.

@ancho @mojavelinux What do you think about that? Would you mind if I move the sources from test to main? The archive will not be published anyway.

mojavelinux:
Yep, that's fine.

I just spoke to Aslak about the organization of code. He had suggested that you could put the Arquillian extension for the Asciidoctor test suite into core/src/test instead of its own module. Then, the other modules could depend on the test-jar from core. That might simply things. Another option is to simply make test-support a test-jar and then explicitly depend on the test-jar (using a dependency qualifier). I've never done it with Gradle, but I understand it in theory.

ancho:
Hi. That's very interesting. I guess with this scenario it is better to bundle a test-jar and use it in the modules. I think it is a better separation of concerns in this case.
Mr.Haki wrote about bundling a test-jar  http://mrhaki.blogspot.de/2010/11/gradle-goodness-create-jar-artifact.html

mojavelinux:
Aha, there we go. Thanks for the link, @ancho!

ancho:
@robertpanzer do you have an example of what exactly you want to do? the link i posted is a bit outdated. Using {{sourceSets.test.classes}} doesn't work. And now I remember, that's exactly what i'm doing. It is now called {{sourceSets.test.output}} ( A sourceSet called 'test' added from the Java Plugin).

See  http://gradle.org/docs/current/userguide/java_plugin.html#N11F7B for more Information.

Sorry about that. It's been a while.

robertpanzer:
Thank you for the link!
My current problem is that I have a compile (or testCompile, tried both) dependency on arquillian-container-jruby in test-support.
The "real" projects core, pdf, diagram etc have a dependency on arquillian-junit-container and the test-support project.
Unfortunately the dependency on arquillian-container-jruby doesn't promote to the core, pdf and diagram projects.
Classes itself are perfectly visible, that's no problem.

Sorry for me not being able to share the current state on github but I am in the midst of a rebase session :-/

If test-support is a "normal" project like all other that simply is not published I don't have these problems.
Then the normal dependency management works.

Maybe someone has an idea how this could be handled.

ancho:
If I get you right, you need arquillian dependencies for testing in each submodule. Exactly as you need junit and hamcrest for testing.

Add the dependencie you need to the subproject. And get specific in the concrete submodule.

See https://github.com/asciidoctor/asciidoctorj/blob/master/build.gradle#L88-L93

robertpanzer:
Yes, maybe that's the way to go. 
Even though it would have been great to have transitive dependencies.

robertpanzer:
What I wanted to say was:

Yes, you're right 
That's the way to go. 
It allows to include these deps without unnecessary repetition. 
Thank you!!!!

ancho:
Ah. I think I finally understand. (At least I hope so)
Maybe http://gradle.org/docs/current/userguide/dependency_management.html#sub:client_module_dependencies is what you are looking for.

robertpanzer:
Now I finally arrived where I wanted to go:
https://github.com/robertpanzer/asciidoctorj/blob/ArquillianJrubyContainer/asciidoctorj-test-support%2Fbuild.gradle#L10

If the tests configuration extends testCompile all dependencies are propagated to the depending projects.
Thank you for not losing your patience with me! :)

ancho:
Nice. Good to know. Thank you too :)

mojavelinux:
Excellent!

PR #276, opened 2015-02-19
Introduce extension proxies between Ruby and Java extensions 

robertpanzer:
This PR is probably already much too large, sorry :(

Introduced a proxy class for every extension type.
Made a larger change to the InlineMacroProcessor so that RubyRegexps are no longer created in the Java Processor but in the Ruby base class and transported back to the Java extension config.


lordofthejars:
Hi Robert, awesome contribution, only one thing, the build is failing when the project is trying to compile against a JDK prior to Java 8. I don't know exactly why this is happening:

```
error: method setConfig in class Processor cannot be applied to given types;
        getProcessor().setConfig(RubyHashUtil.convertRubyHashMapToMap(newConfig));
                      ^
  required: Map<String,Object>
  found: Map<Object,Object>
  reason: actual argument Map<Object,Object> cannot be converted to Map<String,Object> by method invocation conversion
Note: Some input files use or override a deprecated API.
```

Can you try to fix this I think it is pretty easy, and I could do it without any problem but I said that to maintain root of the branch runnable.

But this work is awesome. My next step will be the refactor to DocumentRuby inerface to Document. But I will open a new issue and I will do it :).

robertpanzer:
Sorry Alex! It built successfully yesterday before I checked in but I did not do a final clean build.

lordofthejars:
Don't worry it is not a problem, for this reason we have the CI environment
:D. No worries fix them when you have time we are not out of date with
AsciidoctorJ 1.6.0 so take your time and thank you so much for your help.

BTW I have added as a team member of AsciidoctorJ, when you can you can
update pom.xml as well in the developers section :)

El Fri Feb 20 2015 at 3:18:42 PM, Robert Panzer (<notifications@github.com>)
va escriure:

robertpanzer:
Thanks! Hope it's ok to do it as part of this PR.

lordofthejars:
Yes if you want we Can use same PR
El dv., 20 de febr., 2015 a les 16.20 Robert Panzer <
notifications@github.com> va escriure:

PR #277, opened 2015-02-20
Improve documentation about usage under WildFly.

IanDarwin:


mojavelinux:
Many thanks for this fix! @lordofthejars or @mgreau, would you like to review?

mgreau:
Yes I will do it this week-end

lordofthejars:
Yes I expect to do this weekend as well
El dv., 20 de febr., 2015 a les 22.00 Maxime Gréau <notifications@github.com>
va escriure:

mgreau:
Thanks @IanDarwin, it's OK for me (only few typos : Wildfly -> WildFly and AsciiDoctorJ -> AsciidoctorJ)

BTW, @mojavelinux and @lordofthejars, do you think that the [asciidoctor-wildfly-docker](https://github.com/mgreau/asciidoctor-wildfly-docker)  project has its place in the organization (maybe with another name) ? 
So that we can improve this documentation about AsciidoctorJ/WildFly and : 
- give to the community an easy way to run a container instead of doing those steps 
- give us a way to check that future versions of asciidoctorj (PDF, EPUB3, diagram...) will be compatible or not with future versions of WildFly

mojavelinux:
Thanks for reviewing, @mgreau!

I definitely agree the docker image fits in the organization. My only concern is that we choose a naming pattern so it's consistent with other docker images we make. So far, we have docker-asciidoctor. Perhaps this one can be docker-asciidoctorj-wildfly (or just docker-asciidoctorj). Sound good?

mgreau:
yes, docker-asciidoctorj sounds good.
I forget to mention that in this project, there are Arquillian tests that used the Dockerfile in order to run tests into a container (thanks to Arquillian Cube :) ). 
WDYT about those tests ? Does the docker-xxx naming pattern is for project with only Docker images ?

IanDarwin:
On 2015-02-22 17:32, Maxime Gréau wrote:
> Thanks @IanDarwin <https://github.com/IanDarwin>, it's OK for me (only 
> few typos : Wildfly -> WildFly and AsciiDoctorJ -> AsciidoctorJ)
Fixed those and pushed to my branch, thanks!

mojavelinux:
> Does the docker-xxx naming pattern is for project with only Docker images?

I leave that decision up to you. I'm happy to have tests included.

mgreau:
> I leave that decision up to you. I'm happy to have tests included.

Ok me too, I will create it with tests included.
Do you think that it's time to create an organization for Asciidoctor on Docker Hub ? Then, We could have a naming pattern for those images.

mojavelinux:
> Do you think that it's time to create an organization for Asciidoctor on Docker Hub ?

Absolutely! Would you like to manage it? I can setup an email alias if we need a shared e-mail address.

mgreau:
Yes, I'd love to manage it :)

mojavelinux:
Nice. Now we just need to get access to it. If DockerHub turns down our request, we'll just use a modified name (such as asciidoctor-project or something). I'll post as soon as I hear from them so we can proceed.

Issue #284, opened 2015-03-21
Rendering a file fails if it is located at the root of the filesystem, e.g. /foo.adoc


robertpanzer:
When a file should be rendered given an absolute path in the root of the filesystem rendering fails with this exception:

```
java.lang.IllegalArgumentException: Directory does not exist: 
	at org.asciidoctor.GlobDirectoryWalker.checkInput(GlobDirectoryWalker.java:97)
	at org.asciidoctor.GlobDirectoryWalker.<init>(GlobDirectoryWalker.java:36)
	at 
```

This is due to an error in my latest PR to allow rendering files defined via absolute paths.
The problem is a classical off-by-1 problem: The last '/' of the path is omitted and therefore the path is empty in case of the root directory.

robertpanzer:
Fixed on master with https://github.com/asciidoctor/asciidoctorj/commit/e101b4bcfba922050b8bd38348d6510cea7f432c
Fixed on asciidoctorj-1.6.0 with https://github.com/asciidoctor/asciidoctorj/commit/bb5c645050ddd2d465b54e1364b5664e4002d89d

PR #285, opened 2015-03-21
#284 Don't fail for files in root directory with absolute path

robertpanzer:
Fix for #284 

@lordofthejars If you accept this PR I would cherry-pick the commit into the master branch.

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/GlobDirectoryWalker.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/GlobDirectoryWalker.java
@@ -26,31 +26,42 @@ public class GlobDirectoryWalker implements DirectoryWalker {
 
     public GlobDirectoryWalker(String globExpression) {
 
-        if (isAbsoluteGlobExpression(globExpression)) {
+        // Determine leading part that is unglobbed, i.e. does not contain a '*'.
+        int indexOfUnglobbedPart = findIndexOfUnglobbedPart(globExpression);
 
-            // If the given path is an absolute path we want to split it into an absolute part containing
-            // directories without wildcards and a glob part with wildcards and file names
-            int indexOfUnglobbedPart = findIndexOfUnglobbedPart(globExpression);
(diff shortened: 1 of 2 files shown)


robertpanzer:
Will check for the test error on windows next week when I have a windows box again at reach

abelsromero:
I cannot send a PR right now, but I tested this on my machine and I'd like to add what I found.

I think the problem is in how Linux absolute paths are cheched in (our friend) Windows here:
https://github.com/robertpanzer/asciidoctorj/blob/d7ce297cab1f2101e4a1216434b3c5d4351ce386/asciidoctorj-core/src/main/java/org/asciidoctor/GlobDirectoryWalker.java#L79
Since `File.separator` returns `\` in Windows (L75), and the path does start with a letter, the path in the test `/this_file_does_not_exist.adoc` is not recognized as an absolute path, and that's why it fails. 
Also, I fear this way to check for paths does not fix all possible options in Windows. That's because java.io.File accepts unix paths also on windows, when something like '/folder/another_folder/file.adoc' is passed, it is translated to 'c:\folder\another_folder\file.adoc'. Basically it gets the default unit letter from windows registry and changes the slashes.

In order to better test this I've been thinking that maybe it makes sense to move `isAbsoluteGlobExpression` into `org.asciidoctor.internal.IOUtils` and call it something like 'isAbsolutePath'. That way we can write tests to validate all possible options and decouple `GlobDirectoryWalker` class from specifc OS treatments (in that case findIndexOfUnglobbedPart
and isWindows should be moved too).

wdyt?

robertpanzer:
:+1: from my side

I am a little embarrassed that I forgot about the edge case of a file in the root. 
I think your proposal is a really good idea!

robertpanzer:
@abelsromero I think I got it right now and it is even simpler than I thought.

Travis failed on one test because the TCP connection was reset while fetching dependencies, so I suppose that everything is working fine now.

@abelsromero @lordofthejars Are you both ok with the way it is resolved now?

PR #287, opened 2015-04-13
Removed redundant config map from extensions.

robertpanzer:
Currently the extension config is held redundantly in the Ruby part and in the Java part of an extension.
This leads to problems that config properties set in the Ruby part do not appear in the Java part and vice versa.
You can see the problem if you change the regular expression in the current test for an inline macro processor so that the extension should _not_ be applied. 
But it applies nevertheless because the regular expression does not translate into the Ruby part and therefore only the name is used.

This PR invents a RubyHashMapDecorator that wraps a RubyHash and decorates a `Map<String, Object>` around it, so that the `@config` member of the Ruby extension is also used in the Java part only with another view on it.

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/extension/Processor.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/extension/Processor.java
@@ -10,6 +10,7 @@ import org.jruby.RubyHash;
 import org.jruby.javasupport.JavaEmbedUtils;
 import org.jruby.runtime.builtin.IRubyObject;
 
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
@@ -36,6 +37,10 @@ public class Processor {
         return null;
     }
 
+    public Block createBlock(AbstractBlock parent, String context, String content, Map<String, Object> attributes) {
+        return createBlock(parent, context, content, attributes, new HashMap<Object, Object>());
+    }
+
     public Block createBlock(AbstractBlock parent, String context, String content, Map<String, Object> attributes,
             Map<Object, Object> options) {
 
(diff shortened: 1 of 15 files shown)


asciidoctorj-core/src/main/java/org/asciidoctor/extension/Processor.java
```diff
@@ -36,6 +37,10 @@ public RubyHash setConfig(Map<String, Object> config) {
         return null;
     }
 
+    public Block createBlock(AbstractBlock parent, String context, String content, Map<String, Object> attributes) {
```
robertpanzer: Overloaded this method to not need options as the parameter is optional in the Ruby counterpart

asciidoctorj-core/src/test/java/org/asciidoctor/extension/GistMacro.java
```diff
@@ -19,7 +19,7 @@ public Block process(AbstractBlock parent, String target, Map<String, Object> at
        		"<script src=\"https://gist.github.com/"+target+".js\"></script>\n" + 
        		"</div>"; 
        
-       return createBlock(parent, "pass", Arrays.asList(content), attributes, this.getConfig());
```
robertpanzer: This only worked because the default config property :content_model => :attributes was not visible in the Java part.

lordofthejars:
It looks good to me. I like the idea of having all parameters in single place (in this case in Ruby) instead of having two different representations of same data which can diverge as you mention.

robertpanzer:
Thanks! I'd like to see what @mojavelinux thinks of it before we merge it.

mojavelinux:
I like the idea of being able to link the map so that we are manipulating the same instance. That was definitely a source of confusion and I think by using a single, canonical source, we should be able to avoid users from getting into strange situations.

I also like the addition of Spock :)

robertpanzer:
Glad you like it! I also like that approach more than intercepting the accessors and changing Asciidoctor to not using direct member access, despite the code of the proxies becoming a bit reflection-like. 

But there is still one "glitch" though, because there is no constructor chaining from Java into Ruby. But IMO it's an improvement.

PR #294, opened 2015-05-05
Adding a section for using AsciidoctorJ in an OSGi context in the README.

twasyl:
I created a section regarding the use of AsciidoctorJ in an OSGi environment.

I placed it after the *GEM_PATH* section mainly because the OSGi context is evoked in it, so I thought it might be the right place to be. wdyt?

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1550,6 +1550,48 @@ Asciidoctor asciidoctor = create("/my/gem/path"); // <1>
 ----
 <1> Creates an `Asciidoctor` instance with given `GEM_PATH` location.
 
+== Using AsciidoctorJ in an OSGi environment
+
+In a non OSGi context, the following snippet will successfully create an Asciidoctor object:
+
+[source,java]
+----
+import static org.asciidoctor.Asciidoctor.Factory.create;
+import org.asciidoctor.Asciidoctor;
+
+Asciidoctor asciidoctor = create();
+----
+
+In an OSGi context it will not work because JRuby needs some paths to find the gems (the Asciidoctor ones and the Ruby themselves ones). In order to make it work, you will need two more classes (RubyInstanceConfig and JavaEmbedUtils) and a small modification of the previous snippet of code. The modifications take care of the class loaders because in OSGi, which are a key point in OSGi:
+
+[source,java]
+----
+import static org.asciidoctor.Asciidoctor.Factory.create;
+import org.asciidoctor.Asciidoctor;
+
+RubyInstanceConfig config = new RubyInstanceConfig();
+config.setLoader(this.getClass().getClassLoader()); <1>
+
+JavaEmbedUtils.initialize(Arrays.asList("META-INF/jruby.home/lib/ruby/2.0", "gems/asciidoctor-1.5.2/lib"), config); <2><3><4>
+
+Asciidoctor asciidoctor = create(this.getClass().getClassLoader()); <5>
+----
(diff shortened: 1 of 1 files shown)


lefou:
Looks good to me.

The hard coded versions of JRuby and AsciidoctorJ are best targeted with a proper AsciidoctorService provided to the OSGi registry, which is of course not the scope of this PR.

twasyl:
Can you detail a little bit more the AsciidoctorService point in the OSGi registry? I didn't know about this and it may be good to point it in the documentation. Wdyt?

lordofthejars:
Yes I think that we can add AsciidoctorService example too in the documentation so anyone who wants to use AsciidoctorJ in OSGi it gets the same solution.

lefou:
IMHO, instead of writing such a service in the documentation, we should add it to the sources/jars. The documentation should cover it's usage.

Currently, the bundle using the Asciidoctor API needs to set it up including knowing all the details. If more than one bundle want to do this, we end up having lots of Asciidoctor (and JRuby) instances in the same JVM, which is not so good. The proper OSGi way would be to register a service into the OSGi registry that covers the Asciidoctor API. Consumers do not need to setup Asciidoctor themselves but instead consume the service from the OSGi registry. Provided that such a service exists, classloading, versioning, bootstrapping and such things would no longer be an issue.

What's needed for this:
* Decide the service API
  * do we want to expose the Asciidoctor class directly?
  * Better would be to have an interface which is not directly coupled to the implementation which means we would need to introduce another JAR/bundle
* Write an OSGi Activator which registers the service into the service registry
* Write tests and documentation how to consume the service
* Think about a proper mechanism for extension, e.g. the [Whiteboard Pattern](http://wiki.osgi.org/wiki/Whiteboard_Pattern) comes to mind.

That's of course some work. That's why I said, it might not be the scope of the PR. ;-)

twasyl:
I totally agree with your approach. I thought this was already present somewhere bu I didn't find it.

In fact I've already done some stuff [SlideshowFX](https://github.com/twasyl/SlideshowFX) which bascially uses the Whiteboard Pattern: each extension has a common interface to define what an extension can do and must provide.

But as you said, that might not be part of this PR :)

lordofthejars:
Ok then I accept the PR :).

lefou:
@twasyl Where do the OSGi bundles for AsciidoctorJ and co come from? (Are you on Gitter.im too? If yes, let's continue there.)

mojavelinux:
> IMHO, instead of writing such a service in the documentation, we should add it to the sources/jars. The documentation should cover it's usage.

Absolutely!

mojavelinux:
@lordofthejars, now that this is merged, could you update the following sentence in the preceding section as it is now a bit misleading?

> But in some circumstances like OSGi environments you may require to store gems in an external folder and be loaded by AsciidoctorJ.

This should instead say

> In certain circumstances, you may want to load gems from an external folder.

Just keep it generic and don't send people using OSGi down the wrong path.

mojavelinux:
@lefou Could you capture the requirements you outlined in a new issue so that they don't get lost?

lordofthejars:
Sure I update it no problem. Tomorrow morning is updated.
El dt., 5 de maig, 2015 a les 22.06 Dan Allen <notifications@github.com> va
escriure:

lefou:
@mojavelinux I opened #297 and added the OSGi requirements.

mojavelinux:
Thanks @lefou!

PR #295, opened 2015-05-05
Fixes #288. And added some additional Javadoc to other methods.

### LINKED ISSUE #288: Attributes.setCacheUri looks like a cut-and-paste error
Somehow I don't think this is intended, though maybe it doesn't matter.

 public void setCacheUri(boolean hardbreaks) {
        this.attributes.put(CACHE_URI, toAsciidoctorFlag(hardbreaks));
    }

robertpanzer:
If this gets merged I would cherry-pick the commit into the asciidoctorj-1.6.0 branch.

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/Attributes.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/Attributes.java
@@ -78,18 +78,68 @@ public class Attributes {
         this.setAttributes(attributes);
     }
 
+    /**
+     * Allow Asciidoctor to read content from an URI.
+     * Additionally the option {@link SafeMode} must be less than {@link SafeMode#SECURE} to enable inclusion of content from an URI.
+     * @see <a href="http://asciidoctor.org/docs/user-manual/#include-content-from-a-uri">Asciidoctor User Manual</a>
+     * @param allowUriRead {@code true} to enable inclusion of content from an URI
+     */
     public void setAllowUriRead(boolean allowUriRead) {
         this.attributes.put(ALLOW_URI_READ, toAsciidoctorFlag(allowUriRead));
     }
 
+    /**
+     * Define how to handle missing attribute references.
+     * Possible values are:
+     * <table>
+     *     <tr>
+     *         <td>{@code skip}</td>
+     *         <td>leave the reference in place (default setting)</td>
+     *     </tr>
+     *     <tr>
+     *         <td>{@code drop}</td>
+     *         <td>drop the reference, but not the line</td>
+     *     </tr>
+     *     <tr>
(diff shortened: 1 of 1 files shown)


lordofthejars:
Hi Robert there is an issue about converting javadoc to asciidoc and rendered with asciidoclet. If you want I accept the PR and we will change in future. WDYT?

robertpanzer:
Oh, I was not aware of that.
I'd like to fix that beforehand.
How can I reproduce that?

lordofthejars:
well the idea is add the asciidoclet plugin on pom.xml and write javadoc with asciidoc format.

mojavelinux:
> the idea is add the asciidoclet plugin on pom.xml and write javadoc with asciidoc format.

And we should depend on a stable version of Asciidoclet. It's okay of Asciidoclet is behind the version of AsciidoctorJ. We want to treat it like a third-party library just like any other project would do. The fact that Asciidoclet uses AsciidoctorJ under the covers should be inconsequential.

:+1: to moving to Asciidoclet. Eat your own dog food :)

mojavelinux:
Of course, we can proceed with this update and switch to Asciidoclet in a seperate issue. See #35.

robertpanzer:
@lordofthejars I probably did not understand you correctly, I thought that there already was a pom.xml that uses asciidoclet.
But I think it is planned in the future to use asciidoclet, isn't it? 
Because I could not find a pom.xml.

If this was your intention I would prefer doing this with a separate issue and first committing this, because probably other classes might also need changes when transitioning to asciidoclet. (For example GlobDirectoryWalker where I added some HTML javadoc just recently ;-) )

mojavelinux:
> Because I could not find a pom.xml.

I think he meant build.gradle :)

robertpanzer:
BTW when adding missing javadoc I stumbled over this method:

public void setSectNumLevels(int sectNumLevels)

I think I can guess what it means (The number of sections that are numbered like 2 to give numbers to level 0 and 1 sections) but I could not find anything that confirms my guesses in the user manual.

lordofthejars:
Sorry yes I mean gradle file. Yes I think the best way is to accept the PR
and change in future. Sorry :)
El dt., 5 de maig, 2015 a les 22.29 Robert Panzer <notifications@github.com>
va escriure:

mojavelinux:
> I think I can guess what it means (The number of sections that are numbered

Yes, the depth to which the numbering goes. I've opened an issue to document this in the user manual. See https://github.com/asciidoctor/asciidoctor.org/issues/383. Thanks for the heads up!

jaredmorgs:
Claimed (can't assign to me).

Issue #267, opened 2015-01-12
Asciidoctor rendering fails when using Java extensions


pmagon:
It seems that there is still a problem (I tried with 1.5.2...)
With TerminalCommandTreeprocessor:

    Hello World

    $ echo "Hello, World!"

    = Title

    $ gem install asciidoctor

By using asciidoctor.convert I get

    Exception in thread "main" java.lang.ClassCastException: org.asciidoctor.ast.SectionImpl cannot be    cast to org.jruby.RubyObject
    at org.asciidoctor.ast.AbstractBlockImpl.blocks(AbstractBlockImpl.java:52)
    at com.imtf.test.asciidoc.TestAsciiDoc.showDoc(TestAsciiDoc.java:301)
    at com.imtf.test.asciidoc.TestAsciiDoc.test6(TestAsciiDoc.java:343)
    at com.imtf.test.asciidoc.TestAsciiDoc.main(TestAsciiDoc.java:362)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:39)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:25)
    at java.lang.reflect.Method.invoke(Method.java:597)
    at com.intellij.rt.execution.application.AppMain.main(AppMain.java:134)

I also have similar errors with the "gistMacro" and "ManpageMacro" examples:

    = Hello
    Test

    == Sect 2

    gist::1234[]
 
Result:

    Exception in thread "main" java.lang.ClassCastException: org.jruby.gen.InterfaceImpl1503916182 cannot be cast to org.jruby.RubyObject
	at org.asciidoctor.ast.AbstractBlockImpl.blocks(AbstractBlockImpl.java:52)
	at com.imtf.test.asciidoc.TestAsciiDoc.showDoc(TestAsciiDoc.java:301)
	at com.imtf.test.asciidoc.TestAsciiDoc.test6(TestAsciiDoc.java:343)
	at com.imtf.test.asciidoc.TestAsciiDoc.main(TestAsciiDoc.java:362)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:39)
[...]

PR #298, opened 2015-05-06
Test that TreeProcessor and BlockMacroProcessor can be used together

robertpanzer:
In the discussion around #267 @pmagon mentioned that he was not able to get a TreeProcessor running together with a BlockMacroProcessor.
This PR adds a test that proves that this configuration works.

Additionally the existing TerminalCommandTreeprocessor was changed to also convert terminal commands that are nested in sections.


mojavelinux:
Tests FTW! This is definitely the way to address this concern and ensure it keeps working. Nice job.

PR #300, opened 2015-05-07
Fixes #299: Removed rerouting of System.out and System.err to JUL

### LINKED ISSUE #299: StackOverflowError when creating asciidoctor object
In  asciidoctorj-1.6.0: 

    System.out.println ("CREATE");
    Asciidoctor asciidoctor = JRubyAsciidoctor.create ();
    System.out.println ("OK");

The second "System.out.println" breaks.... It seems that "System.out" has been changed on "create"
If I save and restore "System.out" after the call I get the same error on "asciidoctor.convertFiles"...

Robert Panzer said:

> I think the StackOverflow comes from here:
>https://github.com/asciidoctor/asciidoctorj/blob/asciidoctorj-1.6.0/asciidoctorj-core%2Fsrc%2Fmain%2Fjava%2Forg%2Fasciidoctor%2Finternal%2FJRubyAsciidoctor.java#L40

>This is what we already discussed about.
>It does not occur in the gradle tests because Gradle injects its own JUL Logger which breaks the cycle between catching System.out and forwarding it to the default ConsoleHandler which again does a System.out.

robertpanzer:
The current solution to capture output from Asciidoctor and route it to a Java logger leads to infinite recursions and therefore StackOverflows when a ConsoleHandler is configured, which is the default in most configurations.
This problem is not reproducable with gradle as it replaces the normal System.out and System.err fields with own streams.
As Dan pointed out in #299 Asciidoctor will provide another solution that allows for capturing the output without having to read from System.out.
Therefore this PR completely removes capturing System.out and .err without replacement.

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/internal/JRubyAsciidoctor.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/internal/JRubyAsciidoctor.java
@@ -33,27 +33,6 @@ import java.util.logging.Logger;
 
 public class JRubyAsciidoctor implements Asciidoctor {
 
-    static {
-
-        final Logger logger = Logger.getLogger("Asciidoctor");
-        final LoggerOutputStream out = new LoggerOutputStream(logger);
-        System.setOut(new PrintStream(out));
-        System.setErr(new PrintStream(out));
-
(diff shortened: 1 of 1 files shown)


mojavelinux:
:+1:

Issue #296, opened 2015-05-06
Allow JRuby mode to be specified


mojavelinux:
Right now we are defaulting the Ruby mode to 2.0, but not providing anyway to change this (that I can tell). It seems we are overridding whatever the user is setting. Expose this as an option, or honor the value set in `JRUBY_OPTS` or `-Djruby.compat.version` so that there is some way to override this value. This is particularly important when working with the Maven plugin.

We've encountered an issue whereby setting the Ruby mode to 2.0 has caused remote includes to note work. See https://github.com/asciidoctor/asciidoctor-maven-plugin/issues/153

mojavelinux:
If it is possible to set the JRuby mode, then this needs to be documented in the manual (aka README).

mojavelinux:
I now believe that we should not be specifying the Ruby mode by default. (Experience has led me to change my opinion on this matter). We allow different versions of JRuby to be used, so by setting the Ruby mode, we cause issues rather than solving them. When JRuby 9.0.0.0 is used, for instance, we are downgrading the Ruby mode from 2.1 to 2.0.

I think the Ruby mode should be purely a user setting. We should allow it to be set however it is normally set in JRuby, plus perhaps we can have an option to the API as well. Most important, we should document how it works.

robertpanzer:
Do you think it would work out to add a parameter `Map<String, String> runtimeConfig` to the  `Asciidoctor.Factory.create()` methods that contain properties to customize the Ruby config?

So an Asciidoctor instance with Ruby 1.9 could be created like this:

```java
List<String> loadPaths = ...
Map<String, String> rubyRuntimeProps = new HashMap<>();
rubyRuntimeProps.put("compatVersion", "1.9");
Asciidoctor asciidoctor = Asciidoctor.Factory.create(loadPaths, rubyRuntimeProps);
```

JRubyAsciidoctor would be in charge of recognizing these parameters and handle them appropriately.

The downside would be that we would reach 10 create methods in Asciidoctor.Factory.

The advantage is that it would keep the door open for supporting other runtimes, that surely will need some kind of configuration as well.

mojavelinux:
This is a great question, but again I'm torn. It's probably very unlikely that you'd need to configure JRuby differently per Asciidoctor instance. On the other hand, if you are using different implementations, then you might find yourself using this configuration more often.

[...]

PR #301, opened 2015-05-10
Remove JRuby runtime context

robertpanzer:
The current code uses one shared Ruby instance in the extensions that can be different from the Ruby runtime where Asciidoctor Ruby runs in.
This might lead to problems.

Please review in particular the changes to the Processor class extensively.
I changed the signature from the config methods from Map<Object, Object> to Map<String, Object>.
IMO that makes sense because a consumer of AsciidoctorJ should not have to hassle with RubySymbols etc.

When a Processor creates a new node, the Ruby runtime is taken from the given parent node. 
I think this is a valid assumption.

I only have the feeling that I am now shooting myself in the foot if we want to factor out an API module.
There have to be some modifications done to the Processor class.
(This only applies to the setConfig() method, adapting the createBlock() methods is quite simple via a service implementation.)


asciidoctorj-core/src/main/java/org/asciidoctor/extension/Processor.java
```diff
@@ -16,25 +17,29 @@
...
     	this.config.putAll(config);
     }
     
-    public Map<Object, Object> getConfig() {
+    public Map<String, Object> getConfig() {
     	return this.config;
     }
 
-    public RubyHash setConfig(Map<String, Object> config) {
-        this.config = RubyHashUtil.convertMapToRubyHashWithSymbols(rubyRuntime, config);
```
robertpanzer: This was still a gap.
If someone sets the config after the ProcessorProxy has connected to this Proxy, the config must not be simply replaced because this would bring the config maps out of sync between Java and Ruby.
I wonder if it makes sense at all to change the config after an extension was is registered.
In particular as for example the InlineMacroProcessor does further actions depending on the config in the initializer.

asciidoctorj-core/src/main/java/org/asciidoctor/extension/Processor.java
```diff
@@ -16,25 +17,29 @@
...
-        return null;
+    public void setConfig(Map<String, Object> config) {
+        if (config instanceof RubyHashMapDecorator) {
+            // ProcessorProxy has already taken over, replace config
+            this.config.clear();
+            this.config.putAll(config);
+        } else {
+            // Processor is still independent, simply set the member
+            this.config = config;
+        }
```
robertpanzer: It will be hard to make this method independent of Ruby if we really decide to factor out an API module.
Any ideas are welcome.

robertpanzer:
These changes will result in some nice merge conflicts with #306 
If we agree on this change I will have to rebase before merging

robertpanzer:
Currently the readme says:

> Sometimes you may need the Ruby runtime used inside AsciidoctorJ. One reason is because you are using JRuby outside AsciidoctorJ and you want to reuse the same instance. Another reason is that you need to instantiate by yourself an Asciidoctor Ruby object.
>
> To get this instance you can use org.asciidoctor.internal.JRubyRuntimeContext.get() to get it.

This is certainly no longer true with this change.
But in certain cases, when you have multiple Asciidoctor instances, it is also not 100% true at the moment.

What do you think of providing access to the Ruby runtime of an Asciidoctor instance like this?

```java
Ruby rubyRuntime = JRubyRuntimeContext.getRubyRuntime(asciidoctor);
```

Maybe we would need something similar to create an Asciidoctor instance inside a given Ruby runtime!?

```java
Ruby rubyRuntime = ...
Asciidoctor asciidoctor = JRubyRuntimeContext.create(rubyRuntime);
```

lordofthejars:
Hi Robert, I like the ideas you are giving here. About creating an
Asciidoctor instance with given Ruby instance (second snippet) I think that
it should be implemented as a method in the Factory implemented in
 Asciidoctor interface.

WDYT?


El ds., 16 maig 2015 a les 18:56, Robert Panzer (<notifications@github.com>)
va escriure:

Currently the readme says:

robertpanzer:
Hi Alex, 

I also thought about this and it would also work well.
I am just unsure about adding any dependencies on the underlying implementation to this interface.
Because at the moment Asciidoctor and Asciidoctor.Factory are (nearly) completely free of dependencies on Ruby.

If we want to add other implementations that also offer the same API to Java we might also need a method to create an Asciidoctor instance in a given Nashorn runtime.
But for all these decisions the API user is already aware of the implementation she or he is using so she or he could also use a different implementation specific API to create the Asciidoctor implementation.

So that's what led me to that proposal.

But I am also open to adding a corresponding create() method in Asciidoctor.Factory.
(And in the end sth like a gem load path is already implementation specific;-) )

lordofthejars:
Yes I understand your thoughts but I think that when we add Nashorn
support, all this Factory code will be refactored for sure. Fir this reason
I thought about doing in this way but of course, the way you mention it can
have much more sense in the future. So I will say implement as you said so
less work to be done in future :)

El ds., 16 maig 2015 a les 23:35, Robert Panzer (<notifications@github.com>)
va escriure:

robertpanzer:
Another way to stay with Asciidoctor.Factory for creating instances could be like this:

We add an interface `RuntimeOptions` and a method `Asciidoctor.Factory.create(RuntimeOptions options)`. 

The JRuby impl would bring an implementation of this interface so you could say:

```java
Ruby rubyRuntime = ...
Asciidoctor.Factory.create(JRubyOptions.options().runtime(rubyRuntime))
```

Another impl could simply bring a class `NashornOptions` and that could also be a way to choose an impl if you have multiple impls on the classpath. 

RWhat do you think about this?

I will reactivate the class JRubyRuntimeContext with the aforementioned get(Asciidoctor) method and adapt the Readme if you're ok with that in this PR.

mojavelinux:
> When a Processor creates a new node, the Ruby runtime is taken from the given parent node. 
I think this is a valid assumption.

Yes

asciidoctorj-core/src/main/java/org/asciidoctor/extension/Processor.java
mojavelinux: I don't think I had any intention for the config to be modified after the initializer is called, so I think it's safe to disable this access for now (at least until we have a case).

The InlineMacroProcessor on the Ruby side is still very buggy in the construction phase. That's something I need to deal with in 1.5.3.

asciidoctorj-core/src/main/java/org/asciidoctor/extension/Processor.java
mojavelinux: Let's just cut it if we can. All configuration should be done in the constructor (or sooner).

mojavelinux:
> I think that it should be implemented as a method in the Factory implemented in Asciidoctor interface.

Agreed. Except, I think when you have to pass a Ruby instance to the factory, you should be using an implementation-specific factory (since you are doing a runtime-specific call). You move away from the generic interface if you have these needs.

mojavelinux:
> I will reactivate the class JRubyRuntimeContext with the aforementioned get(Asciidoctor) method and adapt the Readme if you're ok with that in this PR.

I like that idea of using the JRubyRuntimeContext to extract the Ruby instance from the Asciidoctor instance. Very fitting.

mojavelinux:
> she or he could also use a different implementation specific API to create the Asciidoctor implementation.

Exactly.

We do want a way to pass options to the factory method so that we can pass through initialization properties like the JRuby mode. But the interface would only see this is a generic Map and the implementation would pull out what it needs. If we have to pass an actual runtime object (like a Ruby instance), then we should offer that only in the implementation (or its API).

mojavelinux:
> the interface would only see this is a generic Map

Or some sort of generic options object.

asciidoctorj-core/src/main/java/org/asciidoctor/extension/Processor.java
robertpanzer: At the moment I don't see how I can get around without setConfig().
To sync the Java part with the Ruby part the ProcessorProxy has to re-set the config after the Ruby part has been initialized.
(Remember that the Ruby part is always initialized after the Java part)

Maybe someone has an idea to make the config only settable from the ProcessorProxies?
They are in a different package so making them package private will not work either...

Any ideas are welcome.

It would also work out to simply overwrite the config member:
```java
public void setConfig(Map<String, Object> config) {
    this.config = config;
}
```
and assume that the API user does not invoke this method again after the constructor has finished.
Because during the constructor execution calling setConfig() is IMO perfectly valid.

mojavelinux: > assume that the API user does not invoke this method again after the constructor has finished.

Ah yes, that's what I was thinking. Can we do an invocation count and fail (or warn) if it's called a second time? We could also just document that it's not a public API. However, you could make it package private and invoke via reflection. That's usually how API designers get around this sort of thing.

robertpanzer: I think that a concrete implementation might want to try to invoke this method in the constructor.
(We should probably add a test case for that, YellStaticBlock currently does it in a way that not every other implementor might go to forward the config in the super constructor invocation: https://github.com/asciidoctor/asciidoctorj/blob/master/asciidoctorj-core/src/test/java/org/asciidoctor/extension/YellStaticBlock.java#L18 )

What do you think about keeping `setConfig(Map)` public but final and add a new package private method `setConfigFinalized()` that is called by the proxies and sets a flag to indicate that the config must not be reset anymore?
`setConfig(Map)` would throw an `IllegalStateException` if called after `setConfigFinalized()`.

(Nevertheless it still feels a little hacky and I'd prefer a way to already make the compiler refuse such a situation, but I don't see how to do it.)

abelsromero:
> Or some sort of generic options object.

I feel like the problem is the same we have when passing options to render the documents. Why not create an object similar to [Options](https://github.com/asciidoctor/asciidoctorj/blob/master/asciidoctorj-core/src/main/java/org/asciidoctor/Options.java).
Imho, I see this as a good option. It encapsulates the logic and offers a base object, it offers a map-like interface to freely put and retrieve values, and we can always provide a [Builder](https://github.com/asciidoctor/asciidoctorj/blob/master/asciidoctorj-core/src/main/java/org/asciidoctor/OptionsBuilder.java) to ease the most common use cases.

robertpanzer:
Implemented the approach that sets a flag to avoid any later re-sets of Processor.config.
We now have a method `setConfigFinalized()` on `Processor` that is of no use for the client of the API.
But it helps to avoid any Reflection and `setAccessible()` stuff.

What approach do you prefer?

robertpanzer:
Regarding creation of an Asciidoctor instance in a given Ruby runtime I'd like to implement that in the context of #296.

mojavelinux:
> Implemented the approach that sets a flag to avoid any later re-sets of Processor.config. We now have a method setConfigFinalized() on Processor that is of no use for the client of the API. But it helps to avoid any Reflection and setAccessible() stuff.

I like this approach. Keep in mind that we shouldn't worry about getting the API too elegant here for two reasons. 1) 90% of users should (eventually) specify this configuration via annotation metadata and never touch the config API, not even in the constructor and 2) over time we'll sort out what changes we need in core to further improve the Java API. So I think the protection we have now is sufficient.

> Regarding creation of an Asciidoctor instance in a given Ruby runtime I'd like to implement that in the context of #296.

:+1:

mojavelinux:
> 90% of users should (eventually) specify this configuration via annotation metadata and never touch the config API

If you can't tell, I really want the annotation-based configuration to happen :)

robertpanzer:
> If you can't tell, I really want the annotation-based configuration to happen :)

I got it on my list. :)
Will try to propose an example.
Next thing I wanted to do after the Arquillian & this PR are merged was trying to implement the AST renaming as proposed in http://discuss.asciidoctor.org/dev-Reorganization-of-the-AsciidoctorJ-AST-data-structures-td2991.html though.

mojavelinux:
> Next thing I wanted to do ... was trying to implement the AST renaming

Sounds good. That's very important too. Lots of exciting changes!

robertpanzer:
@mojavelinux @lordofthejars Do you think this is ready for merge? It would make my life a little easier if I could continue on that.

mojavelinux:
:+1:

PR #303, opened 2015-05-11
Fixes #302: Pass config options defined in a Java BlockProcessor cons…

### LINKED ISSUE #302: BlockProcessor ist not called for context other than paragraph
In the current 1.6.0 branch a BlockProcessor is not invoked if it is registered as a class and the context is not paragraph.

For this content:

```
[yell]
----
The time is now. Get a move on.
----
```

this processor will currently not be invoked:

```java
package org.asciidoctor.extension;

import org.asciidoctor.ast.AbstractBlock;

import java.util.Arrays;
import java.util.HashMap;
import java.util.List;
import java.util.Map;

public class YellStaticListingBlock extends BlockProcessor {

    private static Map<String, Object> configs = new HashMap<String, Object>() {{
        put("contexts", Arrays.asList(":listing"));
        put("content_model", ":simple");
    }};

    public YellStaticListingBlock(String name) {
        super(name, configs);
    }

    public YellStaticListingBlock(String name, Map<String, Object> config) {
        super(name, configs);
    }

    @Override
    public Object process(AbstractBlock parent, Reader reader, Map<String, Object> attributes) {
        ...
    }

}
```

The problem is root in the order in which Java constructor and Ruby initializer are called in the BlockProcessorProxy.

Paragraphs seem to be a default, therefore this error does not occur in this case.

I always wondered which might be the best order between invoking the Ruby initializer and the Java constructor.
Now I know :)

Will prepare a PR.

robertpanzer:
…tructor to the Ruby part when registered as a class.

See #302 for details.

Changed also the API a bit by requiring that a BlockProcessor registered as a class contains a constructor with only one String parameter.

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/extension/processorproxies/BlockProcessorProxy.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/extension/processorproxies/BlockProcessorProxy.java
@@ -80,13 +80,22 @@ public class BlockProcessorProxy extends AbstractProcessorProxy<BlockProcessor> 
             // The extension config in the Java extension is just a view on the @config member of the Ruby part
             getProcessor().setConfig(new RubyHashMapDecorator((RubyHash) getInstanceVariable(MEMBER_NAME_CONFIG)));
         } else {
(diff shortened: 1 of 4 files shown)


mojavelinux:
Looks good. A necessary step forward.

PR #305, opened 2015-05-12
Added javadoc for extension config options

robertpanzer:
This PR is at the moment WIP!

It adds Javadoc for some config options for processors.

I added constants for `:contexts` config key and the values mentioned in (1).
I guess that this config parameter only makes sense with BlockProcessors.
If the location of the constants is correct I think I got it not too bad.

I added constants for `:content_model` and the values mentioned also mentioned in (1) to Processor.
I think this option makes also sense at least for a BlockMacro.
But I am very unsure about my comments and would like to get some feedback about the descriptions because I did not fully understand them yet.

(1) https://github.com/asciidoctor/asciidoctor-documentation/issues/30 to BlockProcessor.

(While typing this I always got the constant names incorrectly, so IMO it absolutely makes sense to add constants for them :wink:)

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/extension/BlockProcessor.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/extension/BlockProcessor.java
@@ -7,6 +7,124 @@ import org.asciidoctor.ast.AbstractBlock;
 
 public abstract class BlockProcessor extends Processor {
 
+    /**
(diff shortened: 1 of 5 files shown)


lordofthejars:
Really simple addition but it improves a lot the code. :+1:

mojavelinux:
You got it right as far as I can tell. Nice work!

PR #306, opened 2015-05-12
Adapted processor instantiation to same way as a BlockProcessor is in…

robertpanzer:
…stantiated.

https://github.com/asciidoctor/asciidoctorj/commit/009d94887d333139ff8264242e07984fe90bd105 changed the way that BlockProcessors are instantiated if they are registered as classes.
This PR repeats the same for all other processor classes.

Thereby all processor classes now registered as a class now require a constructor without a config map, that is either only with a String or a no-arg constructor.
IMO it does not make much sense to always invoke a constructor with an empty or null config map.


lordofthejars:
:+1:

mojavelinux:
:+1:

PR #308, opened 2015-05-13
Fixes #292. Use own ServiceLoader per Asciidoctor instance to support…

### LINKED ISSUE #292: org.asciidoctor.extension.internal.ExtensionRegistryExecutor use badly service loader
Hi

Shouldnt org.asciidoctor.extension.internal.ExtensionRegistryExecutor use a ServiceLoader by classloader (ie use TCCL each time you create a new instance) to be able to pick up contextual extensions?

I particularly think to the case when adocj is in the container and some extensions in webapps for EE case

If not would be good to at least be able to switch this logic.

robertpanzer:
… multiple differing TCCLs in server envs.

When Asciidoctor instances are created from different contexts in an Enterprise Application the extension registry loader currently uses the first TCCL it gets.
So when an Asciidoctor instance is first created in WebApp 1 and another one in WebApp 2 that are part of the same Enterprise Application and the asciidoctor.jar is in the ear/lib it might not find extensions available in WebApp 2.


mojavelinux:
Absolutely.

I know this one is tough to test for, but if you are ambitious, it's within the realm of what Arquillian can handle (thank goodness for Arquillian).

mojavelinux:
The CI build failure looks like a fluke. I triggered a new build.

robertpanzer:
Please do not merge until I added a test!

robertpanzer:
Should be able to merge now. 
If you revert ExtensionRegistryExecutor to the previous version the test will fail.

mojavelinux:
Super!

lordofthejars:
So I push the button :) Thank you so much Robert this is awesome

PR #309, opened 2015-05-15
Fixes #307. Upgraded slim to 3.0.3 (was 2.1.0)

### LINKED ISSUE #307: Upgrade the Slim gem
Upgrade the Slim gem (to either 2.1.0 or 3.0.3).

AsciidoctorJ currently bundles Slim 2.0.3. The Slim templates in https://github.com/asciidoctor/asciidoctor-backends require Slim 2.1.0 or better. Therefore, it's not possible to use these templates with AsciidoctorJ without using new version of Slim loaded from an external path.

We probably want to consider upgrading related gems as well, such as Tilt. We also need to figure out if there are any migration issues and document those.

robertpanzer:
As Asciidoctor currently depends on sth slim 2.0.0-ish I chose to only upgrade to slim 2.1.0.
I added only a very basic test to see if the slim template is used at all.
It would probably be better to somehow check that really slim 2.1.0 is used...

### DIFF
--- a/asciidoctorj-core/src/test/groovy/org/asciidoctor/WhenSlimTemplatesAreUsed.groovy
+++ b/asciidoctorj-core/src/test/groovy/org/asciidoctor/WhenSlimTemplatesAreUsed.groovy
@@ -1,0 +1,39 @@
+package org.asciidoctor
+
+import org.asciidoctor.internal.JRubyAsciidoctor
+import org.asciidoctor.util.ClasspathResources
+import org.jsoup.Jsoup
+import org.jsoup.nodes.Document
+import org.jsoup.nodes.Element
+import org.junit.Rule
+import spock.lang.Specification
+
+import static org.asciidoctor.OptionsBuilder.options
+
+class WhenSlimTemplatesAreUsed extends Specification {
+
+    @Rule
+    ClasspathResources classpath = new ClasspathResources()
+
+    Asciidoctor asciidoctor = JRubyAsciidoctor.create()
+
+    def 'the slim paragraph template should be used when rendering a document inline'() {
+        given:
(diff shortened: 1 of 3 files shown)


mojavelinux:
I reviewed the changelog for Slim, and did some testing, and have determined that it would be fine to upgrade to Slim 3.0.3.

Here are the only two breaking changes:

* Drop support for Ruby 1.8.7
* Deprecate `='`, `=='` and `tag'` syntax for trailing whitespace. Use `=<` etc. instead.

The first one is not really an issue since Slim 2.1.0 dropped support for jruby18 mode. We don't use this mode by default (and no one else should), so it's a non-issue.

The second one does not break existing templates, it merely issues a warning. So it's an ideal opportunity to notify users to upgrade their templates.

Let's move to Slim 3.0.3.

robertpanzer:
Is slim 3.0.3 (or 3.0.6 respectively) also a candidate for 1.5.3?

mojavelinux:
I think we can move to Slim in 1.5.3. We don't care about Ruby 1.8.7 support in this environment and we've long since fixed the templates that used the old syntax. Since it poses no risk, I think it's smart to switch to it now.

LightGuard:
+1

On Mon, Nov 2, 2015 at 3:38 AM, Dan Allen <notifications@github.com> wrote:

> I think we can move to Slim in 1.5.3. We don't care about Ruby 1.8.7
> support in this environment and we've long since fixed the templates that
> used the old syntax. Since it poses no risk, I think it's smart to switch
> to it now.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/asciidoctor/asciidoctorj/pull/309#issuecomment-152983898>
> .
>



-- 
Jason Porter
http://en.gravatar.com/lightguardjp

PR #310, opened 2015-05-16
Added Arquillian extension that creates, destroys  and injects Asciid…

robertpanzer:
…octor as Arquillian Resources and ClasspathResources.
The Arquillian extension is part of asciidoctorj-core in the test sources.
An own arquillianExtension configuration bundles only the Arquillian extension excluding the core test classes and is referenced by the pdf, epub and diagram modules.

Asciidoctor instances are destroyed using the already available method `shutdown()`.

TemporaryFolder Rules are not replaced yet by the Arquillian extension.
Maybe they could also be replaced in the next step by the Arquillian extension so that it creates the TemporaryFolder instances and invokes `before()` and `after()` (so that we can completely remove the Rule annotation)

### DIFF
--- a/asciidoctorj-core/build.gradle
+++ b/asciidoctorj-core/build.gradle
@@ -1,3 +1,9 @@
+configurations {
+  arquillianExtension
+  testCompile.extendsFrom arquillianExtension
+}
+
+
 dependencies {
   compile "org.jruby:jruby-complete:$jrubyVersion"
   compile "com.beust:jcommander:$jcommanderVersion"
@@ -15,6 +21,12 @@ dependencies {
   testCompile "com.google.guava:guava:$guavaVersion"
   testCompile "org.jsoup:jsoup:$jsoupVersion"
   testCompile project(path: ':asciidoctorj-test-support', configuration: 'tests')
+
+  arquillianExtension "org.jboss.arquillian.container:arquillian-container-spi:$arquillianVersion"
+  arquillianExtension "org.jboss.arquillian.container:arquillian-container-test-spi:$arquillianVersion"
+  arquillianExtension "org.jboss.arquillian.container:arquillian-container-impl-base:$arquillianVersion"
+  arquillianExtension "org.jboss.arquillian.container:arquillian-container-test-impl-base:$arquillianVersion"
+  arquillianExtension "org.jboss.arquillian.junit:arquillian-junit-container:$arquillianVersion"
 }
 
 def gemFiles = fileTree(jruby.gemInstallDir) {
@@ -33,3 +45,15 @@ jrubyPrepareGems << {
 //jruby {
 //	execVersion = '1.7.20'
 //}
+
+task arquillianExtensionJar(type: Jar, dependsOn: testClasses) {
+  baseName = "arquillianExtension-${project.archivesBaseName}"
+  from sourceSets.test.output
+  include 'org/asciidoctor/arquillian/**/*'
+  include 'META-INF/services/*'
+}
+
+
+artifacts {
+  arquillianExtension arquillianExtensionJar
+}
--- a/asciidoctorj-core/build.gradle
+++ b/asciidoctorj-core/build.gradle
@@ -22,6 +22,7 @@ dependencies {
   testCompile "org.jsoup:jsoup:$jsoupVersion"
   testCompile project(path: ':asciidoctorj-test-support', configuration: 'tests')
 
+  arquillianExtension project(path: ':asciidoctorj-test-support', configuration: 'tests')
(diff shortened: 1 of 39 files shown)


mojavelinux:
Yes, I like it.

> TemporaryFolder Rules are not replaced yet by the Arquillian extension.

That's fine to do that as a follow-up task.

> TemporaryFolder instances and invokes before() and after() (so that we can completely remove the Rule annotation)

It makes me happy when Rule annotations go away because it means a more portable test :)

I think we should just make the temporary folder always available, or make it injectable. That way, when you are writing a test, you never have to think about it. That's the key. To make the infrastructure fade away so writing tests seems simple and you want to do it, no matter how complex the scenario.

robertpanzer:
Finally added the ResourceProvider for TemporaryFolder so that all import of Rule and ClassRule should be banned now :)
Injecting a TemporaryFolder supports the same parameters as Asciidoctor.
That is
`@ArquillianResource(Unshared.class) TemporaryFolder tmp`
will give a temp dir only for the test method.

`@ArquillianResource(Shared.class) TemporaryFolder tmp`
will give a temp dir that is valid for the whole test class and is therefore shared among the test methods.

mojavelinux:
Nice!

abelsromero:
I have some questions about this, mainly because I haven't played with arquillian yet. I hope you can help.

Since the extensions are in asciidoctorj-core test space they cannot be used in other projects. But they could be used only with a different packaging right? Maybe as a subproject? If this is possible, I think we could use them also in the maven/gradle plugin (confluence too?).
In maven's, we're manually handling temp and classpath resources and its becoming ugly, not to mention it' prone to errors. Imho, even if it means moving from spock to junit, a common test framework would be a nice addition.

Does it make sense?

mojavelinux:
Indeed. I would be very interested to see how this can be reused for testing in other projects. It's possible to package and distribute the test classes from AsciidoctorJ so that any dependency management system can use them, even without creating a subproject. Gradle is pretty skilled like that.

robertpanzer:
I'd really love to see this reused in other projects!

The Arquillian extension works with junit, as well as spock and should work equally well with testng.
(It will be interesting though to reuse junit's TemporaryFolder in a testng based test...)

I just wonder if it wouldn't be better to create an own submodule for the arquillian extension and make the sources of test-support and the arquillian extension the main sources.
While it will be perfectly possible to publish these 2 jars it will increase the complexity of the build files. 
(And you wouldn't see from the artifact `asciidoctorj-1.x.y-tests.jar` that it contains an Arquillian extension.)


I played around a little bit and publishing a test jar should be as easy as applying the nebula-test-jar plugin:
```
apply plugin: 'nebula.nebula-test-jar'
```
But publishing the test-support module would mean to remove the "standard" publication computed by the nebula-publishing-plugin and activating the jar task again:
```
jar.enabled = true // It's enabled be default, only to highlight the change from the current file

afterEvaluate {
    publishing.publications.mavenNebula.artifacts = publishing.publications.mavenNebula.artifacts.findAll { it.classifier == 'tests' }
}
```

lordofthejars:
Robert I think that you will fall in love to this extension (
https://github.com/arquillian/arquillian-governor). I am currently working
with Stefan to integrate with Github. So you will be able to write tests
with issue information. :)

El dl., 18 maig 2015 a les 10:17, Robert Panzer (<notifications@github.com>)
va escriure:

robertpanzer:
:+1: I also followed that after being announced on Twitter. The idea is really awesome!

mojavelinux:
> I just wonder if it wouldn't be better to create an own submodule for the arquillian extension and make the sources of test-support and the arquillian extension the main sources.

If we can tease apart the API (which was the limiting factor before), then we can certainly do this. I would name it something like asciidoctorj-test (similar in naming to spring-test). Then, we could add different qualifiers if we need more modules or jars. For instance (but just hypothetical), asciidoctorj-test-arquillian or asciidoctorj-test-extra.

robertpanzer:
Got me!:)
Sorry, forgot about this point again, probably working too much again. 
But I'd prefer to do the export of the Arquillian extension and the test module as a separate PR. 
(Maybe we can create an issue so that if someone from the community likes to do it gets a chance, even though I can do it as well)

abelsromero:
> Maybe we can create an issue so that if someone from the community likes to do it gets a chance, even though I can do it as well

I vote for that. On the other issue, we can continue the discussion about the distribution and let Robert merge this awesome code :+1:

robertpanzer:
@mojavelinux @lordofthejars And this as well: Do you think this is ready for merge? It would make my life a little easier if I could continue on that.

lordofthejars:
For me is ok as first iteration.

El dt., 26 maig 2015 a les 14:55, Robert Panzer (<notifications@github.com>)
va escriure:

PR #314, opened 2015-05-23
Rename AST classes

robertpanzer:
This PR contains 2 changes:

First the AST node classes are renamed as discussed in http://discuss.asciidoctor.org/dev-Reorganization-of-the-AsciidoctorJ-AST-data-structures-td2991.html#a2992:
That is the interfaces are now called like this in the package `org.asciidoctor.ast`:
```
Node
  +- BlockNode
  |   +- Block
  |   +- Section
  |   +- Document
  |   +- List
  |   +- ListItem
  +- Inline
```
The implementations are moved to the package `org.asciidoctor.ast.impl` and their class name is the interface name plus `Impl`. (I know that's not very creative...)
This is also the first commit of the two.

The second change is that I extracted interfaces from the classes belonging to the "StructuredDocument" API keeping the name.
The implementing classes were also moved to the package `org.asciidoctor.ast.impl`.
One change that could break current users is that whenever a List<Foo> is returned the new methods return a List<? extends Foo> and the new implementing classes return List<FooImpl>.
I suppose that this API is read-only, as opposed to the Node API, and therefore I think this makes sense.
But I could change that if we want to avoid this incompatibility.


robertpanzer:
Force-pushed after rebasing on top of last changes from Schalk.

robertpanzer:
PLEASE DO NOT MERGE!!!

This PR should server as a discussion how the AST classes should look like.
I will redo the changes once we found a solution and #317 is merged.

PR #315, opened 2015-05-23
Adds target to inline

ysb33r:
This adds a missing method for the Inline type. Anchors of type `link` require the `target` property.


robertpanzer:
Thanks!

Issue #234, opened 2014-12-10
Change convert return type


lordofthejars:
Currently `convert` method returns a String, this works for text backends (html, docbook) but it doesn't work for binary backends like pdf. For this reason we need to change the `convert` method to something like:

``` java
public <T> T convert(Content, Options, Class T)
```

ysb33r:
I see a problem where multiple output files need to be created from a single input source - multi-page HTML would be a good example. What if it just returns an Object, which can then be interpreted by a Writer, which in turn can decide how to create the output.  If asciidoctorj can provide a
```java
interface Writer {
  void write( Object src, File target );
} 
```
which maps to the ruby `Asciidoctor::Writer` module then, then returned output can simply be given to the writer.

mojavelinux:
I'm not quite sure this is the same use case. I agree we probably need the Writer interface, but that's more for the converter than for the main API. The main API is supposed to be getting the final converted result. If you produce multiple documents into memory, then I imagine the OutputStream would be a MultiOutputStream. That's how you'd handle getting back a collection of documents. But what creates those documents is the converter.

robertpanzer:
I have the impression that inventing sth like a Writer and a MultiOutputStream could solve other problems as well:
If AsciidoctorJ could make sth like a Writer accessible to the converter it would be possible to catch all the generated output. 
(In fact with my limited Ruby knowledge the Ruby Writer looks like a FileOutputStream, you pass a stream or a filename and content and this gets appended.)

Because at latest when you generate multi file output the option to not write to a file but directly return the content is not obeyed. 
A Writer that acts according to these options could either directly write to a file or to memory)
But this is just a very vague idea that I have.
(And it would change the programming model of a converter.)

mojavelinux:
@robertpanzer You have the right idea. Asciidoctor checks to see if a converter implements the Writer interface and, if so, allows the converter to handle writing to disk or a stream. The Converter is available from the Document, so it's possible to get access to this information.

To move this forward, I think what we need is to look at this from a high level and create a table of the various scenarios we have. Just to seed it, here are the possible things a converter is going to want to toss back to the caller in some way:

* A String (in the case of a single HTML page)
* A binary object (in the case of the PDF converter)
* A collection of Strings (in the case of multi-page HTML)

[...]

PR #317, opened 2015-05-26
Cache Java AST nodes at Ruby nodes to always convert every Ruby node …

robertpanzer:
…to the same Java node

With this PR the NodeConverter will first ask the NodeCache if there is already a Java Node for the given Ruby node.
The NodeCache caches such information in the new member `@asciidoctorj_cache` that is a RubyHash and will contain the Java object under the key `:asciidoctorj_node`.

Additionally I changed the methods like getDocument(), getParent() & findBy() to always invoke the NodeConverter when returning a node.

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/ast/AbstractBlockImpl.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/ast/AbstractBlockImpl.java
@@ -5,24 +5,22 @@ import java.util.List;
 import java.util.Map;
 
 import org.asciidoctor.converter.ConverterProxy;
+import org.asciidoctor.internal.RubyBlockListDecorator;
 import org.asciidoctor.internal.RubyHashUtil;
 import org.asciidoctor.internal.RubyUtils;
 import org.jruby.Ruby;
 import org.jruby.RubyArray;
 import org.jruby.RubyObject;
 import org.jruby.javasupport.JavaEmbedUtils;
+import org.jruby.runtime.builtin.IRubyObject;
 
 public class AbstractBlockImpl extends AbstractNodeImpl implements AbstractBlock {
 
     private static final String BLOCK_CLASS = "Block";
     private static final String SECTION_CLASS = "Section";
     
-    protected AbstractBlock delegate;
-
-
-    public AbstractBlockImpl(AbstractBlock blockDelegate, Ruby runtime) {
-        super(blockDelegate, runtime);
-        this.delegate = blockDelegate;
+    public AbstractBlockImpl(IRubyObject blockDelegate) {
+        super(blockDelegate);
     }
 
     @Override
@@ -32,7 +30,7 @@ public class AbstractBlockImpl extends AbstractNodeImpl implements AbstractBlock
 
     @Override
     public String getTitle() {
-        return delegate.getTitle();
+        return getString("title");
     }
 
     @Override
@@ -42,7 +40,7 @@ public class AbstractBlockImpl extends AbstractNodeImpl implements AbstractBlock
 
     @Override
     public String getStyle() {
-        return delegate.getStyle();
+        return getString("style");
     }
 
     @Override
@@ -52,16 +50,8 @@ public class AbstractBlockImpl extends AbstractNodeImpl implements AbstractBlock
 
     @Override
     public List<AbstractBlock> getBlocks() {
-        List<AbstractBlock> rubyBlocks = delegate.getBlocks();
-
-        for (int i = 0; i < rubyBlocks.size(); i++) {
-            Object abstractBlock = rubyBlocks.get(i);
-            if (!(abstractBlock instanceof RubyArray)) {
-                rubyBlocks.set(i, (AbstractBlock) NodeConverter.createASTNode(abstractBlock));
-            }
-        }
-
-        return rubyBlocks;
+        RubyArray rubyBlocks = (RubyArray) getProperty("blocks");
+        return new RubyBlockListDecorator(rubyBlocks);
     }
 
     @Override
@@ -71,44 +61,25 @@ public class AbstractBlockImpl extends AbstractNodeImpl implements AbstractBlock
 
     @Override
     public Object getContent() {
-        return delegate.content();
-    }
-
-    @Override
-    public String getNodeName() {
-        return delegate.getNodeName();
+        return JavaEmbedUtils.rubyToJava(getProperty("content"));
     }
 
     @Override
     public String convert() {
-        return delegate.convert();
+        return getString("convert");
     }
 
     @Override
     public int getLevel() {
-        return delegate.getLevel();
-    }
-
-    @Override
-    public AbstractBlock delegate() {
-        return delegate;
+        return getInt("level");
     }
 
     @Override
     public List<AbstractBlock> findBy(Map<Object, Object> selector) {
 
-        @SuppressWarnings("unchecked")
-        List<?> findBy = delegate.findBy(RubyHashUtil.convertMapToRubyHashWithSymbolsIfNecessary(runtime,
+        RubyArray rubyBlocks = (RubyArray) getProperty("find_by", RubyHashUtil.convertMapToRubyHashWithSymbolsIfNecessary(runtime,
                 selector));
-
-        List<AbstractBlock> ret = new ArrayList<AbstractBlock>(findBy.size());
-        for (Object abstractBlock: findBy) {
-            if (!(abstractBlock instanceof RubyArray)) {
-                ret.add((AbstractBlock) NodeConverter.createASTNode(abstractBlock));
-            }
-
-        }
-        return ret;
+        return new RubyBlockListDecorator(rubyBlocks);
     }
 
 }
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/ast/AbstractBlockImpl.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/ast/AbstractBlockImpl.java
@@ -1,17 +1,11 @@
 package org.asciidoctor.ast;
 
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 
-import org.asciidoctor.converter.ConverterProxy;
 import org.asciidoctor.internal.RubyBlockListDecorator;
 import org.asciidoctor.internal.RubyHashUtil;
-import org.asciidoctor.internal.RubyUtils;
-import org.jruby.Ruby;
 import org.jruby.RubyArray;
-import org.jruby.RubyObject;
-import org.jruby.javasupport.JavaEmbedUtils;
 import org.jruby.runtime.builtin.IRubyObject;
 
 public class AbstractBlockImpl extends AbstractNodeImpl implements AbstractBlock {
@@ -50,7 +44,7 @@ public class AbstractBlockImpl extends AbstractNodeImpl implements AbstractBlock
 
     @Override
     public List<AbstractBlock> getBlocks() {
-        RubyArray rubyBlocks = (RubyArray) getProperty("blocks");
+        RubyArray rubyBlocks = (RubyArray) getRubyProperty("blocks");
         return new RubyBlockListDecorator(rubyBlocks);
     }
 
@@ -61,7 +55,7 @@ public class AbstractBlockImpl extends AbstractNodeImpl implements AbstractBlock
 
     @Override
     public Object getContent() {
-        return JavaEmbedUtils.rubyToJava(getProperty("content"));
+        return getProperty("content");
     }
 
     @Override
@@ -77,7 +71,7 @@ public class AbstractBlockImpl extends AbstractNodeImpl implements AbstractBlock
     @Override
     public List<AbstractBlock> findBy(Map<Object, Object> selector) {
 
-        RubyArray rubyBlocks = (RubyArray) getProperty("find_by", RubyHashUtil.convertMapToRubyHashWithSymbolsIfNecessary(runtime,
+        RubyArray rubyBlocks = (RubyArray) getRubyProperty("find_by", RubyHashUtil.convertMapToRubyHashWithSymbolsIfNecessary(runtime,
                 selector));
         return new RubyBlockListDecorator(rubyBlocks);
     }
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/ast/AbstractBlockImpl.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/ast/AbstractBlockImpl.java
@@ -1,5 +1,6 @@
 package org.asciidoctor.ast;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 
@@ -55,9 +56,8 @@ public class AbstractBlockImpl extends AbstractNodeImpl implements AbstractBlock
 
         for (int i = 0; i < rubyBlocks.size(); i++) {
             Object abstractBlock = rubyBlocks.get(i);
-            if (!(abstractBlock instanceof RubyArray) && !(abstractBlock instanceof AbstractNode)) {
-                RubyObject rubyObject = (RubyObject) abstractBlock;
-                rubyBlocks.set(i, (AbstractBlock) NodeConverter.createASTNode(rubyObject));
+            if (!(abstractBlock instanceof RubyArray)) {
+                rubyBlocks.set(i, (AbstractBlock) NodeConverter.createASTNode(abstractBlock));
             }
         }
 
@@ -98,18 +98,17 @@ public class AbstractBlockImpl extends AbstractNodeImpl implements AbstractBlock
     public List<AbstractBlock> findBy(Map<Object, Object> selector) {
 
         @SuppressWarnings("unchecked")
-        List<AbstractBlock> findBy = delegate.findBy(RubyHashUtil.convertMapToRubyHashWithSymbolsIfNecessary(runtime,
+        List<?> findBy = delegate.findBy(RubyHashUtil.convertMapToRubyHashWithSymbolsIfNecessary(runtime,
                 selector));
 
-        for (int i = 0; i < findBy.size(); i++) {
-            Object abstractBlock = findBy.get(i);
-            if (!(abstractBlock instanceof RubyArray) && !(abstractBlock instanceof AbstractBlock)) {
-                RubyObject rubyObject = (RubyObject)abstractBlock;
-                findBy.set(i, (AbstractBlock) NodeConverter.createASTNode(rubyObject));
+        List<AbstractBlock> ret = new ArrayList<AbstractBlock>(findBy.size());
+        for (Object abstractBlock: findBy) {
+            if (!(abstractBlock instanceof RubyArray)) {
+                ret.add((AbstractBlock) NodeConverter.createASTNode(abstractBlock));
             }
 
         }
-        return findBy;
+        return ret;
     }
 
 }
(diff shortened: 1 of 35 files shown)


robertpanzer:
I went further on this PR and now completely separated Java AST objects from Ruby AST objects!
All Java AST nodes have IRubyObjects as delegates instead of magic JRuby proxies that magically transform method calls in a way that we cannot predict. (See also the discussion in #323)

The Java AST implementation classes will also ensure that only Ruby objects will be put into the Ruby classes and therefore removes the mixup between Ruby objects and JRuby proxies generated via JavaEmbedUtils.rubyToJava (in Processor.createBlock())

I also tested with the leanpub converter by @ysb33r and it seems to work fine with this converter as well.

asciidoctorj-core/src/test/java/org/asciidoctor/WhenAsciiDocIsRenderedToDocument.java
```diff
@@ -146,7 +149,7 @@ public void should_be_able_to_get_roles() {
         AbstractBlock abstractBlock = document.blocks().get(0);
         assertThat(abstractBlock.getRole(), is("famous"));
         assertThat(abstractBlock.hasRole("famous"), is(true));
-        //assertThat(abstractBlock.isRole(), is(true));
```
robertpanzer: I don't know why this was commented, but I guess it did not work before :)

lordofthejars: I think it was because of bug in Asciidoctor which didn't return correctly. Well if I remember well it was a problem on how the method was named in Ruby side that didn't match so well in Java part.

lordofthejars:
it looks good to me

asciidoctorj-core/src/test/java/org/asciidoctor/WhenAsciiDocIsRenderedToDocument.java
robertpanzer: I think it was rather the strange behavior of JRuby to map isRole to role instead of role?
That's why I think we should get away from this automatic method mapping.

robertpanzer:
Failed on appveyor due to a timeout on one of the two build environments.
I think there is no real error that made the build fail and build times on travis were pretty normal as well as the succeeding build on appveyor.

mojavelinux:
It's interesting how we've moved away from the automatic method mapping provided by an interface. I think it says a lot about the (lack of) maturity of that feature in JRuby. I agree with this new approach since it gives us much more flexibility and, thus, requires us to create less hacks.

Instead of calling JavaEmbedUtils.rubyToJava explicitly in the accessor methods, could you create a helper method on the abstract class to delegate to this method? Something like "toJava". That way, we don't have the static call hardcoded everywhere.

I made some inline comments where the mappings aren't quite lined up. Aside from those, this looks good.

asciidoctorj-core/src/main/java/org/asciidoctor/ast/AbstractBlockImpl.java
```diff
@@ -71,45 +61,25 @@ public Object content() {
...
-    public String getNodeName() {
-        return delegate.getNodeName();
+        return JavaEmbedUtils.rubyToJava(getProperty("content"));
     }
 
     @Override
     public String convert() {
-        return delegate.convert();
+        return getString("convert");
     }
```
mojavelinux: The `convert` method should return an Object, even though we've been returning a String. This is related to #234.

asciidoctorj-core/src/main/java/org/asciidoctor/ast/AbstractNodeImpl.java
```diff
@@ -54,157 +66,215 @@ public DocumentRuby document() {
 
     @Override
     public DocumentRuby getDocument() {
-        return this.abstractNode.getDocument();
+        return (DocumentRuby) NodeConverter.createASTNode(getProperty("document"));
```
mojavelinux: I'm assuming in 1.6.0, this will return a Document instead of a DocumentRuby. Right?

asciidoctorj-core/src/main/java/org/asciidoctor/ast/AbstractNodeImpl.java
```diff
@@ -54,157 +66,215 @@ public DocumentRuby document() {
...
 
     @Override
     public String normalizeWebPath(String path, String start, boolean preserveUriTarget) {
-        return this.abstractNode.normalizeWebPath(path, start, preserveUriTarget);
+        return getString("normalize_web_path", path, start, preserveUriTarget);
     }
 
     @Override
     public String getStyle() {
+        return getString("@style");
```
mojavelinux: I think this works to call it as a property (style) instead of an instance variable (@style). Right?

asciidoctorj-core/src/main/java/org/asciidoctor/ast/AbstractNodeImpl.java
```diff
@@ -54,157 +66,215 @@ public DocumentRuby document() {
...
+    }
 
-        IRubyObject style = ((RubyObjectHolderProxy) this.abstractNode).__ruby_object()
-                .getInstanceVariables()
-                .getInstanceVariable("@style");
+    @Override
+    public String listMarkerKeyword() {
+        return getString("list_marker_keyword");
```
mojavelinux: We probably want to make this getListMarkerKeyword. Though, honestly, this is an internal method and probably shouldn't be exposed.

asciidoctorj-core/src/main/java/org/asciidoctor/ast/AbstractNodeImpl.java
```diff
@@ -54,157 +66,215 @@ public DocumentRuby document() {
...
+    }
 
-        if (style == null) {
+    @Override
+    public String listMarkerKeyword(String listType) {
+        return getString("list_marker_keyword", listType);
```
mojavelinux: Again, probably an internal method.

asciidoctorj-core/src/main/java/org/asciidoctor/ast/BlockImpl.java
```diff
@@ -3,22 +3,21 @@
...
-    public BlockImpl(Block blockDelegate, Ruby runtime) {
-        super(blockDelegate, runtime);
-        this.blockDelegate = blockDelegate;
+    public BlockImpl(IRubyObject blockDelegate) {
+        super(blockDelegate);
     }
 
     @Override
     public List<String> lines() {
-        return blockDelegate.lines();
```
mojavelinux: We should have an alias getLines() and deprecate lines().

asciidoctorj-core/src/main/java/org/asciidoctor/ast/BlockImpl.java
```diff
@@ -3,22 +3,21 @@
...
+        return getList("lines", String.class);
     }
     
     @Override
     public String source() {
-        return blockDelegate.source();
+        return getString("source");
```
mojavelinux: We should have an alias getSource() and deprecate source().

asciidoctorj-core/src/main/java/org/asciidoctor/ast/Document.java
```diff
@@ -3,51 +3,48 @@
...
-    public DocumentRuby getDocumentRuby() {
-        return documentDelegate;
+    public Document(IRubyObject documentRuby) {
+        super(documentRuby);
     }
 
     @Override
     public boolean basebackend(String backend) {
-        return documentDelegate.basebackend(backend);
+        return getBoolean("basebackend?", backend);
```
mojavelinux: This method should be named isBasebackend()

asciidoctorj-core/src/main/java/org/asciidoctor/ast/InlineImpl.java
```diff
@@ -1,38 +1,36 @@
...
     @Override
     public String render() {
-        return delegate.render();
+        return getString("render");
     }
 
     @Override
     public String convert() {
-        return delegate.convert();
+        return getString("convert");
```
mojavelinux: render() and convert( ) should return an Object (unless we are waiting on #234).

asciidoctorj-core/src/main/java/org/asciidoctor/ast/ListImpl.java
```diff
@@ -1,35 +1,33 @@
...
     @Override
     public List<AbstractBlock> getItems() {
-        return blocks();
+        return getBlocks();
     }
 
     @Override
     public boolean isItem() {
-        return isBlock();
+        return getBoolean("items?");
```
mojavelinux: Should isItem() be named isItems() or, better yet, hasItems()?

asciidoctorj-core/src/main/java/org/asciidoctor/ast/SectionImpl.java
```diff
@@ -1,39 +1,37 @@
 package org.asciidoctor.ast;
 
 import org.jruby.Ruby;
+import org.jruby.runtime.builtin.IRubyObject;
 
 public class SectionImpl extends AbstractBlockImpl implements Section {
```
mojavelinux: We should add "get" methods in Section and SectionImpl and deprecate the methods that don't start with "get".

asciidoctorj-core/src/main/java/org/asciidoctor/ast/SectionImpl.java
```diff
@@ -1,39 +1,37 @@
...
     @Override
     public boolean special() {
-        return this.delegate.special();
+        return getBoolean("special");
     }
 
     @Override
     public int numbered() {
-        return this.delegate.number();
+        return getInt("numbered");
```
mojavelinux: Shouldn't the return type of numbered (or isNumbered) be a boolean?

mojavelinux:
Sorry, I had a lot of inline comments :) Hopefully you can work through them.

asciidoctorj-core/src/main/java/org/asciidoctor/ast/AbstractNodeImpl.java
robertpanzer: Yes! But I think the naming of the nodes is not final yet and Schalk has an immediate need for these changes.
So I'd like to see this coming first and I'll redo the class renaming once we've finished the renaming.

asciidoctorj-core/src/main/java/org/asciidoctor/ast/AbstractNodeImpl.java
robertpanzer: I found no attribute accessor for style on Asciidoctor::AbstractNode.
So I guess I have to get the instance variable myself.

asciidoctorj-core/src/main/java/org/asciidoctor/ast/AbstractBlockImpl.java
robertpanzer: Are you ok if we move this change? 
Because changing the type of this method has some other implications that lead away from what this PR wants to achieve.
See https://github.com/asciidoctor/asciidoctorj/blob/asciidoctorj-1.6.0/asciidoctorj-core/src/main/java/org/asciidoctor/internal/JRubyAsciidoctor.java#L213

asciidoctorj-core/src/main/java/org/asciidoctor/ast/AbstractNodeImpl.java
robertpanzer: I can simply remove these methods if they do not make any sense.
The test suite is still running, except for the test that tests this method ;-)

asciidoctorj-core/src/main/java/org/asciidoctor/ast/AbstractNodeImpl.java
mojavelinux: Aha. `getStyle` shouldn't be available on AbstractNode. The style is only relevant for blocks, so it should only first be available on AbstractBlock. We should remove it here.

asciidoctorj-core/src/main/java/org/asciidoctor/ast/InlineImpl.java
robertpanzer: Like before, I'd like to handle this outside this PR.

asciidoctorj-core/src/main/java/org/asciidoctor/ast/AbstractBlockImpl.java
mojavelinux: Yep, that's fine.

asciidoctorj-core/src/main/java/org/asciidoctor/ast/AbstractNodeImpl.java
mojavelinux: Yep, let's not expose them.

Btw, I do plan on marking private methods on the Ruby side starting in the 1.6.0 release (maybe even earlier). I didn't do it before because, believe it or not, Ruby didn't have a private keyword until 1.9.3. I think it's really important that we start to mark the APIs more explicitly now that we are mapping everything from the Java side. At least, it should make you keep your sanity :)

asciidoctorj-core/src/main/java/org/asciidoctor/ast/InlineImpl.java
mojavelinux: :+1:

asciidoctorj-core/src/main/java/org/asciidoctor/ast/AbstractNodeImpl.java
robertpanzer: :) Now I found the accessor! Thank you!

robertpanzer:
Thank you Dan for these comments!
I really appreciate that!

robertpanzer:
@mojavelinux I think I implemented all your proposed changes. Could you please take another look?

asciidoctorj-core/src/main/java/org/asciidoctor/ast/AbstractNodeImpl.java
```diff
@@ -54,157 +63,211 @@ public DocumentRuby document() {
 
     @Override
     public DocumentRuby getDocument() {
-        return this.abstractNode.getDocument();
+        return (DocumentRuby) NodeConverter.createASTNode(getRubyProperty("document"));
+    }
+
+    public IRubyObject getRubyObject() {
+        return rubyNode;
```
mojavelinux: I think getRubyObject should be either protected or package-private. I'm not sure we should have it exposed (yet) as a public API method.

asciidoctorj-core/src/main/java/org/asciidoctor/ast/AbstractNodeImpl.java
```diff
@@ -54,157 +63,211 @@ public DocumentRuby document() {
...
     @Override
     public Map<String, Object> getAttributes() {
-        return this.abstractNode.getAttributes();
+        return new RubyAttributesMapDecorator((RubyHash) getRubyProperty("attributes"));
     }
 
     @Override
     public Object getAttr(Object name, Object defaultValue, boolean inherit) {
-        return this.abstractNode.getAttr(name, defaultValue, inherit);
+        return JavaEmbedUtils.rubyToJava(getRubyProperty("attr", name, defaultValue, inherit));
```
mojavelinux: I wonder if we should rename getAttr to getAttribute and isAttr to isAttribute. It's sort of in the Java style to have more formal names. What do you think would be more friendly?

robertpanzer: :+1:

robertpanzer: Should we deprecate or remove the old methods?

asciidoctorj-core/src/main/java/org/asciidoctor/ast/AbstractNodeImpl.java
```diff
@@ -1,30 +1,39 @@
...
-        this.abstractNode = abstractNode;
+    public AbstractNodeImpl(IRubyObject rubyNode) {
+        this.rubyNode = rubyNode;
+        this.runtime = rubyNode.getRuntime();
     }
 
     @Override
     public String id() {
-        return this.abstractNode.id();
+        return getString("id");
```
mojavelinux: Can you add getId() and mark id() as deprecated?

asciidoctorj-core/src/main/java/org/asciidoctor/ast/Document.java
```diff
@@ -1,53 +1,51 @@
...
+        } else {
+            return toJava((IRubyObject) doctitle, Title.class);
         }
 
-        return doctitle;
     }
 
     public String doctitle() {
-        return (String) doctitle(new HashMap<Object, Object>());
+        return getString("doctitle");
```
mojavelinux: Could you add getDoctitle() and mark doctitle() as deprecated. Same for the method that takes arguments.

asciidoctorj-core/src/main/java/org/asciidoctor/ast/AbstractNodeImpl.java
robertpanzer: I'm at the phone so I cannot look at the code right now. 
But the method should only be available in the Impl class, not in the interface. 
So the user of the API will not see this method.

mojavelinux:
A few more small things, but we're almost there. This is really taking shape!

asciidoctorj-core/src/main/java/org/asciidoctor/ast/AbstractNodeImpl.java
mojavelinux: Ah, that makes sense. Then I suppose it's good. :+1:

asciidoctorj-core/src/main/java/org/asciidoctor/ast/AbstractNodeImpl.java
mojavelinux: Probably best to discuss as a separate issue. For this pull request, we'll leave it as is.

asciidoctorj-core/src/main/java/org/asciidoctor/ast/Document.java
robertpanzer: Another question:
Do we need the method with the options map?
Why not one method that returns the title as a string and another that returns the Title structure, both without any parameters?

mojavelinux: I'm okay with that. I combined it as a single method in the Ruby side because that's more the Ruby way. But we could totally have a different name in the Java side. Something like getStructuredDoctitle() or something.

asciidoctorj-core/src/main/java/org/asciidoctor/ast/AbstractNodeImpl.java
robertpanzer: Done

asciidoctorj-core/src/main/java/org/asciidoctor/ast/Document.java
robertpanzer: I added `String getDoctitle()` and `Title getStructuredDoctitle()`.
I left `String ductile()` but deprecated it.
And finally I removed `Object doctitle(Map)`

mojavelinux:
Solid! Let's go with it!

robertpanzer:
:beers:

ysb33r:
I ran your changes against the Leanpub converter and I am happy to report that it went through first time without any rework!!

robertpanzer:
Awesome!
Your original problem should work as well!
node.document.options.attributes.docdir

mojavelinux:
\o/

PR #319, opened 2015-05-28
Added new module that publishes the Arquillian extension

### LINKED ISSUE #318: Publish Asciidoctor Arquillian extension
#310 added an Arquillian extension that allows to inject Asciidoctor instances, the ClasspathResources and a TemporaryFolder into the test case.

@abelsromero indicated that he'd like to reuse it in the Asciidoctor Maven plugin.
Therefore the Arquillian extension should be published.

I tried to do sth like that but failed due to apparently lacking knowledge about gradle and the nebula publishing plugin.

There are at least two ways to publish the Arquillian extension:

- Make an own publication asciidoctorj-arquillian-extension with an own pom.xml
- Add the Arquillian extension as an additional artifact to the asciidoctorj module with the classifier `tests`.
  What I dislike about this is that it is afaict difficult to define the dependencies of this artifact, i.e. arquillian-container-impl-base etc. so that they are pulled when depending on the asciidoctorj tests artifact.

robertpanzer:
This should fix #318 

I added a new module asciidoctor-arquillian-extension that creates the jar from dedicated test classes from the asciidoctorj-core project (i.e. all Arquillian classes)
The dependencies on Arquillian are defined there again, so this is maybe not so nice.

I also changed the asciidoctorj-test-support plugin to use the nebula-test-jar plugin.
But maybe it would be better to publish the test-support module not under the tests classifier but without a classifier.
Because now all dependencies in the generated pom are also test dependencies.
And the dependency from asciidoctorj-arquillian-extension to asciidoctorj-test-support does not get the tests classifier so a consumer of the arquillian extension still has to define the dependencies like this:

```
dependencies {
    compile 'org.asciidoctor:asciidoctorj-arquillian-extension:1.5.3-SNAPSHOT'
    compile 'org.asciidoctor:asciidoctorj-test-support:1.5.3-SNAPSHOT:tests'
}
```


robertpanzer:
Apparently nebula does not like Java 6.
Will have to find a different way.

Maybe even simply remove asciidoctorj-test-support and move the classes to the arquillian-extension?

mojavelinux:
> Maybe even simply remove asciidoctorj-test-support and move the classes to the arquillian-extension?

I'm fine with that.

> Apparently nebula does not like Java 6.

I think it's okay if our build requires Java 7. We just need to maintain Java 6 support for running AsciidoctorJ for now (I don't think the world is ready for us to drop the runtime choice yet).

robertpanzer:
I am fine with that. 
I was just surprised that the plugin does not run with Java 6 as I did not test with it locally.

ancho:
Hi,

as you use the helper classes in the test-support package in the context of arquillian, it makes sense to move them to that package.

The test-support package was introduced to remove the duplication of the helper classes. 

But I don't understand why you have to add the test-support dependency as well, if you are using the arquillian-extension dependency. 

The test-support dependency should be transient in your current setup.

You can check that with `gradlew :asciidoctorj-pdf:dependencies` for example.

asciidoctorj-arquillian-extension/build.gradle
```diff
@@ -0,0 +1,28 @@
...
+  compile "org.jboss.arquillian.container:arquillian-container-test-spi:$arquillianVersion"
+  compile "org.jboss.arquillian.container:arquillian-container-impl-base:$arquillianVersion"
+  compile "org.jboss.arquillian.container:arquillian-container-test-impl-base:$arquillianVersion"
+}
+
+jar.dependsOn project(':asciidoctorj').testClasses
+
+jar {
+  from project(':asciidoctorj').sourceSets.test.output
+  include 'org/asciidoctor/arquillian/**/*'
```
ancho: Why don't you move the arquillian specific classes from core to this module?
You wouldn't need the dependency to the core module anymore.

robertpanzer: That's because the core tests use the Arquillian extension and the Arquillian extension uses the core main classes. 
If I pulled these classes into an own module I had a cyclic dependency.

The only way I see at the moment to resolve this would be to split into core into an API and an Impl part.

ancho: Ah. Nevermind. Didn't saw the dependency of the  `Asciidoctor` class in the other classes.

robertpanzer: Maybe it'll work with inter module task deps.

ancho: Thanks for the clarification. Splitting it is maybe a good idea.
I tried to use the extension in an example project. 

```
apply plugin: 'groovy'
apply plugin: 'idea'

group 'de.calmdevelopment'
version '0.1'

defaultTasks 'clean', 'build', 'eclipse'

repositories {
  mavenLocal()
  jcenter()
}

dependencies {
  compile 'org.codehaus.groovy:groovy-all:2.4.3'
  testCompile 'org.spockframework:spock-core:1.0-groovy-2.4'
  testCompile 'org.asciidoctor:asciidoctorj-arquillian-extension:1.5.3-SNAPSHOT'
}
```

Just to see the dependency-graph:

```
testCompile - Compile classpath for source set 'test'.
+--- org.codehaus.groovy:groovy-all:2.4.3
+--- org.spockframework:spock-core:1.0-groovy-2.4
|    +--- org.codehaus.groovy:groovy-all:2.4.1 -> 2.4.3
|    \--- junit:junit:4.12
|         \--- org.hamcrest:hamcrest-core:1.3
\--- org.asciidoctor:asciidoctorj-arquillian-extension:1.5.3-SNAPSHOT
     +--- org.asciidoctor:asciidoctorj:1.5.3-SNAPSHOT
     |    +--- org.jruby:jruby-complete:1.7.20
     |    \--- com.beust:jcommander:1.35
     +--- org.asciidoctor:asciidoctorj-test-support:1.5.3-SNAPSHOT
     |    \--- junit:junit:4.12 (*)
     +--- org.jboss.arquillian.container:arquillian-container-spi:1.1.8.Final
     |    +--- org.jboss.arquillian.core:arquillian-core-spi:1.1.8.Final
     |    |    \--- org.jboss.arquillian.core:arquillian-core-api:1.1.8.Final
     |    +--- org.jboss.arquillian.config:arquillian-config-api:1.1.8.Final
     |    |    \--- org.jboss.shrinkwrap.descriptors:shrinkwrap-descriptors-api-base:2.0.0-alpha-7
     |    +--- org.jboss.arquillian.config:arquillian-config-impl-base:1.1.8.Final
     |    |    +--- org.jboss.arquillian.config:arquillian-config-api:1.1.8.Final (*)
     |    |    +--- org.jboss.arquillian.core:arquillian-core-api:1.1.8.Final
     |    |    +--- org.jboss.arquillian.core:arquillian-core-spi:1.1.8.Final (*)
     |    |    \--- org.jboss.shrinkwrap.descriptors:shrinkwrap-descriptors-spi:2.0.0-alpha-7
     |    |         \--- org.jboss.shrinkwrap.descriptors:shrinkwrap-descriptors-api-base:2.0.0-alpha-7
     |    +--- org.jboss.shrinkwrap:shrinkwrap-api:1.2.2
     |    \--- org.jboss.shrinkwrap.descriptors:shrinkwrap-descriptors-api-base:2.0.0-alpha-7
     +--- org.jboss.arquillian.container:arquillian-container-test-spi:1.1.8.Final
     |    +--- org.jboss.arquillian.container:arquillian-container-spi:1.1.8.Final (*)
     |    +--- org.jboss.arquillian.test:arquillian-test-spi:1.1.8.Final
     |    |    +--- org.jboss.arquillian.core:arquillian-core-spi:1.1.8.Final (*)
     |    |    \--- org.jboss.arquillian.test:arquillian-test-api:1.1.8.Final
     |    |         \--- org.jboss.arquillian.core:arquillian-core-api:1.1.8.Final
     |    \--- org.jboss.arquillian.container:arquillian-container-test-api:1.1.8.Final
     |         \--- org.jboss.shrinkwrap:shrinkwrap-api:1.2.2
     +--- org.jboss.arquillian.container:arquillian-container-impl-base:1.1.8.Final
     |    +--- org.jboss.arquillian.core:arquillian-core-api:1.1.8.Final
     |    +--- org.jboss.arquillian.core:arquillian-core-spi:1.1.8.Final (*)
     |    +--- org.jboss.arquillian.config:arquillian-config-api:1.1.8.Final (*)
     |    +--- org.jboss.arquillian.config:arquillian-config-impl-base:1.1.8.Final (*)
     |    +--- org.jboss.arquillian.container:arquillian-container-spi:1.1.8.Final (*)
     |    \--- org.jboss.shrinkwrap.descriptors:shrinkwrap-descriptors-spi:2.0.0-alpha-7 (*)
     \--- org.jboss.arquillian.container:arquillian-container-test-impl-base:1.1.8.Final
          +--- org.jboss.arquillian.container:arquillian-container-spi:1.1.8.Final (*)
          +--- org.jboss.arquillian.test:arquillian-test-api:1.1.8.Final (*)
          +--- org.jboss.arquillian.container:arquillian-container-test-api:1.1.8.Final (*)
          \--- org.jboss.arquillian.container:arquillian-container-test-spi:1.1.8.Final (*)
...
```

asciidoctorj as full dependency seems a bit to much.

robertpanzer: Can't really read it on my Phoenix atm;-)
I think it would be best to have a provided dependency, though I don't know how to create it in gradle.

ancho: Hmm....that's an idea...

Well, there is another nebula plugin for that. 
https://github.com/nebula-plugins/gradle-extra-configurations-plugin

robertpanzer: If you like to try it out and prepare a PR I would happily close this one:)

robertpanzer: BTW: I am a little disappointed that you always need some extra plugins to fully cover Maven's possibilities. 
While I prefer to not export test deps I just can't understand why there has to be a pile of gradle plugins for provided and optional. Nebula is one and there is at least another one from spring that does probably the same.
But that's just my own opinion.

ancho: Yeah. It is a very old issue. https://issues.gradle.org/browse/GRADLE-784 ;(
Strange. But gradle is very strong and flexible. I like it anyway :)

I'll try my luck applying the extra-configurations plugin.

ancho: Ok. So the extra-configurations plugin is incompatible to java6, like the publishing plugin.
What about that issue?

robertpanzer: Maybe this module could be skipped on Java 6 (like loading publish.gradle is skipped). 
In the end the only sense of the Arquillian extension module is to export these classes, so it should be no problem to publish with Java >= 7.

mojavelinux: > I am a little disappointed that you always need some extra plugins to fully cover Maven's possibilities.

My feeling exactly. I love Gradle's power and I think it is fully capable of making tasks like this easy...but it just seems to be left as unfinished work. The best thing we can do for Gradle (and ourselves) is to at least make the core engineers aware of these shortcomings (in a blog, forum post, etc). If they continue to ignore the gravity of the situation, then we can let the disappointment show through.

robertpanzer:
@mojavelinux If you have the time, could you please take another look at this PR? It got some great additions from @ancho!

The sources of the arquillian extension are now in the asciidoctorj-arquillian-extension module without a split into API and Impl.

robertpanzer:
@ancho Thanks!!!!

Rebased and merged locally and pushed to branch asciidoctorj-1.6.0

ancho:
Thank you too @robertpanzer. You are doing a great job.

mojavelinux:
Agreed!

PR #321, opened 2015-05-29
Added new CLI option --cp that allows to set a classpath to load exte…

robertpanzer:
…rnal converters and extensions that register themselves via ExtensionRegistry or ConverterRegistry

Before instantiating Asciidoctor the TCCL is set to the user defined classpath.
The classpath elements have to defined separately and may contain wildcards, the GlobDirectoryWalke is reused.
That is an invocation could look like this:

    asciidoctorj thisis.adoc --cp /my/dir/containing/jars/*.jar my/other/relative/dir/containing/jars/*

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/cli/AsciidoctorInvoker.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/cli/AsciidoctorInvoker.java
@@ -1,6 +1,9 @@
 package org.asciidoctor.cli;
 
 import java.io.File;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.net.URLClassLoader;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


asciidoctorj-core/src/main/java/org/asciidoctor/cli/AsciidoctorInvoker.java
```diff
@@ -100,13 +103,37 @@ private void setVerboseLevel(AsciidoctorCliOptions asciidoctorCliOptions) {
...
+            } else {
+                asciidoctor = JRubyAsciidoctor.create();
+            }
+            return asciidoctor;
+        } finally {
+            Thread.currentThread().setContextClassLoader(oldTccl);
+        }
+    }
+
+    private URLClassLoader createUrlClassLoader(List<String> classPaths) {
```
ysb33r: What I am wondering is whether  `--cp /path/to/dirFullOfJars`, which is what I think you implemented, or whether `--cp /path/tp/myconverter.jar` is the better cmdline interface. The latter might be more Java-like, so more understandable for JVM people, whereas the first might be more friednly towards less-technical people.

robertpanzer: The first example `--cp /path/to/dirFullOfJars` will actually *not* work!
Instead it should be `--cp /path/to/dirFullOfJars/*` and this is imo very java-like.
Because you can also invoke `java -cp /path/to/dirFullOfJars/* my/MainClass`.

ysb33r: > Because you can also invoke java -cp /path/to/dirFullOfJars/* my/MainClass.

Which AFAIK means it will look for *.class files. 

So if I specify `--cp /path/tp/myconverter.jar` will it still work?

robertpanzer: `--cp mydir` will load class files directly from `mydir`
`--cp mydir/*` will load classes from archives that are in `mydir` like `mydir/myconverter.jar`
`--cp mydir/myconverter.jar` will load classes from exactly that jar

ysb33r: :+1:

mojavelinux:
I think the option name should be `--classpath` instead of `--cp`. If an abbreviation is used, it should only be a single letter (like `-p`). We could even consider the pattern that JRuby uses, which would be `-J-cp`...though I'm not a huge fan of that. (I hate that Sun didn't follow the Unix conventions for the `java` command.)

mojavelinux:
Another option for the short flag is `-P` (for path).

robertpanzer:
Then I prefer --classpath because there is also the load path to find the Ruby files. 
Only -P(ath) looks ambiguous. I'd make it wrong in one of two cases;-)

ysb33r:
I'm not bothered between `--classpath` and `--cp`. The only reason I used `--cp`, is becasue I'm used to in Groovy that you can pass either. (But I'm with @mojavelinux, I'm not for a single dash, it has to be two dashes).

mojavelinux:
> . The only reason I used --cp, is becasue I'm used to in Groovy that you can pass either.

Ah! If it's in Groovy, then it's an established flag and therefore I change my vote to +1. I think we should support both.

robertpanzer:
BTW: Is everyone comfortable with separating multiple paths with blanks?
Java prefers the : or ;

ysb33r:
I would stick to what the platform convention is. Using spaces would just be confuring in my opinion. I would rather say allow the `--classpath` option to be repeated.

ysb33r:
> >    . The only reason I used --cp, is becasue I'm used to in Groovy that you can pass either.

> Ah! If it's in Groovy, then it's an established flag and therefore I change my vote to +1. I think we should support both.

I am retracting my statement. I went back to check Groovy and it was `-cp`, not `--cp`. Groovy has `-cp` & `--classpath`, so I say et's just stick to `--classpath`.

robertpanzer:
I used the blank initially because it's the same format as the parameter for the Ruby load path. 

Should we 
A) support ; and : respectively for classpath and keep blank for load path?
B) Support blank as well as : and ; respectively for classpath
C) Change both parameters to using : and ; respectively and break the current behavior of the load path?

One thing I am also not content with is that as soon as you add the load path option the input file name has to be defined before the option, that is you can't do

asciidoctorj -I dirA my.adoc

my.adoc will always be treated as if it belongs to the options.

robertpanzer:
BTW: It's not only groovy that supports `-cp` but also java itself.
`-classpath` and `-cp` seem to be the same:
```
    -cp <class search path of directories and zip/jar files>
    -classpath <class search path of directories and zip/jar files>
                  A : separated list of directories, JAR archives,
                  and ZIP archives to search for class files.
```

mojavelinux:
The behavior in AsciidoctorJ should match the behavior of Java (and Groovy), including the path separators.

In fact, I'm even willing to say that we should just go ahead and match the option name too, `-cp` and `-classpath`. Perhaps for consistently we can add `--classpath` as well. It seems like a lot, but having all three would cover all basis.

robertpanzer:
Changed to support `-cp`, `-classpath` and `--classpath`
Additionally multiple paths must now be separated by a path separator (; or :) instead of blanks.
Thereby defining the input file as last parameter is possible now when a load path or class path is defined.

mojavelinux:
Super!

Issue #297, opened 2015-05-06
OSGi support for AsciidoctorJ


lefou:
As a follow up to #293 and #294, we came to the conclusion that it would be nice to have better support for running AsciidoctorJ in OSGi environments.

Currently, the bundle using the Asciidoctor API needs to set it up including knowing all the details (e.g. setup of JRuby and it's gems, classloading, ...). If more than one bundle want to do this, we end up having lots of Asciidoctor (and JRuby) instances in the same JVM, which is not so good. The proper OSGi way would be to register a service into the OSGi registry that covers the Asciidoctor API. Consumers do not need to setup Asciidoctor themselves but instead consume the service from the OSGi registry. Provided that such a service exists, classloading, versioning, bootstrapping and such things would no longer be an issue.

What's needed for this:
* Make the asciidoctor-core jar a proper OSGi bundle (MANIFEST.MF headers)
* Decide the service API, e.g. `Asciidoctor` interface or something else
* Write an OSGi Activator which registers the service into the service registry
* Write tests and documentation how to consume the service
* Think about a proper mechanism for extension, e.g. the [Whiteboard Pattern](http://wiki.osgi.org/wiki/Whiteboard_Pattern) comes to mind.

lordofthejars:
I have never done anything in OSGi but I know the concepts. Do you know any simple library, example who uses this approach so we can see how it can be done?

mojavelinux:
@lordofthejars What I recommend here is an OSGi Arquillian test. It's a perfect use case because we can isolate the environment and ensure the changes function properly without manual testing. It's also a good opportunity / excuse to play with OSGi container support in Arquillian.

mojavelinux:
Of course, one example of AsciidoctorJ in use in OSGi is SlideshowFX.

lefou:
@lordofthejars Do you need an example of how to register/use services in OSGi? I'm not sure I understand your request correctly.

@mojavelinux Yes, I already had a discussion wirth @twasyl about this topic and SlideshowFX might be a good testbed/client for this feature. Currently, it's use of AsciidoctorJ differs from my view of the strived target solution: SlideshowFX (SFX) uses one single bundle to provide it's own API which uses Asciidoctor internally. That bundle embeds AsciidoctorJ and all required dependencies including JRuby. That is, IMHO, not the preferred general solution we want to provide. It think we should:

* Use the official JRuby bundle (at least jruby-complete looks like a proper OSGi bundle)
[...]

PR #328, opened 2015-06-03
Add OSGi metadata to asciidoctorj-core jar

lefou:
This just added some OSGi Metadata to the asciidoctorj-core.jar and makes it a OSGi bundle. That way, it can be loaded and used in OSGi environments as-is.

It does not add any special OSGi support, e.g. a `BundleActivator` or some service registrations. See also #297.

### DIFF
--- a/asciidoctorj-core/build.gradle
+++ b/asciidoctorj-core/build.gradle
@@ -39,12 +39,12 @@ jar {
   manifest {
     symbolicName = 'org.asciidoctor'
     instruction 'Export-Package',
-      'org.asciidoctor',
-      'org.asciidoctor.ast',
-      'org.asciidoctor.converter',
-      'org.asciidoctor.converter.spi',
-      'org.asciidoctor.extension',
-      'org.asciidoctor.extension.spi'
+      "org.asciidoctor;version=\"${version}\"",
+      "org.asciidoctor.ast;version=\"${version}\"",
+      "org.asciidoctor.converter;version=\"${version}\"",
+      "org.asciidoctor.converter.spi;version=\"${version}\"",
+      "org.asciidoctor.extension;version=\"${version}\"",
+      "org.asciidoctor.extension.spi;version=\"${version}\""
     instruction 'Import-Package',
       'com.beust.jcommander;resolution:=optional',
       '*'
--- a/asciidoctorj-core/build.gradle
+++ b/asciidoctorj-core/build.gradle
@@ -1,3 +1,5 @@
+apply plugin: 'osgi'
+
 dependencies {
   compile "org.jruby:jruby-complete:$jrubyVersion"
   compile "com.beust:jcommander:$jcommanderVersion"
@@ -32,3 +34,19 @@ jrubyPrepareGems << {
 //jruby {
 //	execVersion = '1.7.20'
 //}
+
+jar {
+  manifest {
+    symbolicName = 'org.asciidoctor'
+    instruction 'Export-Package',
+      'org.asciidoctor',
+      'org.asciidoctor.ast',
+      'org.asciidoctor.converter',
+      'org.asciidoctor.converter.spi',
+      'org.asciidoctor.extension',
+      'org.asciidoctor.extension.spi'
+    instruction 'Import-Package',
+      'com.beust.jcommander;resolution:=optional',
+      '*'
+  }
+}


lordofthejars:
I have no experience in OSGi so I cannot say anything, but it seems good to me for what I know about OSGi.

asciidoctorj-core/build.gradle
```diff
@@ -32,3 +34,19 @@ jrubyPrepareGems << {
...
+    symbolicName = 'org.asciidoctor'
+    instruction 'Export-Package',
+      "org.asciidoctor;version=\"${version}\"",
+      "org.asciidoctor.ast;version=\"${version}\"",
+      "org.asciidoctor.converter;version=\"${version}\"",
+      "org.asciidoctor.converter.spi;version=\"${version}\"",
+      "org.asciidoctor.extension;version=\"${version}\"",
+      "org.asciidoctor.extension.spi;version=\"${version}\""
+    instruction 'Import-Package',
+      'com.beust.jcommander;resolution:=optional',
```
mojavelinux: Do you think this line is necessary? Would the CLI ever be used from an OSGi environment. It's not really intended to be an API atm. But perhaps I'm missing a use case.

mojavelinux:
:+1:

asciidoctorj-core/build.gradle
lefou: Absolutely. It means that this import isn't required, so that the bundle will resolve even if the package `com.beust.jcommander` isn't available. All other packages are required though.

The gradle osgi plugin uses the bnd tool (https://github.com/bndtools/bnd) which processes these lines to produce the actual JAR and the final MANIFEST.MF.

You can also see that `org.asciidoctor.cli` is not part of the exported packages. It is still in the JAR but not exported, thus in an OSGi environment, other bundles can not use/see it.

mojavelinux: Thanks for the clarification! Now I get it.

lefou:
Are there any more objections or questions?

mojavelinux:
Not from me. My vote is to merge. @robertpanzer or @lordofthejars?

robertpanzer:
I wonder if we want to add integration tests in an osgi environment, i.e. Arquillian-osgi?

lefou:
Unfortunately, I'm not familiar with Arquillian-osgi. Also, there is not much to test, currently; only that the bundle resolves properly after all required dependencies are resolved.

robertpanzer:
> Unfortunately, I'm not familiar with Arquillian-osgi. 
> 
No problem, I can have a try once I find some time. I haven't done much yet wrt osgi;-) except for some eclipse plugin development. 

> Also, there is not much to test, currently
> 

I was just thinking about some smoke test so that we would be warned before releasing when sth changed. 

Best regards
Robert

lefou:
Yeah, you are right. We should test if using the Asciidoctor API in OSGi actually works. So, we need to write and test a consumer bundle. Later, when we introduce a proper µService via the OSGi registry, we would need to refactor that test though.

If somebody comes up with a skeleton for the test, I can help to fill in the gaps.

robertpanzer:
Cherry-picked in asciidoctorj-1.6.0 87ff77327bc94ef6acd8331b4d68f43a03bf001f
Started playing with an Arquillian test but on first try current arquillian-osgi container seems to be incompatible with arquillian-core 1.1.8.Final.

robertpanzer:
Just to keep the info somewhere:
I am seeing this error when trying to start an arquillian-container-felix-embedded:

```
Caused by: javax.management.ReflectionException: Signature mismatch for operation runTestMethod: (java.lang.String, java.lang.String, java.util.Map) should be (java.lang.String, java.lang.String)
	at com.sun.jmx.mbeanserver.PerInterface.noSuchMethod(PerInterface.java:170)
	at com.sun.jmx.mbeanserver.PerInterface.invoke(PerInterface.java:135)
	at com.sun.jmx.mbeanserver.MBeanSupport.invoke(MBeanSupport.java:252)
	at com.sun.jmx.interceptor.DefaultMBeanServerInterceptor.invoke(DefaultMBeanServerInterceptor.java:819)
	at com.sun.jmx.mbeanserver.JmxMBeanServer.invoke(JmxMBeanServer.java:801)
	at javax.management.MBeanServerInvocationHandler.invoke(MBeanServerInvocationHandler.java:292)
	... 97 more
Caused by: java.lang.NoSuchMethodException: runTestMethod(java.lang.String, java.lang.String, java.util.Map)
	at com.sun.jmx.mbeanserver.PerInterface.noSuchMethod(PerInterface.java:169)
	... 102 more
```

The arquillian-jmx-protocol invokes `runTestMethod(String, String, Map)` via JMX.
The target of this call is part of the arquillian-osgi-bundle to be retrieved from Maven central separately. 
This contains the arquillian libs in version 1.1.2.Final. (Strange that it has an own versioning that aligns with the OSGI adapter instead of aligning with arquillian-core)
And this method does not exist in 1.1.2.Final. There the method had only 2 params.

PR #332, opened 2015-06-04
Added documentation about how to write a custom converter in Java

robertpanzer:


README.adoc
```diff
@@ -1511,6 +1511,110 @@ assertThat(new File("target/test-classes/index.epub").exists(), is(true));
 // TODO document how to convert to KF8/MOBI (need to set KINDLEGEN environment variable)
 // TODO add section that covers using Asciidoctor PDF via the AsciidoctorJ PDF jar
 
+
+== Writing a custom converter
+
+For output formats that are not natively supported by Asciidoctor it is possible to write an own converter in Java.
+To get an own converter running in {uri-repo}[AsciidoctorJ] these steps are required:
```
mojavelinux: Instead of "an own converter", you should write "your own converter".

README.adoc
```diff
@@ -1511,6 +1511,110 @@ assertThat(new File("target/test-classes/index.epub").exists(), is(true));
...
+.org.asciidoctor.converter.spi.TextConverterRegistry
+----
+package org.asciidoctor.converter.spi;
+
+import org.asciidoctor.Asciidoctor;
+
+public class TextConverterRegistry implements ConverterRegistry{
+    @Override
+    public void register(Asciidoctor asciidoctor) {
+        asciidoctor.javaConverterRegistry().register(TextConverter.class, "textt");
```
mojavelinux: There's a double t in "text".

README.adoc
```diff
@@ -1511,6 +1511,110 @@ assertThat(new File("target/test-classes/index.epub").exists(), is(true));
...
+----
+org.asciidoctor.converter.spi.TextConverterRegistry
+----
+
+To render a document with this converter the backend name `text` has to be passed via the options:
+
+[source,java]
+----
+File adocFile = ...
+asciidoctor.renderFile(adocFile, OptionsBuilder.options().backend("text"));
```
mojavelinux: We should be using "convertFile" instead of "renderFile" for all new docs.

README.adoc
```diff
@@ -1511,6 +1511,110 @@ assertThat(new File("target/test-classes/index.epub").exists(), is(true));
...
+<4> Calls to the method `getContent()` of a node will recursively call the method `convert()` with the child nodes again.
+    Thereby the converter can collect the rendered child nodes, merge them appropriately and return the rendering of the whole node.
+
+Then the converter has to be registered like this with a well defined backend name:
+
+[source,java]
+.Register a converter
+----
+Asciidoctor asciidoctor = Asciidoctor.Factory.create();
+asciidoctor.javaConverterRegistry().register(TextConverter.class, "text"); // <1>
```
mojavelinux: I wonder if it would be possible to read the backend name from the converter class so that it is optional when registering it. Something like:

```
register(TextConverter.class);
```

robertpanzer: Via an annotation? ;-)

mojavelinux: Ideally, an annotation. Another option is a static method like `public static String getName()`. The idea is, like with processors, this is the name that is used if one is not explicitly provided at the point of registration.

mojavelinux: ...btw, I like this option because it keeps the data in the extension class (in this case the converter).

robertpanzer:
Completely removed any hint on passing backend name when registering a converter.
Instead document annotation based registration and make use of the term _target format_.

mojavelinux:
:+1:

PR #335, opened 2015-06-08
Added first support for tables

robertpanzer:
Introduced a new AST type `Table` that is an `AbstractBlock`.
This owns
- a list of `Column`s that are `AbstractNode`s.
- a list of header `Row`s
- a list of footer `Row`s
- a list of body `Row`s

A `Column` has accessors for
- the column number
- the width
- horizontal alignment
- vertical alignment

A row just owns a list of `Cell`s that are `AbstractNode`s.
A `Cell` has accessors for
- colspan
- rowspan
- the text
- the content Object in case of an embedded asciidoc document (not tested yet)


mojavelinux:
Very good!

Btw, the best way to determine what is needed in the public API is to look at what is used in the converter.

* https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/converter/html5.rb#L742-L814
* https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/converter/docbook5.rb#L374-L434

I know this isn't the best way to define an API, but hopefully it helps you with this task.

robertpanzer:
I added getters and setters for alignment, grids and frames.
And a test case that creates a table and renders it to docbook and html.

There are some attributes and options that are computed or have to be defined in the Ruby initializer that are used by the html5 and docbook converters like the node option `pgwide` that does not appear in table.rb and `tablepcwidth` that is computed from the `width` attribute in `Asciidoctor::Table::initialize`.
How should they be supported?

I like the uber-cool extension that creates a table from the contributors of an arbitrary github project :smile:

mojavelinux:
> There are some attributes and options that are computed or have to be defined in the Ruby initializer that are used by the html5 and docbook converters 

This is where we need to push for changes upstream (core) so that they are more accessible. To be honest, some of them we might not even want to support (or perhaps redesign) because they were legacy requirements for mapping to AsciiDoc Python. A good opportunity for improvement. The way forward is to think about what you want it to be, then we'll align core.

> I like the uber-cool extension that creates a table from the contributors of an arbitrary github project

Indeed! Would you be willing to seed the asciidoctorj-extension-labs (notice the "J" part) so that we can start to build up some extension prototypes on the AsciidoctorJ side? https://github.com/asciidoctor/asciidoctorj-extensions-lab. Maybe this would be a good candidate.

mojavelinux:
> And a test case that creates a table and renders it to docbook and html.

Super cool!

mojavelinux:
Small change request. Let's spell out the alignment properties, verticalAlignment and horizontalAlignment instead of HAlign and VAlign. I think that's a friendly API for Java land.

asciidoctorj-core/src/test/groovy/org/asciidoctor/extension/GithubContributorsBlockMacro.groovy
```diff
@@ -0,0 +1,45 @@
...
+
+class GithubContributorsBlockMacro extends BlockMacroProcessor {
+
+    GithubContributorsBlockMacro(String macroName) {
+        super(macroName)
+    }
+
+    @Override
+    Object process(AbstractBlock parent, String target, Map<String, Object> attributes) {
+        String content = new URL("https://api.github.com/repos/${target}/contributors").text
```
mojavelinux: For the purpose of the test, I recommend using local date for this extension. You can activate the real data for the version in the asciidoctorj-extensions-lab.

robertpanzer:
> Indeed! Would you be willing to seed the asciidoctorj-extension-labs (notice the "J" part) so that we can start to build up some extension prototypes on the AsciidoctorJ side?

Sure! Once we have table support merged I could contribute an extension like this, as I believe that this is sth that can be used be many.
Just imagine someone having to provide a release document containing resolved issues etc.
Being able to simply generate this information via a macro instead of exporting this data from the issue management system is ground-breaking. (I know what I am speaking about, believe me.)

> Small change request. Let's spell out the alignment properties, verticalAlignment and horizontalAlignment instead of HAlign and VAlign. I think that's a friendly API for Java land.

Sure, I was also thinking about this, but somehow got stuck with the existing names.
Will change it.

> For the purpose of the test, I recommend using offline data for this extension. You can activate the real data for the version in the asciidoctorj-extensions-lab.

Yes, will provide some static test data that has a similar format with some names like Alice, Bob and Charlie.
Nevertheless we also have some tests that require a connection to github, i.e. all tests that use the UriIncludeProcessor.

I will take a look again at the other attributes and options used by tables etc and try to sort out what I think makes sense to make public to the user and what is more sort of an implementation detail.
Should we drive changes in Asciidoctor and this branch in parallel until everything is sorted out or do you think it would be worth first start merging this and then possibly collecting further input from other folks like Schalk? (I just can't stress enough the value of his work on the leanpub converter for AsciidoctorJ)

mojavelinux:
> Nevertheless we also have some tests that require a connection to github, i.e. all tests that use the UriIncludeProcessor.

This can actually be solved by using an in-memory server like I do in the Ruby-based tests. See https://github.com/asciidoctor/asciidoctor/blob/master/test/test_helper.rb#L287-L337. Probably even simpler in Java.

robertpanzer:
> Probably even simpler in Java.

And probably even simpler with Groovy super powers :)

mojavelinux:
I don't think you need to wait for changes in core to merge this, but there may be some gaps until the necessary changes are made in core. To use an old saying we once had in a previous company, "we don't want big changes to get stuck behind small ones."

> I just can't stress enough the value of his work on the leanpub converter for AsciidoctorJ

robertpanzer:
Renamed properties and enums HAlign and VAlign to HorizontalAlignment and VerticalAlignment respectively.

All remote resources are now delivered by a local HTTP server so that the tests will now no longer access resources from the internet. (And the build therefore also work if you're sitting behind a proxy)
This test server opens an ephemeral port, so that no port conflicts should occur.
It also serves only resources defined by the test case, nothing else.

mojavelinux:
Great job!

robertpanzer:
@mojavelinux @alexsotob BTW I first wanted to code the HTTP server in groovy. 
But Java classes in src/test/java can't see classes from src/test/groovy. 
Do you think it would make sense to move all Java test code to src/test/groovy?
Afaict Java and Groovy code should then be able to interact seamlessly.

mojavelinux:
Perhaps ask the Twitters about this. There may be a solution we're not seeing. If not, then I say go for it.

ancho:
Phew. That's an interesting testscenario. It's common to access java classes from groovy but not the other way around.

I did a little bit of research and found this one https://stackoverflow.com/questions/22159699/gradle-groovy-and-java-class-dependency-compile

So, the easiest way is to move the java tests to src/test/groovy I guess.

robertpanzer:
I think it's more common like that because groovy isn't that established yet as Java is. But my feeling is that in the future it will be much more common to call between multiple JVM languages in any direction back and forth and gradle should pay attention to it. (My personal opinion)

I posted a question to the gradle and groovy folks. 
But I think the answer will be same, move all Java code to the groovy directory.

abelsromero:
*A bit off-topic, but i felt like I could offer some light.*

> in the future it will be much more common to call between multiple JVM languages in any direction back and forth and gradle should pay attention to it. (My personal opinion)

I totally agree you that, for me this is more an issue with how gradle's Groovy plugin works.
Going from Java to Groovy (and even mixing them totally) is not so common, but not uncommon neither. This is something I do and is covered in books like `Making Java Groovy` or `Maven 3 CookBook`, I did some research and I am surprised how little documentation I could find.
Things I usually do are:
. Using groovy to implement Java interfaces when groovy offers an easy way (e.g. XML or file parsing).
. For testing purpouses/mocking like here.
In [maven](http://groovy.github.io/GMavenPlus/plugin-info.html) this is easy to do using "stubs" options, but not in gradle for what I've seen.

Back to the issue, I'd rather use the option proposed by @ancho ans just set both sourceSets as groovy. I really don't like the idea of moving the Java code to the groovy folder :%

regcole:
@robertpanzer Was this change ever released in any  `GA` version of AsciiDoctorJ?

PR #336, opened 2015-06-08
Fixes #334. Add support for defining the backend name of a converter …

### LINKED ISSUE #334: Add support for defining backend name for a converter in the converter class itself
Currently the backend name of a converter is only defined when it is registered at the converter registry: `asciidoctor.javaConverterRegistry().register(MyConverter.class, "mybackendname")`

This came up in the discussion in #332:

It should be possible to define at least one backend name in the converter class itself so that the converter can be registered without a backend name: `asciidoctor.javaConverterRegistry().register(MyConverter.class)`

Options are:
- Allow the converter writer to add an annotation to the `Converter` implementation like `@Backend("mybackendname")`
- Add a method `String getBackendName()` to the `Converter` class so that the converter writer is forced to define a backend name. (Even though the abstract base class `AbstractConverter` could return `null`.)

When no default backend name is defined by the converter `register(Class)` could throw an IllegalArgumentException.

robertpanzer:
…via an annotation @Backend.

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/converter/Backend.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/converter/Backend.java
@@ -1,0 +1,29 @@
+package org.asciidoctor.converter;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Assigns a default backend name to a converter so that it is not required to define
+ * the backend name at registration time.
+ *
+ * <p>Example:
+ * <pre>
+ * <code>&#64;Backend("my")
+ * public class MyConverter extends AbstractConverter {
+ *     ...
+ * }
+ *
+ * asciidoctor.javaConverterRegistry().register(MyConverter.class);
+ * </code>
+ * </pre>
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface Backend {
+
+    String value();
+
+}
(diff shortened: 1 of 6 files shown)


mojavelinux:
Annotations FTW!

I want to address one open question before we move forward with this. We inherited the term "backend" from AsciiDoc Python. However, it's not the best term to describe the purpose it serves. The long-term goal in Asciidoctor is to replace the usage of this term with one that is more fitting. Since we're defining a public API here, we should use it as an opportunity to migrate to the newer terminology.

On the Ruby side, I decided to name the DSL method `register_for`. I like this because it leaves the term implicit. Thus, on the Java side we might go with `@RegisterFor`. Another way of writing this is `@Handles` or `@Converts`. If you'd rather keep it as a noun, then I'd suggest `@Name` or `@ConverterName`. I'm open to other ideas.

I'm not sure which of those is the right choice, but for sure we should avoid `@Backend`.

mojavelinux:
I'm liking `@Converts(<name>)`. That reads really nicely.

```java
@Converts("leanpub")
class LeanpubConverter extends AbstractTextConverter {
    ...
}
```

robertpanzer:
What about @ConvertsTo?
I think we cannot get completely get rid of the term backend, as it's used in the options, attributes, etc. 
With the term @ConvertsTo the meaning of backend could be defined more precise as a target format instead of referring to the whole bundle of a of converter and a target. 
With only @Converts I'm somehow missing a preposition.

mojavelinux:
> I think we cannot get completely get rid of the term backend, as it's used in the options, attributes, etc.

We will eventually...and by that I mean have alternative terminology that maps to the same functionality (for backwards compatibility). In most of those places, the word "converter" will fit.

> What about @ConvertsTo? With only @Converts I'm somehow missing a preposition.

Good point. Another option is `@ConverterFor`, but I could get behind `@ConvertsTo` as well.

robertpanzer:
> Good point. Another option is @ConverterFor, but I could get behind @ConvertsTo as well.
> 
@ConverterFor is better. I've not seen verbs in annotation names yet.

mojavelinux:
> In most of those places, the word "converter" will fit.

Correction. I meant to say "format".

Issue #326, opened 2015-06-03
Add support for parsing child blocks


robertpanzer:
From http://discuss.asciidoctor.org/AsciidoctorJ-Block-Processor-Problems-tp3150p3297.html

Second, we need to provide a way to parse child blocks. We talked about adding this feature to the createBlock method. I also think it's reasonable to add a method to AbstractBlock named either `parseChildren()` or `parse(List<String>)`. In the first case, it would parse child blocks from the source lines passed to the block constructor. In the second case, it would parse blocks from an arbitrary set of lines and append them to the block as child nodes. I'm leaning more towards the second option.

robertpanzer:
Fixed with #337

PR #337, opened 2015-06-13
Fixes #325. Added method AbstractBlock.append(AbstractBlock) to appen…

### LINKED ISSUE #325: Add method AbstractBlock.append(AbstractNode) to add a child node to a block
See http://discuss.asciidoctor.org/AsciidoctorJ-Block-Processor-Problems-tp3150p3297.html:

As you can see from my previous message, we are missing two key methods in the AsciidoctorJ API.

First, we need to add the append method to the AbstractBlock interface. This method allows child blocks to be added to a block.

robertpanzer:
…d child blocks


robertpanzer:
Added `AbstractBlock.parse(List<String)` and it seems to work fine.
So this PR will also fix #326.

I wonder though, how to make extensions apply to the new blocks? :)
Should there be an option to make processors process the new content again, like block macros etc?
And if so how could I do it?
At the moment I am using this https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/parser.rb#L20-L22

mojavelinux:
> At the moment I am using this https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/parser.rb#L20-L22

OMG, don't reference that example. That is seriously outdated :)

This is what is relevant:

https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/extensions.rb#L112-L119

mojavelinux:
> I wonder though, how to make extensions apply to the new blocks? :)

That will happen automatically if the parser API is used (as it continues to use the same root document).

This gets back to what we were saying about having an option to createBlock that triggers the parser automatically. I actually think it should be in core, though you can do it in AsciidoctorJ before then and we can update the code in the future.

robertpanzer:
> OMG, don't reference that example. That is seriously outdated :)

Sorry ;-)
But I think it ends in the same code being executed.

> This is what is relevant:

> https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/extensions.rb#L112-L119

I can't invoke this method directly from within `AbstractBlockImpl` because it is an instance method of a processor. (I don't see the reason why it is bound to a processor instance though)

I must have done sth wrong yesterday, because now I see that a BlockMacroProcessor is being invoked. Will add a test case for that.

mojavelinux:
> But I think it ends in the same code being executed.

Correct. I was just expressing my embarrassment over my own docs. I really need to revise them. It's an open issue I've been meaning to fix.

> I can't invoke this method directly from within AbstractBlockImpl

Correct. That issue isn't yet public as I'm still deciding where to put it. You just need to implement what it does. I was just pointing out that the code in that method is more complete than what's in the API docs.

> I must have done sth wrong yesterday, because now I see that a BlockMacroProcessor is being invoked. 

Cool!

robertpanzer:
So you're ok to merge?

mojavelinux:
One small optimization. Before invoking `next_block`, verify that there are lines remaining using `has_more_lines?`. Otherwise, that looks good.

I'm not sure whether the name and placement of the `parse` method is right yet. I think I would rather it be that the (source) lines be passed to the blocks constructor and then the block can parse itself. Or, perhaps we need both scenarios.

1. The no-args `parse()` method parses the source lines currently held by the block.
2. As an alternative, lines can be passed into the `parse(List<String>)` method, which effectively replaces the source lines the block currently holds and parses them.

I also think it should be called `parseContent` instead of `parse` since what is being parsed is the content in the block, not the block itself.

That would make it more consistent with the way the Ruby side works.

wdyt?

robertpanzer:
I somehow didn't understand the issue correctly.;-)

>  I think I would rather it be that the (source) lines be passed to the blocks constructor and then the block can parse itself.

I am not sure if that really works out as an alternative.
At the moment, and that's what the `Processor#parse_content` is doing, is that it reads multiple blocks from the input.

This would no longer work if you let `AbstractBlock.parse()` parse the source lines of that block.
First, because the source lines could contain multiple blocks (throw an IllegalArgument/StateException in that case?)
Second the source lines could simply contain other Blocks, like a Section or a List instead of a Block.
(BTW we have no methods createList/createListItem yet, we will need that, don't we?)

So I would prefer going for option 1 plus moving the method `AbstractBlock.parseLines()` to the `Processor` so that it looks the same as in the Ruby world.

What do you think about that?

asciidoctorj-core/src/main/java/org/asciidoctor/ast/AbstractBlockImpl.java
```diff
@@ -81,4 +90,46 @@ public int getLevel() {
...
+        public AbstractBlock nextBlock() {
+            try {
+                IRubyObject nextBlock = getRubyProperty("next_block", reader, AbstractBlockImpl.this.getRubyObject());
+                if (nextBlock.isNil()) {
+                    return null;
+                } else {
+                    return (AbstractBlock) NodeConverter.createASTNode(nextBlock);
+                }
+            } catch (Exception e) {
+                e.printStackTrace();
```
robertpanzer: Oops, that certainly must be removed.
Maybe we could add checkstyle to the build so that things like that don't happen.

robertpanzer:
I think it's even problematic to implement option 1. (`AbstractBlock.parse()`)
The `Asciidoctor::Parser` API in fact creates new blocks.
Calling `AbstractBlock.parse()` would therefore require to set *all* members of the block, like style, roles, contentModel, etc.

Therefore I'd prefer to only implement `Processor.parse(List<String>)`.
It seems more logical to me to have a factory method for blocks instead of having an already existing block and replace its whole state with what the source contains.

Wdyt?

robertpanzer:
I certainly meant `Processor.parse(AbstractBlock parent, List<String> lines)`.
The new parent must be passed to the Parser API.

robertpanzer:
@mojavelinux Is this how you imagined the parse method?

robertpanzer:
The only thing that I find somehow strange about the `append()` and `parse()` methods is that you have to fall back to List operations if you want to do something else than appending content to the end of a block.
Say that you want to add raw asciidoctor content at the beginning of a section:
You would have to invoke `parse(section, content)` to make Asciidoctor parse the content and append it as last blocks to the section.
Now the extension must know somehow what blocks were added and move them in the `blocks` list via `remove()` and `add`.
Maybe this is very uncommon but it feels somehow strange.
At the moment `Parser::next_block` already sets the parent of the new nodes, so the parsed nodes are not completely "free".
But what about making the AsciidoctorJ `Processor.parse()` method return the nodes so that the API consumer can add them where he needs them to be?

```
section.getBlocks().addAll(0, parse(section, newContent));
```

(And possibly adding another method `Processor.append(parent, content)` that directly does the same as the Asciidoctor Ruby parse method does?) 
(Having said that a user coming from Asciidoctor Ruby would stumble over the changed behavior of the supposedly same method...)

robertpanzer:
@mojavelinux: Do you think this is the right way to do it? Should I merge it?

robertpanzer:
Rebased and renamed `parse` to `parseContent`.

mojavelinux:
@robertpanzer I think this looks good the way it is now. The method `parseContent` leaves room for the other APIs you suggested, but those ideas probably need a bit more time to simmer as we feel out the use cases. Merge away!

PR #338, opened 2015-06-15
Use enums instead of arbitrary strings to create nodes.

robertpanzer:
Also overloaded Processor.createBlock() with methods that don't require an attributes parameter.

This PR mainly addresses some internal details.
At the moment to create a node an arbitrary String is passed to the NodeConverter.
This is changed by defining an enum that defines all supported node types.
Another side effect is that the classes are now no longer retrieved via `Ruby.evalScriptlet()` which is, as I guess, faster and more robust.

The methods `createBlock` and `createInline` are also overloaded to work without an attributes parameter as this is often an empty HashMap anyway.

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/ast/NodeConverter.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/ast/NodeConverter.java
@@ -1,6 +1,8 @@
 package org.asciidoctor.ast;
 
 import org.jruby.Ruby;
+import org.jruby.RubyClass;
+import org.jruby.RubyModule;
 import org.jruby.java.proxies.RubyObjectHolderProxy;
 import org.jruby.runtime.builtin.IRubyObject;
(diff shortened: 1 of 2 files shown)


robertpanzer:
Strange, build failed on Appveyor after the 40 minute timeout although the build says it was successful...
And build time on Appveyor with Java 7 and on Travis were normal.

mojavelinux:
Looks like the time of the build itself is enough to make AsciidoctorJ fail the build.

> Build execution time has reached the maximum allowed time for your plan (40 minutes).

We should probably look into caching some of the artifacts to speed up the build some.

mojavelinux:
Very interesting idea, btw. That definitely makes the code a lot cleaner and more reliable.

mojavelinux:
Amazing how slow AppVeyor is in comparison to Travis. Travis is ~ 4 times as fast.

Issue #339, opened 2015-06-19
All Asciidoctor instances created with a Classloader share one extension registry


robertpanzer:
The test below fails, showing that multiple Asciidoctor instances created with a classloader use the same extension registry.
This is due to using ScriptingContainer to initialize the Ruby runtime, as this returns a per VM singleton instance.

```
class WhenTwoAsciidoctorInstancesAreCreated extends Specification {

    private static final String PARAGRAPH = 'paragraph'

    private static final String TEST_STRING = 'Hello World'

    def "then every Asciidoctor instance has its own extension registry"() {

        given:
        String document = '''= Test document

testmacro::Test[]
'''

        when:
        Asciidoctor asciidoctor1 = Asciidoctor.Factory.create(getClass().classLoader)
        Asciidoctor asciidoctor2 = Asciidoctor.Factory.create(getClass().classLoader)

        asciidoctor1.javaExtensionRegistry().blockMacro('testmacro', TestBlockMacroProcessor)

        then:
        asciidoctor1.convert(document, OptionsBuilder.options().headerFooter(false)).contains(TEST_STRING)
        // fails in the next line
        !asciidoctor2.convert(document, OptionsBuilder.options().headerFooter(false)).contains(TEST_STRING) 
    }


    static class TestBlockMacroProcessor extends BlockMacroProcessor {
        TestBlockMacroProcessor(String macroName) {
            super(macroName)
        }

        @Override
        Object process(AbstractBlock parent, String target, Map<String, Object> attributes) {
            createBlock(parent, PARAGRAPH, TEST_STRING)
        }
    }
}
```

robertpanzer:
Fixed with #340

PR #340, opened 2015-06-19
Use only one way to initialize a Ruby instance

robertpanzer:
This PR solves #339 
Instead of getting the Ruby instance from ScriptingContainer when passing a ClassLoader it is now always created via `JavaEmbedUtils.initialize()` and the classloader is passed to `RubyInstanceConfig.setLoader()`.

Personally I'd prefer to remove the factory method with a classloader completely, but we have to keep it because the ant plugin uses it, so I had to keep this method.
I think this way to initialize an Asciidoctor instance could be removed because:

1. Passing a classloader is equivalent to passing load paths (see https://github.com/jruby/jruby/wiki/ClasspathAndLoadPath#classpath-and-load-path-unification )
2. A new `RubyInstanceConfig` initially uses the current TCCL as classloader until it is reset (by `Asciidoctor.Factory.create(ClassLoader)` atm). So instead of passing the classloader to the factory method the ant plugin could also set the TCCL before creation of the instance and reset it afterwards again.

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/Asciidoctor.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/Asciidoctor.java
@@ -705,6 +705,8 @@ public interface Asciidoctor {
          * @param classloader
          * @return Asciidoctor instance which uses JRuby to wraps Asciidoctor
          *         Ruby calls.
+         * @deprecated Please use {@link #create()} and set the TCCL before or {@link #create(List)} passing the paths
+         * that you would have used to create the ClassLoader.
          */
         public static Asciidoctor create(ClassLoader classloader) {
             return JRubyAsciidoctor.create(classloader);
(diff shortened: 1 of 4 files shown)


robertpanzer:
One build failed again on appveyor due to a timeout...
I assume that the PR itself is working fine.

mojavelinux:
> we have to keep it because the ant plugin uses it

We could talk to @binout and see if he's willing to update the Ant plugin once AsciidoctorJ 1.6.0 is out so that you can drop the method. For now, marking it as deprecated is probably reasonable.

mojavelinux:
I like the consolidation.

robertpanzer:
If my theory is correct it should already work now to use the default create method and set the TCCL appropriately. 
I will try it out and try to make a PR:)

robertpanzer:
Strange, I just tested with the ant task and removed passing the class loader and even did not set the TCC.
Everything still seems to run fine!
Maybe @binout knows why it had to be passed.
@lordofthejars mentioned an SBT plugin in the comments, but I could not find but an empty placeholder project.

mojavelinux:
The SBT plugin still lives in @ktoso's repo. https://github.com/ktoso/asciidoctor-sbt-plugin

robertpanzer:
Created PR https://github.com/ktoso/asciidoctor-sbt-plugin/pull/2 for the SBT plugin that makes use of the TCCL approach.
I will deprecate `Asciidoctor.Factory.create(ClassLoader)`.

robertpanzer:
PR for SBT plugin got merged:)

mojavelinux:
:+1:

ktoso:
Thanks a lot Robert :-)

-- 
Konrad 'ktoso’ Malawski
Akka @ Typesafe
java.pl / geecon.org / krakowscala.pl / lambdakrk.pl

On 21 June 2015 at 09:26:37, Robert Panzer (notifications@github.com) wrote:

Merged #340.

binout:
In fact, in Ant task, the create method with classloader is needed.... But the tests doesn't show it... 
If you try a 'real' build.xml, you will have the following error : 
/Users/benoit/Coding/asciidoctor-ant-examples/build.xml:26: org.jruby.exceptions.RaiseException: (LoadError) no such file to load -- jruby/java

I will investigate for 1.6.0 to use other create methods with paths.

robertpanzer:
Oh, I think I tried exactly that. 
But I also realized some strange differences in the class loader behavior between tests and "real" execution. 
I'll try do retest again and try to provide a PR. 
With setting the TCCL appropriately before creating the Asciidoctor instance it should work. 

Having said that the current approach of passing the class loader to the create method will still work. 
I just marked it as deprecated because it overlaps with the load paths. 
I fixed the existing problem that could occur with multiple Asciidoctor instances anyway.

Issue #196, opened 2014-08-28
Introduce annotations for configuring an extension


mojavelinux:
Introduce annotations for configuring an extension. Currently, extension configuration must be passed to the super class in the extension constructor. However, this information is mostly static and best suited for annotations in the Java world.

Here's what is currently necessary (shown in Groovy):

```groovy
class YellBlock extends BlockProcessor {
    YellBlock(String name, Map<String, Object> config) {
        super(name, [contexts: [':paragraph'], content_model: ':simple']);
    }
    ...
}
```

Here's an example of how this might look (shown in Java):

```java
@Context(":paragraph")
@ContentModel(SIMPLE)
public class YellBlock extends BlockProcessor {
  ...
}
```

There should be annotations for each of the DSL methods in the Ruby APIs.

robertpanzer:
I looked through extensions.rb and I think that the Asciidoctor Ruby base classes currently support the config options below.
(I already translated them to dedicated annotations as in the initial approach)

* DocInfoProcessor has 
  * `@Location(String)`
    Values are `:header` and `:footer`.
    Would it make sense to define an enum for that or are other values conceivable?
* BlockProcessor & MacroProcessor & BlockMacro have
  * `@Name(String)`
  * `@Contexts(String[])`
  * `@ContentModel(String)`
  * `@PositionalAttributes(???)`
    * I could not find any values in the Asciidoctor sources...
  * `@DefaultAttributes(@DefaultAttribute[])`
    * `@DefaultAttribute(String, String)` (Does it work out to simply have Strings mapped to Strings here? Or do we need to support list etc as well?)

* InlineMacro has the same as MacroProcessor plus
  * `@UsingFormat(String)`
    * Values are `:short` and anything else (`:long`?).
       Would it make sense to define an enum for that or are other values conceivable?
  * `@Regexp(String)`

Do you think it is reasonable to create these annotations?
If so it should not be too hard to evaluate the annotations when someone registers a class at the JavaExtensionRegistry.

mojavelinux:
> `@Location(String)` Values are :header and :footer. Would it make sense to define an enum for that or are other values conceivable?

Not currently. Though, we have been considering an additional placement at the top of the content (not the head). Technically, a converter can request any placement key it wants when resolving the docinfo. However, for now probably best to start with an enum.

> `@PositionalAttributes(???)` I could not find any values in the Asciidoctor sources...

These are attribute names, so they must be string values.

[...]

PR #341, opened 2015-06-22
Added annotation based processor configuration

robertpanzer:
This should implement #196 

It is now possible to add these annotations:
- `@Name` to define block or macro names for BlockProcessors, BlockMacro- and InlineMacroProcessors
- `@ContentModel` to define the content model of a BlockProcessor
- `@Contexts` to define the contexts of a BlockProcessor
- `@DefaultAttribute` to define default attributes to pass to the process method of a processor
- `@DefaultAttributes` to define multiple `@DefaultAttribute`s on one type
- `@Location` to define where content generated by a DocinfoProcessor is to be placed.

`@PositionalAttributes` are missing yet because I am not aware what they should do.

In general the implementation adds options to the processors *class* object that serve as a default configuration for the instances. (see https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/extensions.rb#L84)

If this is a way to pursue docs and javadocs should be enhanced.
Also the javadocs of the new annotations should be extended with examples.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,7 +6,6 @@ clone_depth: 10
 environment:
   TERM: dumb
   matrix:
-    - JAVA_HOME: C:\Program Files\Java\jdk1.7.0
     - JAVA_HOME: C:\Program Files\Java\jdk1.8.0
 install:
   - SET PATH=%JAVA_HOME%\bin;%PATH%
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/extension/ContentModel.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/extension/ContentModel.java
@@ -8,6 +8,19 @@ import java.lang.annotation.Target;
 
 /**
  * This annotation defines how to handle content created by a {@code BlockProcessor}.
+ * Applicable for:
+ * <table>
+ * <tr><td>BlockMacroProcessor</td><td></td></tr>
+ * <tr><td>BlockProcessor</td><td>&#10003;</td></tr>
+ * <tr><td>BlockProcessor</td><td></td></tr>
+ * <tr><td>DocInfoProcessor</td><td></td></tr>
+ * <tr><td>IncludeProcessor</td><td></td></tr>
+ * <tr><td>InlineMacroProcessor</td><td></td></tr>
+ * <tr><td>Postprocessor</td><td></td></tr>
+ * <tr><td>Preprocessor</td><td></td></tr>
+ * <tr><td>Treeprocessor</td><td></td></tr>
+ * </table>
+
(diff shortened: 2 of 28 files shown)


lordofthejars:
:scream:

abelsromero:
I just skimmed through de code and it’s amazing, not only that, it’s 100% compatible with the current registration methods.

robertpanzer:
Thank you! :)
Staying compatible is imo a top priority.
But it is only possible because Alex has provided such an excellent test suite!

robertpanzer:
Appveyor failed again due to a timeout.
What about only testing one platform on Appveyor?

mojavelinux:
@robertpanzer, this is really amazing. Extensions for AsciidoctorJ just got a whole lot more awesome. I'm raising my glass to your work tonight.

> Staying compatible is imo a top priority.

:+1: Certainly a tough challenge, but since a bulk of the Asciidoctor user base goes through AsciidoctorJ, giving them a clear path forward is crucial.

> But it is only possible because Alex has provided such an excellent test suite!

:+1: :+1:

One for the test suite, and one for teamwork.

> @PositionalAttributes are missing yet because I am not aware what they should do.

Positional attributes map indexed positions in the attribute list to attribute names. For instance, the positional attributes for the image block macro are "alt", "width", "height" so you can write a shorthand version of the block macro, like `image::pic.jpg[Picture,100,200]`. Effectively, the block macro gets interpreted as `image::pic.jpg[alt=Picture, width=100, height=200]`. The value of the annotation would then be `@PositionalAttributes("alt", "width", "height")` Does that make sense?

mojavelinux:
> What about only testing one platform on Appveyor?

Yes, I think that is sufficient. Let's proceed with that change.

robertpanzer:
> Am 22.06.2015 um 21:43 schrieb Dan Allen <notifications@github.com>:
> 
> Does that make sense?

That makes sense. I will try to get that working tomorrow.

robertpanzer:
> Am 22.06.2015 um 21:43 schrieb Dan Allen <notifications@github.com>:
> 
> One for the test suite, and one for teamwork.

Hehe, just had a Pale 31...

lordofthejars:
@robertpanzer this is really awesome, my test suites are only this, test suites, necessary of course, but the real work is here and it works super

robertpanzer:
Added `@PositionalAttributes` annotation and added some Javadocs.

mojavelinux:
:+1: 

...and more :beers:

robertpanzer:
@mojavelinux We should have some real :beers: together soon again!

mojavelinux:
:+1:

PR #342, opened 2015-06-26
#296: Add support for overriding Ruby properties

robertpanzer:
This PR is a simple way to solve #296 
If the system property jruby.compat.version is set, AsciidoctorJ will not override this with the default value 2_0.
If such a system property is not set, the default 2_0 will be applied.

The CLI options are also extended to accept options like `-R compat.version=1.9` which result in setting the system property `jruby.compat.version=1.9`.

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/internal/JRubyAsciidoctor.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/internal/JRubyAsciidoctor.java
@@ -132,8 +132,12 @@ public class JRubyAsciidoctor implements Asciidoctor {
 
     private static RubyInstanceConfig createOptimizedConfiguration() {
         RubyInstanceConfig config = new RubyInstanceConfig();
-        config.setCompatVersion(CompatVersion.RUBY2_0);
-        config.setCompileMode(CompileMode.OFF);
+        if (System.getProperty("jruby.compat.version") == null) {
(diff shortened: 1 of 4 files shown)


asciidoctorj-core/src/main/java/org/asciidoctor/internal/JRubyAsciidoctor.java
```diff
@@ -132,8 +132,12 @@ private static void injectEnvironmentVariables(RubyInstanceConfig config, Map<St
...
     private static RubyInstanceConfig createOptimizedConfiguration() {
         RubyInstanceConfig config = new RubyInstanceConfig();
-        config.setCompatVersion(CompatVersion.RUBY2_0);
-        config.setCompileMode(CompileMode.OFF);
+        if (System.getProperty("jruby.compat.version") == null) {
+            config.setCompatVersion(CompatVersion.RUBY2_0);
+        }
+        if (System.getProperty("jruby.compile.mode") == null) {
+            config.setCompileMode(CompileMode.OFF);
+        }
```
mojavelinux: I think this needs to be the other way around. If the JRuby properties are not specified, we do nothing. I'm not concerned about breaking things because this can be controlled from the outside...and it's really not right that we are setting options without the user's control.

robertpanzer:
Will prepare a new PR that does not set any Ruby properties at all.

PR #343, opened 2015-06-27
Support block and macro processors with no arg constructors if name i…

robertpanzer:
…s defined via an annotation

As discussed in #196 this PR brings support for macro and block processors that only provide a no arg constructor if the name is defined via an annotation.

All processor proxies now use a more generic way to instantiate the processor by using the constructor with the most matching arguments, so this could also be used if we would support passing a configuration at processor registration time.
(JavaExtensionRegistry does not support passing a config yet, but I think Asciidoctor itself supports that.)


robertpanzer:
Force pushed to fix usage of Java8 only method Constructor.getParameterCount().

mojavelinux:
:+1:

PR #344, opened 2015-06-28
Removed default setting of any JRuby options

robertpanzer:
This is to make a start for #296 
This PR removes any setting of the ruby compat version or compile mode.
AsciidoctorJ will still run with the defaults and a user can now override the defaults if required.


mojavelinux:
:+1: 

We definitely want to make it very clear in the AsciidoctorJ manual (currently the README) about how to set these flags, especially the compile.mode flag, since it can help improve speed of operation.

robertpanzer:
I will update the Optimization section and prepare a new PR.

mojavelinux:
:+1:

PR #347, opened 2015-07-01
#296 Load JRuby properties from ./.jrubyrc and ~/.jrubyrc

robertpanzer:
One step further:
This PR parses the files `./.jrubyrc` and `~/.jrubyrc` when asciidoctorj is started via the CLI.
(IMO integrators should set these properties themselves, either use the same method or define own properties.)

Fortunately JRuby already provides a public static method to do this in `org.jruby.Main` that is simply used to accomplish this.

Additionally extended the version output (-V) to show the JRuby version and RUBY_VERSION.

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/cli/AsciidoctorInvoker.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/cli/AsciidoctorInvoker.java
@@ -40,8 +40,8 @@ public class AsciidoctorInvoker {
             
             if (asciidoctorCliOptions.isVersion()) {
                 System.out.println("Asciidoctor " + asciidoctor.asciidoctorVersion() + " [http://asciidoctor.org]");
-                System.out.println("JRuby : " + JRubyRuntimeContext.get(asciidoctor).evalScriptlet("JRUBY_VERSION"));
-                System.out.println("Ruby compat version: " + JRubyRuntimeContext.get(asciidoctor).evalScriptlet("RUBY_VERSION"));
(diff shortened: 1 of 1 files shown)


mojavelinux:
Very nice! This looks like it might be the most convenient way for users to set these values.

I would prefer if the output of `-V` fit more with the new output being used in Asciidoctor 1.5.3. Here's how it now looks:

```
Asciidoctor 1.5.3.dev [http://asciidoctor.org]
Runtime Environment (ruby 2.2.1p85 (2015-02-26 revision 49769) [x86_64-linux]) (lc:UTF-8 fs:UTF-8 in:- ex:UTF-8)
```

I'm thinking perhaps it could look like this:

```
AsciidoctorJ 1.5.3 [http://asciidoctor.org]
Runtime Environment: jruby 1.7.19 (1.9.3p551)
```

Notice I changed "Asciidoctor" to "AsciidoctorJ".

The template for the Runtime Environment value is:

```
jruby JRUBY_VERSION (RUBY_VERSION)
```

Sound good?

mojavelinux:
I propose this template because it is consistent with the output of `jruby -v` (without all the trailing superfluous info).

robertpanzer:
Sounds good!
Changed the output to look like this now:

```
Asciidoctor 1.5.2 [http://asciidoctor.org]
Runtime Environment: jruby 1.7.20 (1.9.3)
```

mojavelinux:
That looks wonderful.

robertpanzer:
:beers:

robertpanzer:
Oh, forgot the J in Asciidoctor*J*.
Pushed c3b77ebded735123cdcc3cebcec094b3c919cb41

PR #348, opened 2015-07-02
Updated description of options for the JRuby runtime

robertpanzer:
Describe that the JRuby runtime can be configured via `jruby.*` system properties or `.jrubyrc` files when launching via the shell script.


mojavelinux:
Excellent! I like the write-up and the options. Very clear.

PR #349, opened 2015-07-03
Readded support for processor constructors with a trailing map parameter

robertpanzer:
With changing the way processor classes get instantiated by AsciidoctorJ I removed the last map argument to the constructor call because there was simply no value in it.
AsciidoctorJ could only pass an empty map.

But removing this parameter would make existing processors incompatible.

So I readded the trailing map parameter again, always passing in an empty HashMap.

For a BlockProcessor registered as a class AsciidoctorJ will now call the constructors in this order:
1. String macroName, Map config
2. String macroName
3. no args

The first constructor that exists will be called.

The same mechanism applies for all other extensions.


mojavelinux:
:+1:

PR #352, opened 2015-07-07
Source location

### LINKED ISSUE #351: Asciidoctorj-1.6.0 support for source_location
`AbstractBlock` is missing `source_location` (I guess it should eb called `getSourceLocation` in Java). Although these values are only set when `sourcemap` is passed as an option, the current interface does not allow access to this asciidoctor-ruby attribute at all.

robertpanzer:
This PR resolves #351 
I added a method `AbstractBlock.getSourceLocation()` that returns a `org.asciidoctor.ast.Cursor`.

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/internal/JRubyAsciidoctor.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/internal/JRubyAsciidoctor.java
@@ -18,12 +18,9 @@ import org.asciidoctor.converter.internal.ConverterRegistryExecutor;
 import org.asciidoctor.extension.JavaExtensionRegistry;
 import org.asciidoctor.extension.RubyExtensionRegistry;
 import org.asciidoctor.extension.internal.ExtensionRegistryExecutor;
-import org.jruby.CompatVersion;
 import org.jruby.Ruby;
 import org.jruby.RubyHash;
 import org.jruby.RubyInstanceConfig;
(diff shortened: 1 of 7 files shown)


asciidoctorj-core/src/main/java/org/asciidoctor/internal/JRubyAsciidoctor.java
```diff
@@ -567,7 +564,7 @@ public Document load(String content, Map<String, Object> options) {
     @Override
     public Document loadFile(File file, Map<String, Object> options) {
         RubyHash rubyHash = RubyHashUtil.convertMapToRubyHashWithSymbols(rubyRuntime, options);
-        return (Document) NodeConverter.createASTNode(this.asciidoctorModule.load(file.getAbsolutePath(), rubyHash));
+        return (Document) NodeConverter.createASTNode(this.asciidoctorModule.load_file(file.getAbsolutePath(), rubyHash));
```
robertpanzer: I had to change this. Otherwise the Cursor will show <stdin> as file.

ysb33r:
From what I can gather the interface looks good by me.

asciidoctorj-core/src/main/java/org/asciidoctor/internal/JRubyAsciidoctor.java
mojavelinux: Seems more correct anyway.

PR #353, opened 2015-07-08
Renamed interface DocumentRuby to Document and the class Document to …

robertpanzer:
…DocumentImpl

This PR renames the interface `org.asciidoctor.ast.DocumentRuby` to `Document` and the current class `Document` to `DocumentImpl`.
The names are simply more logical.

Additionally users who have written extensions with 1.5.2 will see an increased compatibility because the Processor methods of a processor like a `Preprocessor` consumed the class `org.asciidoctor.ast.Document`.


lordofthejars:
You are my hero!!! Sorry for given these names at first, I had something in mind initially about multiple engines running Ruby code. But this has a lot of more sense.

robertpanzer:
No worries, Alex.
Thank you Alex!
But it was actually my fault. Today I understood your decision: `DocumentRuby` was only made to serve internally as the Ruby proxy, the purpose of the class `Document` was to form the API.
As I changed the way Ruby objects are proxied in Java I built on top of `DocumentRuby` instead of changing the class `Document` to an interface because it's a pattern for me to only present interfaces to the consumer of the API.
So it was really more my fault.

robertpanzer:
@ysb33r I fear that this will break the leanpub-converter once this gets merged.

ysb33r:
@robertpanzer Thanks for telling me, but please go ahead.

mojavelinux:
I have been looking forward to this pull request for a long time. There's just something very gratifying about having clean API names like `Document`. Not only is it a more logical name, it's also more portable as we make changes in the future because, regardless of what a `Document` can do, at least we know that a document is a document ;)

There's no reason to assign fault. We just focus on being better each day than the previous one and we'll always be improving! That's the spirit.

mojavelinux:
Once this is merged, we can probably keep an eye out for artifacts that are remnants of the old hierarchy. For instance, I see at least one explicit cast from Document to AbstractBlock. Now that we have the correct hierarchy, casts like that should no longer be necessary.

robertpanzer:
Now that this is merged: How should we proceed with the renaming of the AST classes?
At the moment I think we have reached a state where we are compatible again with 1.5.2. Both for extensions as well as for most API users. (I still have to check this by getting the test extensions from 1.5.2 and using them against 1.6.0-SNAPSHOT.)

Should we continue and go for this?
http://discuss.asciidoctor.org/dev-Reorganization-of-the-AsciidoctorJ-AST-data-structures-tp2991p2993.html
That is:
```
Node
  +- BlockNode
  |   +- Block
  |   +- Section
  |   +- Document
  |   +- List
  |   +- ListItem
  +- Inline
```

lordofthejars:
Yeah there is no faults, this is open source :) there is only a community that wants to do the things as better as possible.

PR #355, opened 2015-07-10
Fixes #354. Add support for defining outfile suffix of a converter

### LINKED ISSUE #354: Allow to set outfilesuffix for a Java converter
Currently Java converters always create files with the extension .html

It should be possible to change this.

Options are:
- Allow to define the suffix via an annotation. If so, add another parameter to the `@ConverterFor` annotation? (Would not read so nicely anymore, change the name to sth else?)
- Add a method `Converter.setOutfilesuffix()` that must be called in the constructor of the converter if it should change.

robertpanzer:
The suffix for the generated files can now be defined via `@ConverterFor(suffix=".md")` or via `Converter.setOutfileSuffix(String)` in the constructor of a converter.

I changed the signature of the constructor of a converter to accept a `Map<String, Object>` instead of a `Map<Object, Object>`.
While this might introduce incompatibilities with current implementations on 1.6.0-SNAPSHOT (Sorry, @ysb33r) this aligns with the signature of the processors.


ysb33r:
Maybe lisghtly off-topic, is there anyway of telling the Asciidoctor not to generate output at all i.e. equivalent of sending everything to `/dev/null`. This will be quite useful in the interim for soemthing like lenapub which handles output of files on it's own.

robertpanzer:
Test failed due to some inconsistency with asciidoctor 1.5.3.dev.
Test runs with asciidoctor 1.5.2

robertpanzer:
Apparently the order of the classes of the div tag has changed.

robertpanzer:
I played with `/dev/null` by setting the option `to_file` to `/dev/null`.
But it seems to me as if the converter is not being called in that case.
Instead the document is simply loaded:
https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor.rb#L1432-L1433

robertpanzer:
You can set `to_file` to `false`.
This will prevent writing a file:
```
OptionsBuilder.options().option("to_file", false)
```
It would probably make sense to add an option to the converter for that.
(Even though I think that it makes sense to create the root content in the file the user has defined, even when other, additional files are being generated by the converter)

ysb33r:
> You can set to_file to false.

I'll give that a go.

> It would probably make sense to add an option to the converter for that.

We sort of discussed it elsewhere. It would have been more ideal for a converter to provide its own stream. In this case providing a `NullStream` would have worked.
> (Even though I think that it makes sense to create the root content in the file the user has defined, even when other, additional files are being generated by the converter)

It does not make sense in the case of Leanpub, where a single large file would not work, as it would nto be possible to control which chapters end up in the full book and which go in the sample book.

mojavelinux:
> Apparently the order of the classes of the div tag has changed.

Oops! I'll fix that.

mojavelinux:
Fixed!

mojavelinux:
> It would have been more ideal for a converter to provide its own stream. In this case providing a NullStream would have worked.

The way I've worked around this in the past (such as the epub3 converter) is that I decide which file is the primary output and I let the Writer write something to that file. Then I write the remainder to other files. In other words, I give the Writer something to do since I know I'll be writing to _at least_ one file. Then, the converter can additionally write to other files. I admit it's a hack, but that's the workaround to avoid having an empty file for now.

mojavelinux:
> I played with /dev/null by setting the option to_file to /dev/null.
But it seems to me as if the converter is not being called in that case.

Correct, this is an optimization.

The other approach is to set to_file to a stream like StringIO. I'm assuming the equivalent in the Java side (if it works) is a StringBuffer.

robertpanzer:
I'll prepare a PR for @ysb33r to align the leanpub converter and merge this afterwards. 
Then I'll look again for a streaming approach, it just didn't make "click" for me yet.

Issue #304, opened 2015-05-12
Asciidoc rendering to pdf stream


johncarl81:
Today we attempted to write a asciidoc -> pdf stream in a web application.  @shannong and I ended up with the following solution:
```java
String output = Asciidoctor.Factory.create()
    .convertFile(new File("input.asciidoc"), 
        OptionsBuilder.options().backend("pdf"));
```

This results in a pdf file (`input.pdf`) being generated in the root of the project.  What I would expect from this API is output would either include the contents of the pdf, or something similar.  What is returned is actually `null` because the following adapter code in `JRubyAsciidoctor`:
```java
    private String returnExpectedValue(Object object) {
        return object instanceof String?object.toString():null;
    }
```

What we need is a stream from Asciidoctor containing the pdf so we can stream it out to the user.  I'd prefer is this was contained in memory and not have to use the disk as a buffer.  Can we change the api to accommodate?

mojavelinux:
While I'm sure this is not yet supported, your hypothesis is slightly incorrect. You first need to specify that you don't want to write to a file:

```java
Object result = Asciidoctor.Factory.create()
    .convertFile(new File("input.asciidoc"), 
        OptionsBuilder.options().backend("pdf").toFile(false));
```

This is where the problem comes in. When `toFile` is set to `false`, the result depends on the converter. At the moment, the PDF converter will return the `Converter` instance, which is also a Prawn `Document` object. If you manage to get a handle on that object, you can call `render` to get the PDF as a String.

The return value of `convert` and `convertFile` is being tracked in #234.

mojavelinux:
This is where we need to allow the type to be coerced in a type-safe way as described in #234. Of course, that argument needs to be optional.

I'd really like to get feedback from the community of Java API designers on this one because this is a tricky one to get right and we want to make the API better, not more complex.

johncarl81:
I commented on this in #234... I'm very nieve to the ruby code behind Asciidoctorj, but would returning an `OutputStream` make sense here?  You could easily map a String to an `OutputStream` and it would accommodate pretty much any other binary format.

mojavelinux:
That's very much a possibility.

johncarl81:
For our use case that would work beautifully, especially if it is backed by memory and not the file system.

johncarl81:
@robertpanzer, looking over #356, should we be able to write a pdf to a byte array using a registered `Converter`?

robertpanzer:
It should be possible.
[...]

Issue #333, opened 2015-06-05
Refactor README document into one or more manuals


mojavelinux:
The README is no longer a README :) It has matured into a full-blown manual (> 64K). I think it's time to setup a docs folder and refactor the README into one or more manuals.

The README should be simplified to a simple introduction, basic setup procedures (“getting started”) and some general information about the project (license, mailing list, etc).

Ideally, we should have a file for each major topic. For instance, the instructions for creating a converter would be it's own file. Same for how to create an extension.

We can use this issue to outline the files and sections.

mojavelinux:
There are several distinct user types, so perhaps several manuals.

* There are people that want to use AsciidoctorJ as a replacement for Asciidoctor (interested primarily in the distribution)
* There are people that want to embed AsciidoctorJ into their application using the API to invoke Asciidoctor (integrators)
* There are people that want to extend AsciiDoc and the Asciidoctor processor to give it more functionality (extension writers)

Where a person creating a custom converter falls is a bit fuzzy because they might be either an integrator doing customization or an extension writer. Hard to say.

lordofthejars:
:+1:

robertpanzer:
> Where a person creating a custom converter falls is a bit fuzzy because they might be either an integrator doing customization or an extension writer. Hard to say.

I think that's true but not a problem. 
It's like having 4 chapters for three different readers. 
Reader A is happy with chapter 1, B with 1, 2 & 4 and C with 1, 3 & 4.

robertpanzer:
What do you think about writing dedicated manuals and thereby reduce the README. 
That means provide a separate documentation like:
- Writing converters for custom formats
- Writing custom extensions
And replace the content in the README with links to these documents?

Should the documents be located under asciidoctorj/docs?

mojavelinux:
> What do you think about writing dedicated manuals

Agree. I think we should plan to write individual documents as though they were chapters in a user manual. The more focused, the better. We can then produce a complete manual by combining these chapters together.

> Should the documents be located under asciidoctorj/docs?

Yes. We don't yet have a defined structure yet across projects, so AsciidoctorJ will lead the way. There's no doubt that "docs" is clear and easy to find...and that's requirement 1.

robertpanzer:
Integrators Guide created as part of #361

PR #356, opened 2015-07-12
Added support for converters that create arbitrary content that write…

robertpanzer:
… it to the file itself.

This PR is far from finished.
It wants to work on the multitude of requirements for converters that create arbitrary content other than Strings (see #234 and #304).
Having said that this PR doesn't try to go for the requirement for multiple streams.

This commit adds the interface `WritingConverter`.
Letting a `Converter` also implement this interface mimics an Asciidoctor Ruby converter that includes the `Asciidoctor::Writer` module.
That is the converter writes the content itself to the target file.
(Writing to streams is still making my brain hurt at the moment, because the current implementation in AsciidoctorJ for converting to a Writer is outside of Asciidoctor/Ruby, that is the document is converted to a string and that is written to the Writer)

This is also the precondition for allowing a `Converter` to create non-stringy content.
Therefore the `Converter` interface is now a generic interface that has to declare the results it produces.
Such a converter that also wants to support writing its result to files should therefore also be a `WritingConverter` because the converter is the only one who knows how to serialize the result.

The `Asciidoctor` interface is extended with several generic convert & convertFile methods that take an expected result type so that the caller can receive arbitrary object content instead of only strings.
The existing convert methods delegate to these new methods with a expected result type of String.
If the result type does not match the actual result a `ClassCastException` will be thrown.


robertpanzer:
Btw: I am pretty fine with refusing this PR. 
I just hope that a discussion around this subject brings us closer to a good solution. 
Because atm I have the impression that it's really difficult to find a good solution that satisfies all expectations.

robertpanzer:
Maybe it would also make sense to make a converter that converts to Strings is a specialization of a converter with arbitrary results and provide a default implementation for `write` with String outputs.

robertpanzer:
Just changed the interface `WritingConverter` to receive an `java.io.OutputStream` instead of a `java.io.File`.
The stream is defined via `OptionsBuilder.toStream()`. 
For `OptionsBuilder.toFile()` the `ConverterProxy` opens and closes a `FileOutputStream`.
Thereby it is now possible to create streaming converters in Java.
A wrapping converter for the existing Ruby PDF backend could implement the desired PDF streaming for PDF documents.

For the current Ruby based converters this would certainly not work out of the box.
(But maybe that's even possible with setting the to_file option to sth that implements `:write`.

robertpanzer:
One build failed on Travis because TorqueBox responded with HTTP 500 when trying to fetch the Asciidoctor 1.5.3.dev gem. The other builds succeeded.

mojavelinux:
I'm thrilled that you are starting to tackle this problem. It's a tough challenge, but has a very rewarding result because it frees us from the String-only world.

I'm not sure the WritingConverter is the right name for this type. I recognize that Writer may be too generic, so how about OutputWriter? I say that because these are orthogonal concerns. One converts from AsciiDoc to another object. The other writes that object somewhere. I suppose we could have a base class that mixes the two, but it seems better if we just require that the converter class extend the OutputWriter interface. I could also get behind `OutputFormatWriter`.

mojavelinux:
In general, I think you are well on your way there.

> Thereby it is now possible to create streaming converters in Java.

I like this. Big step forward.

What I can't tell is whether converters can leave off the type and still work with a String result. Is it necessary to specify? If so, we might want to think about having a StringConverter as a base class that fills in this type for convenience. wdyt?

I'm very interested to know what @ysb33r thinks about this new contract. This could really bring the AsciidoctorJ API the next level of maturity.

robertpanzer:
> I'm not sure `WritingConverter` is the right name for this type.

I'm always bad at naming things :)

> I could also get behind `OutputFormatWriter`.

I like that name. At least the `Writer` cannot be completely decoupled from the Converter (Asciidoctor Ruby requires that) so I'd like to keep a link.
So `abstract class AbstractConverter implements Converter, OutputFormatWriter`?

> If so, we might want to think about having a StringConverter as a base class that fills in this type for convenience.

Good idea!
For someone who wants to get into this technology it will be much simpler to start with a `StringConverter`.

> it seems better if we just require that the converter class extend the OutputWriter interface.

Yes! And the abstract `StringConverter` could provide an implementation for `write`.
Thereby we would already get a streaming StringConverter for free.
And the `write` method would not surface before the user wants to go one step further and implement a binary converter.

mojavelinux:
> I'm always bad at naming things :)

It's the hardest thing in programming. A community of minds is the only weapon we have against it :)

mojavelinux:
> abstract class AbstractConverter implements Converter, OutputFormatWriter?

Precisely what I had in mind.

mojavelinux:
> For someone who wants to get into this technology it will be much simpler to start with a StringConverter.

Exactly. This is the first step and likely the tutorial will introduce the first-time converter author in this way.

ysb33r:
> abstract class AbstractConverter implements Converter, OutputFormatWriter?

I like that

robertpanzer:
Renamed `WritingConverter` to `OutputFormatWriter`.
Require every Java based Converter to be an `OutputFormatWriter`.
Introduced `StringConverter` for simple string based converters.

ysb33r:
What is still not clear for me is at what stage the `write` will be called. Is the expectation that asciidoctor core will invoke this after processing the AST?

robertpanzer:
Well, it works fully integrated with Asciidoctor Ruby:

First `convert()` will be called. This returns a result, of type String or anything else.

If the converted result should _not_ be written to a file this result is returned to the caller.

If the converted result should be written to a file and the converter mixed in the `Writer` module the method `write()` is invoked passing in the converted result and the stream.
If the converter does not include the `Writer` module Asciidoctor simply writes the content to the file (No stream!) hoping that Rubys `::File` knows how to handle that.

With this PR every Java based converter includes the `Writer` module!

So actually the purpose of the `Writer` module was never to invoke it from the `convert` method as we had guessed some time before but to be called after conversion by Asciidoctor to write the generated content somewhere.
(That's still sth that has to be mentioned, we write to streams but there is no real streaming where we can read very large documents and write them somewhere while rendering, it's always parse, convert, write.)

When writing to a file Asciidoctor Ruby returns the parsed `Document`, but in AsciidoctorJ we return `null` because we have no union types or sth like that to return alternatives depending on the options.

Did it become clearer now?
We definitely have to explain that very carefully in the manuals.

mojavelinux:
> actually the purpose of the Writer module was never to invoke it from the
> convert method as we had guessed some time before but to be called after
> conversion by Asciidoctor to write the generated content somewhere.
>

Spot on!

robertpanzer:
Fixed an NPE with string converters that return `null` from `convert` like the leanpub converter.
Aligned the README with the changes to the converters.
But I feel we have to go for providing dedicated documentation soon. I could have added much much more to the README which would let it explode.

mojavelinux:
>  I could have added much much more to the README which would let it explode.

It's probably the right time to start working on the layout of the AsciidoctorJ manual. See https://github.com/asciidoctor/asciidoctorj/issues/333.

MonaElsa:
Hi everyone, 
what is the status on this topic. I need a way to convert to a pdf stream.

Issue #357, opened 2015-07-13
automatically resolved gems in OSGi bundles need .jrubydir files 


ginseil:
Hi,

I was playing around with asciiidoctorj in an OSGi environment within the Eclipse BNDTools toolcahin.
* Eclipse (at least Luna should work)
* BNDTools (install via market place)
* jruby 9.0.0.0rc2 & jcommander 1.48 (both are valid OSGi bundles and are managed as OSGI bundle dependencies by BNDTools)

Because asciidoctor itself is not an OSGi bundle I unroled it's contents into one unsing BNDTools.
Each time I tried to create an Asciidoctor instance I got some class / module loading exceptions.

Yes it is possible to manually reference the required gems but wouldn't it be nice if they were automatically resolved? See pros & cons in your README.

I took a look into the jruby.jar and found a loot of ".jrubydir" files pointing to files located next to them (embedded gems need directory file .jrubydir). So I just tried to manually add those files into my unrolled asciidoctor bundle and it worked like a charm.

The only code needed is:

		OSGiScriptingContainer container = new OSGiScriptingContainer(context.getBundle());
		Asciidoctor asciidoctor = Factory.create(container.getOSGiBundleClassLoader());
		System.out.println(asciidoctor.asciidoctorVersion()); // --> prints 1.5.2

The unrolled contents in my asciidoctor bundle:

![image](https://cloud.githubusercontent.com/assets/7522976/8648483/451ed550-2961-11e5-8293-1cab3b87dc39.png)

lefou:
Might be related to #297

mojavelinux:
@lefou indeed. @ginseil could you try a SNAPSHOT (you need to build it yourself). We know at least have some OSGi information in the AsciidoctorJ jar. We are trying to determine what is still missing so we can fill in the voids.

lefou:
@ginseil In master, asciidoctor is a bundle nowadays, but we are still in the process of figuring out, what is required to provide a fully functional Asciidoctor service via the OSGi service registry. Your setup might help with this.

I don't use Eclipse BNDTools myself, so it's a bit hard to understand, what's behind your screenshots. Could you please post the content of your `bnd.bnd` and also the `.jrubydir` files?

ginseil:
@mojavelinux I tried to build the SNAPSHOT on my own. The result is I did not work with jruby 1.7.20. I always got the following exception, even if I explicitly point to the asciidoctor gem:

```java
[...]

PR #358, opened 2015-07-14
WIP: Register an Asciidoctor service into the OSGi registry when activated

lefou:
This PR is work in progress. It works when executed with JRuby 9.0.0.0.rc2 but not with JRuby 1.7. It adds an OSGi bundle activator. When the bundle becomes active, it creates an `Asciidoctor` instance and registers it under the `Asciidoctor` interface into the OSGi registry. When stopped, the service will be unregistered and it's `shutdown` method is invoked. In resolved (but not activated) state, the bundle does not create any instances nor JRuby sessions.

This PR is based on the info from #357 and should provide the next step of #297. The newly added files contain hard coded version numbers. Feedback or PRs to make these dynamic are very welcome. Also I have not tested if both `.jrubydir` files are required or only one of them.

Review by @mojavelinux and @ginseil

[Review on Reviewable](https://reviewable.io/reviews/asciidoctor/asciidoctorj/358)


ginseil:
As far as i know from my testing both .jrubydir files are required. If one of them is missing you'll get an exception indication that either `./.jrubydir` or `./specifications/.jrubydir` could not be found.

robertpanzer:
Regarding the failed build on Travis you might want to rebase at least on a6598bc

lefou:
@robertpanzer This PR is against master, not the asciidoctorj-1.6.0 branch. Isn't master used for development anymore?

robertpanzer:
On Travis AsciidoctorJ is first built against Asciidoctor Ruby! 1.5.2 like you do locally.
Afterwards the build pulls the current Asciidoctor 1.5.3.dev and builds against that again.
You can do it yourself by executing the script `test-asciidoctor-upstream.sh`

asciidoctorj-core/src/main/resources/specifications/.jrubydir
```diff
@@ -0,0 +1,11 @@
...
+.
+asciidoctor-1.5.2.gemspec
+coderay-1.1.0.gemspec
+erubis-2.7.0.gemspec
+haml-4.0.5.gemspec
+open-uri-cached-0.0.5.gemspec
+slim-2.0.3.gemspec
+temple-0.6.10.gemspec
+thread_safe-0.3.4-java.gemspec
+tilt-2.0.1.gemspec
```
robertpanzer: Personally I'd prefer to generate this file from the build script.
In particular without a complete OSGI integration test suite it will be very probable that we miss one name in the future.

robertpanzer:
Ah ok, don't rebase on the commit I mentioned ;-)
I'll cherry-pick that onto master.

robertpanzer:
Just pushed the fix to master: 1523a4a2233e414a5d1009122dbdd939973514f4
You can rebase on that to fix the travis build.

ginseil:
@lefou The Activator implementation is a good starting point. It provides a global Asciidoctor OSGi service (Alternatively the Asciidoctor service could be registered using the OSGi Declarative Services approach - would allow multiple and / or configurable instances...). 

One disadvantage is that any extension provider has to lookup the Asciidoctor service to be able to add any desired extension. This indeed could be improved by the Whiteboard pattern mentioned in [#297](https://github.com/asciidoctor/asciidoctorj/issues/297). For every extension registry (JavaExtensionRegistry, RubyExtensionRegistry and JavaConverterRegistry) a dedicated service tracker could listen for appropriate extension services and dynamically add (remove - only all or nothing?!) them.

asciidoctorj-core/src/main/resources/specifications/.jrubydir
lefou: Yeah, agreed, as said in the first comment. I will change that as soon as I figured out how, if nobody is faster than me. ;-)

lefou:
@ginseil I would prefer to not depend on DS (to avoid the dependency and also to provide more flexibility for our users, e.g. I prefer https://github.com/domino-osgi/domino over DS or Blueprint.)

To use the whiteboard pattern is indeed the next step. As soon we can verify, the current solution works as expected, we can enhance it. I havn't used the extension mechanism, so I'm not familiar with the requirements these could possibly have. Implementing the whiteboard pattern without DS is easy. Of course, any registered extension would be available for each consumer. If that's is not what most users need, we should discus alternatives. IMHO, it should be fine for most people.

The other question is how we handle registrations done directly via the `Asciidoctor` interface in addition to the white board?

asciidoctorj-core/src/main/resources/specifications/.jrubydir
mojavelinux: Totally agreed. I'm pretty sure that Gradle collects all the gem file versions, so it's just a matter of following up with a post-resolution step that injects these versions into the file. But I realize the challenge is in the doing :)

mojavelinux:
>  It works when executed with JRuby 9.0.0.0.rc2 but not with JRuby 1.7.

I think it's very safe to say that if you want to use AsciidoctorJ as an OSGi bundle, you have to use JRuby 9.0.0.0. No need trying to fight an uphill battle against a JRuby version that will soon be deprecated (which will be swift once 9.0.0.0 is final).

lefou:
Do you see any need for the single `Asciidoctor` instance to be configured. And if, should we react on configuration changes, too? In both cases, I would go the [Configuration Admin](http://wiki.osgi.org/wiki/Configuration_Admin) route.

asciidoctorj-core/src/main/resources/specifications/.jrubydir
lefou: Ok, now I generate this file with gradle.

mojavelinux: :+1:

robertpanzer:
Before officially moving to JRuby 9.0.0.0 we will have to check how the LoadService has changed though.
With JRuby 1.7.x resources were loaded by default using the TCCL, which obviously doesn't play well with OSGi, even though the AsciidoctorActivator passes in an OSGi class loader (I don't understand atm what kind of classloader this is, still so much to learn ;-) )
But at least there was one negative test that is now passing, which seems nice at first look, but it shows that things work differently than how we thought they work. So I'll try to learn how this works in the next weeks.

One question I have about Asciidoctor in OSGi is if we need another way of publishing, finding and registering converters and extensions from other bundles.
I just wildly guess that ServiceLoader does not work transparently over bundles but that there is some kind of ExtensionPoint/Extension thing as I have used a time ago while writing Eclipse plugins.

robertpanzer:
You can find the docs for the arquillian OSGi container at https://github.com/arquillian/arquillian-container-osgi. This is a scarce though given the many implementations it supports.
If you want to have a real book the "Arquillian Testing Guide" by John Ament also covers Arquillian & OSGi.
And if you have some time there will surely be another excellent book coming soon.

lefou:
> One question I have about Asciidoctor in OSGi is if we need another way of publishing, finding and registering converters and extensions from other bundles.

I'm in favour of doing it the OSGi way, by implementing the [Whiteboad Pattern](http://wiki.osgi.org/wiki/Whiteboard_Pattern). In this case, it means, a bundle that provides some extentions to Asciidoctor, e.g. a converter, will register the converter instance into the OSGi service registry. The Asciidoctor bundle will react on such a registration event by using the converter. That pattern has many advantages, e.g. loose coupling and it takes the independent life-cycles (of the Asciidoctor bundle and the extention bundles) into account. But, there are some limitations in the current Asciidoctor API, namely #359: there is no way to unregister a single extension, only all at once.

robertpanzer:
Seems like we have to give up on Java 6 compatibility with JRuby 9.0.0.0!

mojavelinux:
> Seems like we have to give up on Java 6 compatibility with JRuby 9.0.0.0!

I think it's time.

robertpanzer:
@mojavelinux: Do you think it would make more sense to rebase this onto the asciidoctorj-1.6.0 branch? I am unsure whether we already want to got for JRuby 9000 and Java 7 with an AsciidoctorJ 1.5.3 release, considering that we only change the last part of the version number.

mojavelinux:
Absolutely. Now that AsciidoctorJ 1.6.0 is a well defined target, and the
recommended path forward, this change definitely fits there.

lefou:
Ok, I can create a new PR against asciidoctorj-1.6.0 branch. But why isn't main development not happening in master? Given the fact, that there is already a asciidoctorj-1.5.x branch. What's then in master?

robertpanzer:
Yes, this is probably not so nice.
We decided a time ago to drive the development for 1.6.0 on a separate branch because much of the internals changed without knowing if these changes really work out, in particular concerning extensions.
As we were a bit unsure how the development continues we also created branch for 1.5.x and a 1.5.2.1 hotfix.
Now reality has shown, that the development on 1.6.0 seems to work out, but it will take some time until we also have an Asciidoctor Core 1.6.0. (AsciidoctorJ version numbers should match the version number of core.)
Regarding the 1.5.x releases PRs certainly appear on the master branch and I was probably simply too lazy to cherry-pick them into the 1.5.x or 1.5.2.x branch.
That's why there are currently two main development branches: master for everything 1.5.x and asciidoctor-1.6.0 for everything 1.6.0.
I think that we will live for some time with this situation until 1.6.0 can fully replace the 1.5.x releases.

We definitely have to communicate this better.

mojavelinux:
I'd just say it was a learning experience and we're aiming to getting everything aligned very soon. It will be one of the main things we discuss at Devoxx, I have no doubt.

robertpanzer:
I don't think it's a problem. 
I just checked old appveyor build logs with 1.7.20. 
They are also full of these entries from the tests.

lefou:
I guess this is too outdated now and I will not find myself the time to bring this forward. Closing now.

PR #361, opened 2015-07-18
#333 Create Integrators Guide

robertpanzer:
I created a first skeleton of a document that should describe how to embed and extend Asciidoctor in Java.
The document has barely any content but I'd like to collect some feedback on the structure of the document. 

I would have preferred to include the source code examples from tests but that would mean that we could not simply serve the document from GitHub.

Wdyt?


mojavelinux:
> I would have preferred to include the source code examples from tests but that would mean that we could not simply serve the document from GitHub.

Actually, there's a way we could do this and I'd like to experiment with it in this project. The idea is that you use an include in the document, but we use a pre-commit hook of some sort to expand these includes in place and update the document so that the full source shows on GitHub.

Here's how the input might look:

```
----
include::path/to/example[]
----
```

Here's how the generated source would look:

```
----
ifdef::env-github[]
example content
endif::[]
ifndef::env-github[]
include::path/to/example[]
endif::[]
----
```

That's just a rough idea. Of course, we might be able to find a way to make it more concise. The idea would be that the "example content" would get updated in a pre-commit hook to the latest include code...but you'd never touch that part.

The other idea is to simply the document through DocGist or the like where includes work.

mojavelinux:
cc: @graphitefriction. Also see #333.

robertpanzer:
I just played a bit with pre-commit hooks.
As far as I understood you can't add them to the repository, they're always local only and have to be distributed on another channel than GitHub.

What about writing the "original" content somewhere else and having a gradle task that does the includes (and runs the included snippets as a test suite?)
So whenever you change and commit a snippet or the document you also commit the processed adoc file like you would commit a binary file as a result of a build process. (We certainly never commit binary files;-) )

robertpanzer:
I added a new submodule `asciidoctorj-documentation` that contains the docs with includes and the sources in test cases.
The documents will only be created in `docs/` if the tests succeed.

I don't know if there is a way to reuse the Asciidoctor include processing without rendering, therefore I reimplemented it in the gradle file. (Spaghetti code though).
But nice as an exercise:)

abelsromero:
> Actually, there's a way we could do this and I'd like to experiment with it in this project. The idea is that you use an include in the document, but we use a pre-commit hook of some sort to expand these includes in place and update the document so that the full source shows on GitHub.

Sorry to interrupt but, @mojavelinux are you thinking about what was discussed here? -> https://github.com/asciidoctor/asciidoctor.org/issues/403?
If that's the case, I'm steping in with my opinions:

* First, I am not 100% sure if the idea of commiting a folder with sources and another with processed files is a good approach. Other that duplication issues, imho, this means using git as some sort of middle "publication" repositoy instead of just as "source" repostiory.
* Secondly, I like the idea of integrating this as a gradle task (at least for some projects). We could even include the pre-compile hook script in the `gradle` folder and add a task to install it. And afterall, the gradle wrapper could be embedded in any kind of project.

But, if instead of creating a copy, the gradle task replaces the same file with the `ifdef::env-github[]` we could call a first step.
That way we have the content shown on GitHub and the original include to update it when needed, all in one file.

Another possible options I can think of are:
* Creating a preprocessor extension. I guess an extension could do what the gradle code does, right?
* Getting help from GitHub team? @mojavelinux do you think they will be interested in implementing nativelly something like what docgist does with links? idea -> `include::{repo}/{path}[fromGitHub]`


The more I think about this problem the more I realize we may need to add some metadata to files and build a separated tool. I see this is a dependency management in any project, and we know how complicated it can become...

robertpanzer:
I am open for every help:) 
Even more as I was planning to concentrate on the content here instead of code as there is definitely a need for a guide about writing extensions and converters.

I also have the impression that a solution to this problem is not only of interest for AsciidoctorJ but for every 'smaller' Open Source project that is not able to host the docs on a separate platform.

mojavelinux:
> Abel wrote:
>  are you thinking about what was discussed here? -> asciidoctor/asciidoctor.org#403?

Exactly.

>  I am not 100% sure if the idea of commiting a folder with sources and another with processed files is a good approach.

I'm not sure yet either. We may have to try different things and see what feels right. It's not a solved problem yet...and we're in that messy design phase :)

> I like the idea of integrating this as a gradle task (at least for some projects).

Agreed. Seems like a natural place to slide it in.

> if instead of creating a copy, the gradle task replaces the same file with the ifdef::env-github[] we could call a first step.

That's where my brain is headed for sure. I think a single file that is self-updating is way better than multiple files (which always gets confusing).

> Getting help from GitHub team?

That's a very long uphill climb. We continue to push for includes to work, but I wouldn't hold my breath or put any money on it. Let's just say it's going to be a long, long road until we get there. Better to work on other ideas until there's some sign of progress.

> Robert wrote:
> I was planning to concentrate on the content here instead of code

Please do. This is more of a side / ongoing conversation about how to make a good experience on GitHub while keeping the content DRY and modularized. I recommend that you write the guide in a modularized fashion (with includes) and we'll continue working on a way to get that rendering nicely on GitHub. Another possibility is that we put a big link at the top of the document to docgist so that readers and view it in it's full glory.

robertpanzer:
Added description of some conversion options like safe mode (copied from the user guide), toFile, backend and attributes.

robertpanzer:
Added description of AST types

robertpanzer:
Added description of BlockMacroProcessors

robertpanzer:
Added description of BlockProcessors

robertpanzer:
Added documentation for preprocessors.
Question #383 should be answered in the documentation though.

robertpanzer:
Aligned docs and examples to renaming of AST data structures

mojavelinux:
You're making excellent progress on this. I recommend merging it soon rather than keeping it in one big pull request so that people are able to start contributing improvements to it.

I noticed a few small things to fix up:

* substitute Asciidoc with AsciiDoc (the D is uppercase in AsciiDoc)
* remove the blank line between the author names and the attribute entries in the header (otherwise, the attribute entries aren't part of the header)

We also might want to think about splitting this into chapters so that it's easier to work with. I'm also wondering whether we should just call this the User Manual for AsciidoctorJ and have one or more chapters focused on integration. In other words, different chapters can target different audiences.

robertpanzer:
Thank you!

I applied your proposed fixes.

I'd like to think about splitting the document into own documents per section once I have the content complete.
Actually the motivation for the integrators guide was to reduce the size of the README.
But this one now also seems to be too large for reading and could be split up.
Let's consider this once the content is complete, maybe at Devoxx?

mojavelinux:
> I'd like to think about splitting the document into own documents per section once I have the content complete.
> ...
> Let's consider this once the content is complete, maybe at Devoxx?

Brilliant.

robertpanzer:
I think the content is complete now.
So imo this could be merged now and further enhanced with the help of the community.

mojavelinux:
@robertpanzer this is fantastic work. This guide is going to be immensely useful for the community and also serve as a role model for documentation within the projects.

My only suggestion is that perhaps we should name the folder "docs" instead of asciidoctorj-documentation. There are two reasons for this. First, I think the word "docs" is generally more approachable (doesn't sound as ominous). Second, it could be misleading that documentation is somehow a code module (because it follows the pattern name of the jars). wdyt?

robertpanzer:
@mojavelinux Thank you! There's probably a lot to add, like how to create all kinds of blocks etc, but I also think that it's a good starting point.

The docs are currently two times in the repo:
1. The source is in asciidoctorj-documentation. This contains includes to the sources.
2. The consolidated document with resolved includes is copied into the `docs/` folder, because as you already said, this location is much more prominent and easier to find. The build process of the asciidoctorj-documentation module resolves the includes (duplicate implementation :disappointed:) and copies the content to `docs/`

The reason for these 2 locations is that Github does not resolve the includes and I am reluctant to use precommit hooks to modify the document, as it makes contributions from the community more difficult.

Do you mean to also move the sources of asciidoctorj-documentation to `docs/` and have the unresolved version in `docs/src/main/asciidoc` and the resolved version in `docs/`?
This should be no problem.

I am also open for any other solution like adding a link to the README that links to the document via docgist. (I would have to change the links in this case though, as the basedir seems to be different and docgist can't resolve them atm: http://gist.asciidoctor.org/?github-robertpanzer%2Fasciidoctorj%2F14a4bd09e8e087a09334ddfc553cddcdf74416b1%2F%2Fasciidoctorj-documentation%2Fsrc%2Fmain%2Fasciidoc%2Fintegrator-guide.adoc )

robertpanzer:
> The ideal solution, and the one I think we should move forward with, is actually getting an automated process to integrate it into either asciidoctor.org or GitHub Pages on this repo so that we link to the published version.

I think I'd prefer to integrate it in asciidoctor.org.
Thereby we could have a central location where all _official_ documentation of Asciidoctor is located.

Then the next question is how do we get the content into the asciidoctor.org repo?
As the documentation contains tests it *must* live within the AsciidoctorJ repo imo.
Would it work to add an empty file to the asciidoctor.org repo that simply includes the remote file from the asciidoctorj repo?
What I mean is adding a file only containing sth like this?

```
include:https://github.com/robertpanzer/asciidoctorj/blob/IntegratorsGuide/asciidoctorj-documentation/src/main/asciidoc/integrator-guide.adoc
```

(The URL should certainly point to a tagged version of that file.)

At the moment it seems as if there is a copy of the README of AsciidoctorJ.
While a manual copy process would work as well I'd prefer to make it less redundant and just reference the original content.

mojavelinux:
> we could have a central location where all official documentation of Asciidoctor is located.

:+1: 

> Then the next question is how do we get the content into the asciidoctor.org repo?

Perhaps the best path to take here is to open a separate issue for publishing the AsciidoctorJ docs and we'll get it sorted out there.

I agree one approach is to use a remote include. The other approach is to use submodules. One way or another, we'll figure it out :)

robertpanzer:
Merged. Next task to publish this on asciidoctor.org will be part of a new PR.

mojavelinux:
We'll raise a glass at YouTube for this great step forward...and another when we get it published. :beers:

robertpanzer:
:beers:

PR #363, opened 2015-07-19
Fixes #362: Explicitely set TCCL as ClassLoader for JRuby LoadService…

### LINKED ISSUE #362: Verify resource loading with JRuby 9.0.0.0
JRuby 1.7 by default uses the TCCL to load resources from Jars like the asciidoctor gem.
As converter and extension registry service implementation are also found via the TCCL we changed our recommendation to setup the Asciidoctor instance without passing a ClassLoader to `Asciidoctor.Factory.create()` but set the TCCL appropriately before invocation of the `create()` method.

JRuby 9.0.0.0 changed this default as you can see in https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/RubyInstanceConfig.java#L1397-L1407
Instead of the TCCL the default loader is the ClassLoader that loaded the `jruby.jar` or the class `RubyInstanceConfig` respectively. The TCCL is only a fallback if this class is loaded via the boot class loader, which would mean to place the `jruby.jar` into `<jre-home>/lib`.
This change in behavior also becomes apparent in the test `WhenClassloaderIsRequired` which fails with JRuby 9.0.0.0.

So tools like the ant task or the SBT plugin would have to pass a class loader again, what we just removed. (And I think it's quite strange to pass in the class loader that loaded Asciidoctor to the create method, `Asciidoctor.Factory.create()` should know which Classloader loaded it;-) )

As a solution I would recommend to let AsciidoctorJ pass the TCCL to `RubyInstanceConfig.setLoader()` if no other class loader is passed to `Asciidoctor.Factory.create()`.
What do you think?

robertpanzer:
… if not defined differently

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/internal/JRubyAsciidoctor.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/internal/JRubyAsciidoctor.java
@@ -110,7 +110,10 @@ public class JRubyAsciidoctor implements Asciidoctor {
         if (classloader != null) {
             config.setLoader(classloader);
         } else {
-            config.setLoader(Thread.currentThread().getContextClassLoader());
+            ClassLoader tccl = Thread.currentThread().getContextClassLoader();
+            if (tccl != null){
+                config.setLoader(tccl);
+            }
         }
 
         injectEnvironmentVariables(config, env);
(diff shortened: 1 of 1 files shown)


robertpanzer:
I don't think this is sth that we should pursue even if that means that we're not able to get new versions of asciidoctor-diagram.
Playing with ClassLoaders will just result in a big mess.

PR #366, opened 2015-07-24
Add own proxying for Reader and PreprocessorReader.

robertpanzer:
At the moment Reader and PreprocessorReader are still the "simple" Ruby wrapper objects with their drawbacks.
For example PreprocessorReader.document() will return a Ruby wrapper again instead of THE Document object that all other AST objects return.

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/extension/PreprocessorReader.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/extension/PreprocessorReader.java
@@ -8,6 +8,13 @@ public interface PreprocessorReader extends Reader {
 
     void push_include(String data, String file, String path, int lineNumber, Map<String, Object> attributes);
 
+    /**
+     *
+     * @return
+     * @deprecated Please use {@link #getDocument()}
+     */
     Document document();
-    
+
+    Document getDocument();
+
 }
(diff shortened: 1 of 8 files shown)


mojavelinux:
Super! While we're touching this class, we need to address a very important deficiency. While it's possible to read all the lines, it's not currently possible to read just one.

We need the following methods added to this interface:

* peek_line - https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/reader.rb#L151
* peek_lines - https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/reader.rb#L183
* read_line - https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/reader.rb#L209
* restore_line - https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/reader.rb#L259
* restore_lines - https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/reader.rb#L271

(For now, ignore the "direct" argument as that's intended more for internal use).

Otherwise, there's very little an extension can do.

mojavelinux:
These incremental methods allow the Preprocessor extension to add and remove lines incrementally as it works through the source.

robertpanzer:
Added these 5 methods to the Reader interface.

mojavelinux:
Very nice! I can't wait to start using them. This change going to make Preprocessors in AsciidoctorJ much more capable.

PR #367, opened 2015-07-24
WIP: Publish Snapshot builds

robertpanzer:
I managed to upload a snapshot version of the asciidoctorj.jar with this via `./gradlew :asciidoctorj:artifactoryPublish`.
But it fails for other projects like the EPUB converter or the test-support project if I want to publish everything with a HTTP 409 (Conflict).

I don't really understand how it is all connected, gradle publishing, the nebula publishing plugin and now the artifactory plugin.
But I thought I already open a PR so that somebody could possibly help bringing this forward.

(And I just don't understand why the binary plugin and the artifactory plugin have to be so different, I mean it's the same company...)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,6 +7,7 @@ buildscript {
     if (JavaVersion.current().isJava7Compatible()) {
       classpath 'com.netflix.nebula:nebula-publishing-plugin:2.2.2'
     }
+    classpath 'org.jfrog.buildinfo:build-info-extractor-gradle:3.1.1'
   }
 }
 
@@ -177,7 +178,11 @@ configure(subprojects.findAll { !it.name.endsWith('-distribution') }) {
 subprojects {
   if (JavaVersion.current().isJava7Compatible()) {
     apply from: rootProject.file('gradle/sign.gradle')
-    apply from: rootProject.file('gradle/deploy.gradle')
+    if (statusIsRelease) {
+      apply from: rootProject.file('gradle/deploy.gradle')
+    } else if (it.version.endsWith('SNAPSHOT')) {
+      apply from: rootProject.file('gradle/deploySnapshot.gradle')
+    }
   }
 }
 
--- a/gradle/deploySnapshot.gradle
+++ b/gradle/deploySnapshot.gradle
@@ -1,0 +1,38 @@
+if (project.name.endsWith('-distribution')) {
+    // artifactory-upload is for publishing configurations
+    apply plugin: 'com.jfrog.artifactory-upload'
+} else {
+    // artifactory is for publishing publications
+    apply plugin: 'com.jfrog.artifactory'
+}
+
+artifactory {
+    contextUrl = 'https://oss.jfrog.org/artifactory'   //The base Artifactory URL if not overridden by the publisher/resolver
+    publish {
+        repository {
+            repoKey = 'oss-snapshot-local'
+            username = System.env['BINTRAY_USER'] ?: (project.hasProperty('bintrayUsername') ? project.bintrayUsername : '')
+            password = System.env['BINTRAY_KEY'] ?: (project.hasProperty('bintrayApiKey') ? project.bintrayApiKey : '')
+            maven = true
+        }
+        defaults {
+
+            if (project.name.endsWith('-distribution')) {
+                // NOTE ignore warning that install task does not exist; the plugin is looking for the wrong task!!
+                publishConfigs('archives')
+            }
+            else {
+                publications('mavenNebula')
+            }
+
+            def dryRun = project.hasProperty('dryRun') && project.dryRun.toBoolean()
+            publishBuildInfo = !dryRun  //Publish build-info to Artifactory (true by default)
+            publishArtifacts = !dryRun   //Publish artifacts to Artifactory (true by default)
+            publishPom = !dryRun && !project.name.endsWith('-distribution')  //Publish generated POM files to Artifactory (true by default).
+            publishIvy = false   //Publish generated Ivy descriptor files to Artifactory (true by default).
+
+        }
+    }
+}
+
+artifactoryPublish.dependsOn preparePublish
--- a/gradle/publish.gradle
+++ b/gradle/publish.gradle
@@ -67,14 +67,11 @@ afterEvaluate {
     //artifacts.clear()
     //from components.java
 
-    // NOTE only build sources and javadoc jars when releasing
-    if (project.statusIsRelease) {
-      if (project.tasks.withType(Jar).findByName('sourcesJar')) {
-        artifact sourcesJar
-      }
-      if (project.tasks.withType(Jar).findByName('javadocJar')) {
-        artifact javadocJar
-      }
+    if (project.tasks.withType(Jar).findByName('sourcesJar')) {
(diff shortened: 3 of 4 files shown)


ancho:
Hi Robert,

i tested it with my local artifactory server. It works without any complain. It just publishes every artifact. So what exactly is the error message you get if you run it?

robertpanzer:
I get a HTTP 409 (Conflict).
I retried publishing the modules one by one  and the SNAPSHOT versions work well.
Only the final artifacts asciidoctorj-pdf and asciidoctorj-epub3 and asciidoctorj-diagram produce that error.
I think the problem is that these modules are in fact no SNAPSHOT versions.

Should we explicitly disable the task artifactoryPublish for these 3 modules?

ancho:
Jep. That's it. I reconfigured the repository on my artifactory server and I got the 409 Conflict, too. 
I wouldn't exclude them per se. Maybe check if the version ends with SNAPSHOT.

Like `if (!it.name.endsWith('-distribution') && it.version.endsWith('SNAPSHOT'))`

ancho:
Btw. Do you know why the asciidoctorj-diagram package isn't published at bintray?

robertpanzer:
The submodule was completely missing when AsciidoctorJ 1.5.2 was released. 
I think Alex added this afterwards. 

I will add your proposed change, sounds reasonable.

ancho:
Ahh. Ok. Thank you!

robertpanzer:
Applied the proposed change. 
Another question: Should javadocs and sources be published as well? Atm only the jars will be uploaded.

robertpanzer:
Always build javadoc and sources jars now.
Published first snapshot based on the current AsciidoctorJ 1.6.0.
It can be used like this:
```
repositories {
    maven {
        url 'http://oss.jfrog.org/oss-snapshot-local/'
    }
    jcenter()
}
dependencies {
    compile 'org.asciidoctor:asciidoctorj:1.6.0-SNAPSHOT'
}
```

gradle/publish.gradle
```diff
@@ -68,13 +68,11 @@ afterEvaluate {
     //from components.java
 
     // NOTE only build sources and javadoc jars when releasing
```
mojavelinux: Can you update this comment since you are now always generating sources & JavaDoc jars.

gradle/deploySnapshot.gradle
```diff
@@ -0,0 +1,25 @@
+apply plugin: "com.jfrog.artifactory"
```
mojavelinux: Could you switch this to single quotes to be consistent with the other build scripts?

build.gradle
```diff
@@ -177,7 +181,11 @@ configure(subprojects.findAll { !it.name.endsWith('-distribution') }) {
 subprojects {
   if (JavaVersion.current().isJava7Compatible()) {
     apply from: rootProject.file('gradle/sign.gradle')
-    apply from: rootProject.file('gradle/deploy.gradle')
+    if (statusIsRelease) {
+      apply from: rootProject.file('gradle/deploy.gradle')
+    } else if (!it.name.endsWith('-distribution') && it.version.endsWith('SNAPSHOT')) {
```
mojavelinux: I think we should publish distribution snapshots as well as that's the easiest way for early adopters to test the application. Agreed?

build.gradle
```diff
@@ -2,11 +2,15 @@
...
     jcenter()
+    maven {
+      url 'http://oss.jfrog.org/artifactory/plugins-release'
+    }
   }
   dependencies {
     if (JavaVersion.current().isJava7Compatible()) {
       classpath 'com.netflix.nebula:nebula-publishing-plugin:2.2.2'
     }
+    classpath "org.jfrog.buildinfo:build-info-extractor-gradle:3.0.3"
```
mojavelinux: Could you switch this to single quotes as well?

build.gradle
```diff
@@ -2,11 +2,15 @@
 buildscript {
   repositories {
     jcenter()
+    maven {
+      url 'http://oss.jfrog.org/artifactory/plugins-release'
+    }
```
mojavelinux: Is it possible to add the URL only when publishing the snapshots? Otherwise, Gradle is going to attempt to resolve jars in the snapshot repository whenever it runs.

mojavelinux:
Working with the publishing plugins in Gradle definitely tests your patience. You've done an amazing job navigating those rough seas. I left a few inline comments, but overall this is looking very good.

I'd like to highlight one of my inline comments here. I think we should publish snapshots of the distribution if possible. I realize this may be tricky as it attempts to bundle the non-snapshot jars of the bundled modules (pdf, epub3, etc). Perhaps the distribution snapshot should only contain the core jar. If you'd like, we can address this in a separate issue and move forward with getting the snapshot of the core jar published for now, at least.

mojavelinux:
> Do you know why the asciidoctorj-diagram package isn't published at bintray?

As Robert mentioned, this came later, so it's still pending. Of course, we could publish it at any time, but I'm sort of waiting for the Asciidoctor Diagram 1.3.0 release to be available first. Then we can follow up with publishing that jar like we did with the Asciidoctor PDF update.

robertpanzer:
Thank you for the feedback, Dan!
Also for the inline comments. 
I forgot that because I was probably simply happy that I got it running at all after following the docs provided by jfrog:)
I will fix that. 
Hope it's ok if I already dropped some jars in the repository. 

I am quite unsure whether I can conditionally set the repositories in the buildscript.classpath. 
Afaik the buildscript part is evaluated in a dedicated step before anything else is processed.
But I will see if I find something. 

Thanks!

mojavelinux:
>  I was probably simply happy that I got it running at all

Applause!!! Cheers!!! Applause!!!

> Hope it's ok if I already dropped some jars in the repository.

Of course. These are just snapshots after all :)

> I am quite unsure whether I can conditionally set the repositories in the buildscript.classpath.

What surprises me is that it even needs to be set. I would think that the plugin would take this location as a configuration parameter and work out internally where to send the jars. But that might lead us back to your original comment that the publishing plugins are rough around the edges.

build.gradle
robertpanzer: Just found that this plugin is also available for the plugins element: https://plugins.gradle.org/plugin/com.jfrog.artifactory
The documentation seems to lag behind. Will try that.

robertpanzer:
Also publish snapshot of distribution now.

robertpanzer:
Jars are stored with timestamp in artifactory and are not replaced on subsequent publications.
The distribution gets no timestamp appended though and is replaced with every publish though.
Is that ok?
Should we manually cleanup old snapshot versions?

mojavelinux:
> Is that ok?

I think so. The idea of the snapshot is to allow people to evaluate the latest master. The version shouldn't be considered stable so we don't want to get people used to relying on it. In other words, change is good :)

> Should we manually cleanup old snapshot versions?

If possible, then perhaps, though I wouldn't worry too much about it. I would expect this would be something the service itself would handle.

robertpanzer:
Added description to README how to publish snapshots and how to consume them. Squashed commits before merging.

mojavelinux:
:+1:

PR #370, opened 2015-08-02
add fix to upstream test script for OSX

mojavelinux:
Add a blank extension value to the `-i` flag when invoking `sed` so that it works on OSX. This should work in theory, but I haven't tested it yet on an OSX box.

The same patch should be applied to the master branch.

### DIFF
--- a/test-asciidoctor-upstream.sh
+++ b/test-asciidoctor-upstream.sh
@@ -19,8 +19,8 @@ unzip -q $SRC_DIR.zip
 cp ../../asciidoctor-gem-installer.pom $SRC_DIR/pom.xml
 cd $SRC_DIR
 ASCIIDOCTOR_VERSION=`grep 'VERSION' ./lib/asciidoctor/version.rb | sed "s/.*'\(.*\)'.*/\1/"`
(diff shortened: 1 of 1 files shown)


robertpanzer:
Jus tried sth like this on my Mac and still getting the error

```
$ sed -i'' 's/a/c/' test.txt
sed: 1: "test.txt": undefined label 'est.txt'
```

The only way to get it working is to do:

```
$ sed -i '' 's/a/c/ test.txt
```

but that would probably not work on Linux again.

What about going the old way?
```
$ sed 's/a/c/' test.txt > test.tmp; mv test.tmp test.txt
```

The answer here seems to be that there is no way of getting it to work on both platforms: http://stackoverflow.com/questions/5694228/sed-in-place-flag-that-works-both-on-mac-bsd-and-linux

mojavelinux:
(╯°□°)╯︵ ┻━┻

Why sed, why?

Let's just do it the old-fashioned way as it ensures portability. Update coming.

mojavelinux:
Updated.

robertpanzer:
Looks good! sed on OS X does not seem to support `+` at all.

mojavelinux:
:+1:

PR #371, opened 2015-08-02
Renamed AST classes

robertpanzer:
Implements the renaming of the AST classes as discussed at http://discuss.asciidoctor.org/dev-Reorganization-of-the-AsciidoctorJ-AST-data-structures-td2991.html#a3606

That is the new AST hierarchy is:

```
FlowNode
 +- StructuralNode
 | +- Block
 | +- Section
 | +- Document
 | +- List
 | +- ListItem
 | +- Table
 +- PhrasingNode
```

Also moved all implementations to the package `org.asciidoctor.ast.impl` to reduce the size of the last package.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -781,7 +781,7 @@ assertThat(document.doctitle(), is("Document Title")); // <2>
 
 But also all blocks that conforms the document can be retrieved.
 Currently there are support for three kinds of blocks.
-`Blocks` itself, `Section` for sections of the document and `AbstractBlock` which is the base type where all kind of blocks (including those not mapped as Java class) are mapped.
+`Blocks` itself, `Section` for sections of the document and `StructuralNode` which is the base type where all kind of blocks (including those not mapped as Java class) are mapped.
 
 [source]
 ----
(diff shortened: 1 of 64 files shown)


mojavelinux:
I've let my mind sleep on these new names and I'm happy to say that they feel right coming back to them, with one exception. I don't think "FlowNode" should be the top of the tree. I think that's overly specific for the root node as we could have other nodes in the future (like MetaDataNode or something) that isn't a FlowNode. I think the right name may be "ContentNode" since all of this is content...that is our domain. wdyt?

robertpanzer:
:+1: Will change the PR asap.

robertpanzer:
Renamed FlowNode to ContentNode.

robertpanzer:
@mojavelinux: Should I merge and add an announcement to the discussion list?

mojavelinux:
How about post to the list under a new subject that a decision has been made and that the change is pending, but we're inviting comment before it happens.

robertpanzer:
Renamed `Inline` to `PhraseNode`.

PR #372, opened 2015-08-03
Fixes #368: Bundles rouge 1.9.1 with asciidoctorj-pdf excluding the d…

### LINKED ISSUE #368: Bundle rouge gem in AsciidoctorJ PDF
Bundle the rouge gem in AsciidoctorJ PDF when we relesae 1.5.0-alpha.9. The rouge gem is a new source highlighter that will come available in Asciidoctor PDF 1.5.0.alpha.9. See https://github.com/asciidoctor/asciidoctor-pdf/issues/247

robertpanzer:
…emo files.

Works in general.
Seeing two warnings though from the shell lexer:
```
/Users/robertpanzer/dev/asciidoctorj/asciidoctorj-pdf/build/resources/main/gems/rouge-1.9.1/lib/rouge/lexers/shell.rb:20 warning: already initialized constant KEYWORDS
/Users/robertpanzer/dev/asciidoctorj/asciidoctorj-pdf/build/resources/main/gems/rouge-1.9.1/lib/rouge/lexers/shell.rb:25 warning: already initialized constant BUILTINS
```

### DIFF
--- a/asciidoctorj-pdf/build.gradle
+++ b/asciidoctorj-pdf/build.gradle
@@ -10,6 +10,7 @@ dependencies {
     exclude module: 'prawn'
   }
   gems "rubygems:prawn:$prawnGemVersion"
+  gems "rubygems:rouge:$rougeGemVersion"
 }
 
 def gemFiles = fileTree(jruby.gemInstallDir) {
@@ -25,6 +26,8 @@ def gemFiles = fileTree(jruby.gemInstallDir) {
   include "gems/prawn-icon-*/fonts/*/*"
   // Include required data file from the addressable gem
   include "gems/addressable-*/data/*.data"
+
+  exclude 'gems/rouge-*/lib/rouge/demos/**'
 }
 
 jrubyPrepareGems << {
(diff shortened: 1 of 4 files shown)


robertpanzer:
Don't see the warnings though if I use rouge 1.9.0.

robertpanzer:
This seems to be the problem: https://github.com/jneen/rouge/commit/1b91ba0aa627c5d72fc9d68a1ab3f017ff2b2837#diff-d8678e4d6b667616f9e91f188cf9f31aR19

The PowerShell Lexer also defines the two constants KEYWORDS and BUILTINS.
Introduced between 1.9.0 and 1.9.1.
Looking at the code I think it doesn't pose any problem because the constants are immediately used to initialize the rules of the lexer.

mojavelinux:
Looks like this has been reported.

https://github.com/jneen/rouge/issues/296

It's a "harmless" warning, but still would be good to see it fixed upstream.

On a separate note, it's weird that the demo files are in the lib directory. Those should be in the data directory of the gem. I like that you are excluding them.

mojavelinux:
:+1:

PR #373, opened 2015-08-04
convince Asciidoctor PDF to use the thread_safe provided by AsciidoctorJ

mojavelinux:
- downgrade the thread_safe version to 0.3.4
- remove out of date hacks for Asciidoctor PDF 1.5.0.alpha.8

### DIFF
--- a/asciidoctorj-pdf/build.gradle
+++ b/asciidoctorj-pdf/build.gradle
@@ -3,9 +3,7 @@ dependencies {
   gems("rubygems:asciidoctor-pdf:$asciidoctorPdfGemVersion") {
     // Exclude gems provided by AsciidoctorJ core
     exclude module: 'asciidoctor'
-    //Include thread_safe again (0.3.5) because AsciidoctorJ core delivers 0.3.4,
-    // which is incompatible since asciidoctor-pdf 1.5.0.alpha.9
-    //exclude module: 'thread_safe'
+    exclude module: 'thread_safe'
     // We must lock Prawn to 1.3.0 until AsciidoctorJ upgrades to JRuby 9.0.0.0
     exclude module: 'prawn'
   }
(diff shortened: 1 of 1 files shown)


mojavelinux:
Rather than bundle thread_safe 0.3.5, I'd rather coerce Asciidoctor PDF to use 0.3.4. We have much to learn about packaging, but this will be a hard earned lesson. Ruby worries way too much about gem versions.

robertpanzer:
Tested with a large document, works fine.
(Not multithreaded though.)

mojavelinux:
> (Not multithreaded though.)

It should be getting thread_safe from the AsciidoctorJ jar and there were no significant changes in thread_safe from 0.3.4 to 0.3.5 that would affect us, so everything should work exact as it had before.

The more I think about how we hand gems, the more I realize that the solution I think we should pursue for 1.6.0 is to repackage every gem we use individually and publish it to Maven Central / Bintray under the org.asciidoctor namespace. That way, we can depend on them like regular jars without bundling them.

I'm thinking something like:

```
<dependency>
  <artifactId>org.asciidoctor.rubygems</artiflactId>
  <groupId>rubygem-prawn</groupId>
  <version>1.3.0</version>
</dependency>
```

This differs from what the TorqueBox proxy provides because these jars would be published as repackaged gems with the same layout we currently have inside the AsciidoctorJ jars. In the long run, it just feels like a safer approach than bundling and, if the JRuby community ever sorts out how to distribute gems so that they are loadable in JRuby directly, we can just switch to them.

There will be an initial effort up front to repackage and publish all the gems we need, but it should remain reasonable stable over time.

wdyt?

robertpanzer:
Sounds reasonable.
This will be a lot of Gradle fun again ;-)
Will play with the idea after vacation:)

mojavelinux:
Excellent! Of course, I'm open to other strategies as well. We want to pick what is going to be the easiest for us to manage. There isn't a right way, just a pragmatic way. I'm still wrestling over what that actually is.

I just wish JRuby would be able to load a gem as is. Maybe we aren't being clear enough about what we expect, but it seems like a completely reasonable expectation to me.

PR #376, opened 2015-08-08
Integrated PDBBox to validate PDF generation in asciidoctorj-pdf

abelsromero:
This PR integrates a PDF parsing API (PDFBox) to validate that PDFs are properly generated.
Two tests have been added; one to validates image inclusion and another for text highlighting.

There's one issue though, the parsing of the PDFs generates many warning errors from PDFBox. This "pollutes" quite a bit the console of the CI. If this is a problem I can investigate some workarround.


robertpanzer:
Wow! This is awesome!
This is the missing link between the document you recently added and the code:)

I'm just seeing many ClassCastExceptions in the source highlighting test:
```
Aug 09, 2015 6:24:17 PM org.apache.pdfbox.util.PDFStreamEngine processOperator
WARNUNG: java.lang.ClassCastException: org.asciidoctor.util.pdf.ColorsProcessor cannot be cast to org.apache.pdfbox.pdfviewer.PageDrawer
java.lang.ClassCastException: org.asciidoctor.util.pdf.ColorsProcessor cannot be cast to org.apache.pdfbox.pdfviewer.PageDrawer
	at org.apache.pdfbox.util.operator.pagedrawer.LineTo.process(LineTo.java:45)
	at org.apache.pdfbox.util.PDFStreamEngine.processOperator(PDFStreamEngine.java:557)
	at org.apache.pdfbox.util.PDFStreamEngine.processSubStream(PDFStreamEngine.java:268)
	at org.apache.pdfbox.util.PDFStreamEngine.processSubStream(PDFStreamEngine.java:235)
	at org.apache.pdfbox.util.PDFStreamEngine.processStream(PDFStreamEngine.java:215)
	at org.asciidoctor.util.pdf.ColorsProcessor.parse(ColorsProcessor.java:63)
	at org.asciidoctor.WhenBackendIsPdf.pdf_source_code_should_be_highlighted(WhenBackendIsPdf.java:77)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
```

asciidoctorj-pdf/build.gradle
```diff
@@ -12,6 +12,8 @@ dependencies {
   }
   gems "rubygems:prawn:$prawnGemVersion"
   gems "rubygems:rouge:$rougeGemVersion"
+
+  testCompile 'org.apache.pdfbox:pdfbox:1.8.10'
```
robertpanzer: Could you add a variable for the version to `asciidoctorj/build.gradle`?

asciidoctorj-pdf/src/test/java/org/asciidoctor/util/pdf/ImageProcessor.java
```diff
@@ -0,0 +1,151 @@
...
+
+            if (xobject instanceof PDXObjectImage)  {
+                PDXObjectImage image = (PDXObjectImage)xobject;
+                int imageWidth = image.getWidth();
+                int imageHeight = image.getHeight();
+                PDPage page = getCurrentPage();
+                double pageHeight = page.getMediaBox().getHeight();
+
+                Matrix ctmNew = getGraphicsState().getCurrentTransformationMatrix();
+                float yScaling = ctmNew.getYScale();
```
robertpanzer: I wonder if all these computations could be removed, as the test only counts the number of images?

abelsromero: I though it can be usefull anyway, for instance if we want to validate image scaling on some other test.
Do you think we should remove it?

asciidoctorj-pdf/build.gradle
abelsromero: Sure thing!

asciidoctorj-pdf/src/test/java/org/asciidoctor/util/pdf/ImageProcessor.java
robertpanzer: If there are other tests coming, keep it.
I just wonder if such a test doesn't better fit into asciidoctor-pdf.

I like the source-highlighting test in particular, because it tests that we got the packaging of rouge right.
Things like image scaling test would imo fit better close to the ruby sources as AsciidoctorJ has no impact on that. 
But I'm also fine with keeping it here.

abelsromero:
> This is the missing link between the document you recently added and the code:)

Knowing that PDF generation was tested using `file.exists()` worried me because (srry, if am wrong) I did not see any test in the ruby gem project.
Also, being self-demanding as we are, the fact that PDFs are not as easy to deal with as HTML should not stop us from testing them to some similar degree.

> I'm just seeing many ClassCastExceptions in the source highlighting test:

Those nasty ClassCasExceptions are the warnings I mentioned. I think this could be removed if `ColorsProcessor` extended `PageDrawer` instead of `PDFTextStripper`. But has some drawbacks, let me explain:

The way PDFBox works is similar to SAX XML parser, it offers events, you have to extend basic class and override some methods. In the case of PDFBox it also offers some basic classes with some features already done.
So the trick is choosing which one offers what is closer to your needs.
For text treatment, the closest one is PDFTextStripper, but to get the colors at some point it calls PageDrawer and some elements inside the document fails \*.
I already tried extending PageDrawer which prevents those warnings, but I had to reimplement some logic provided by PDFTextStripper and override some methods to avoid other issue. To sum up, using PageDrawer mean a less clean solution and actual exceptions, while PDFTextStripper offers a cleaner solutions but lots of warnings.

*Still, I'll give it a try and come back later to report, maybe I can find a better solution*

 \* btw, this does not happen on all PDFs I have tested, but Asciidoctor one always show some warning. But given the preview state I won't worry about this right now.

mojavelinux:
> Knowing that PDF generation was tested using file.exists() worried me because (srry, if am wrong)

Don't be sorry. You are right!

> I did not see any test in the ruby gem project.

That's one of the requirements to satisfy before we can move to beta. I'd like to use something from the Ruby ecosystem to test them. See https://github.com/asciidoctor/asciidoctor-pdf/issues/37

> Also, being self-demanding as we are, the fact that PDFs are not as easy to deal with as HTML should not stop us from testing them to some similar degree.

Correct. It is a bit trickier to test them, but as I mention in the linked report, there are still plenty of things that we can test.

mojavelinux:
For now, I don't mind having tests in AsciidoctorJ. Extra coverage certainly can't hurt (in this case, coverage at all).

abelsromero:
I squashed the following changes:
* Added 'pdbboxVersion': following the current pattern I added this to the build.gradle at thet root, but I'm not sure if we should add a properties that only apply to one sub-project.
* Found a solution for the ClassCastException messages: just created a custom configuration for the processor removing those components that used PageDrawer. Quite simple and clean in the end.

@mojavelinux while the testing in ruby is not set in stone, open issues If you think some specific feature could be tested here. I only focused in the ones that have been reported so far.

robertpanzer:
> I'm not sure if we should add a properties that only apply to one sub-project.

We also have properties for netty or jsoup that are only used in core.

robertpanzer:
> Found a solution for the ClassCastException messages: 

:+1:

robertpanzer:
> some specific feature could be tested here.

It would then be great if we could test against pdf prereleases like we test against asciidoctor prereleases. 
Otherwise we find the errors only after a release:)

abelsromero:
I see the build failed on Opendk7 for
```
Could not GET 'http://rubygems-proxy.torquebox.org/releases/rubygems/rubyzip/1.1.7/rubyzip-1.1.7.pom'. Received status code 503 from server: Service Temporarily Unavailable
```
The rest worked fine. Robert, I think you can relaunch the job since you have write access on the repo.


> It would then be great if we could test against pdf prereleases like we test against asciidoctor prereleases.
> Otherwise we find the errors only after a release:)

I looked into `test-asciidoctor-upstead.sh` and this should be easy. But this makes me think if we should test all other asciidoctor gems:epub or diagram.

robertpanzer:
Thank you!

No need to rebuild. We often have the situation that we randomly see 503 when resolving gems.
As all other builds succeeded, this would have been fine as well.

PR #378, opened 2015-08-15
Test pdf upstream

robertpanzer:
Make Travis test against asciidoctor-pdf dev as well.
This makes sense as asciidoctorj-pdf now also tests on how the PDF eventually looks like after recent additions from @abelsromero with #376


mojavelinux:
:+1:

PR #379, opened 2015-08-19
Upgrade jruby-gradle plugin to 1.0.2

### LINKED ISSUE #375: Upgrade JRuby Gradle plugins to 1.0
The JRuby Gradle plugins have reached 1.0. We should upgrade and align with them.

http://jruby-gradle.org/news/2015/08/04/jrubygradle-one-point-oh/

Note that the default JRuby runtime is JRuby 9.0.0.0.

We should also look into the jar plugin again and see if we can use it now. Currently, we're only using the base plugin because the jar plugin didn't fit our use case. If that's still the case, we should make sure to file issues upstream so that eventually we can use it.

I think it's fine to upgrade only the 1.6.0 branch, but I'll leave that decision up to @robertpanzer.

robertpanzer:
This PR fixes #375 

As the jruby-gradle base plugin has not been upgraded on the Gradle Plugin Portal the plugin is now pulled again from the buildscript section.

Additionally the jruby-gradle-plugin is compiled to the Java 7 class file format, so the build will no longer run with Java 6.
(AsciidoctorJ should still run on Java6, though we won't be able to verify it.)

### DIFF
--- a/asciidoctorj-core/build.gradle
+++ b/asciidoctorj-core/build.gradle
@@ -26,7 +26,7 @@ def gemFiles = fileTree(jruby.gemInstallDir) {
   include "gems/asciidoctor-${asciidoctorGemVersion}/data/**"
 }
 
-jrubyPrepareGems << {
+jrubyPrepare << {
   copy { // bundles the gems inside this artifact
     from gemFiles
     into sourceSets.main.output.resourcesDir
--- a/asciidoctorj-diagram/build.gradle
+++ b/asciidoctorj-diagram/build.gradle
@@ -16,7 +16,7 @@ def gemFiles = fileTree(jruby.gemInstallDir) {
   include "gems/asciidoctor-diagram-${asciidoctorDiagramGemVersion}/data/**"
 }
 
-jrubyPrepareGems << {
+jrubyPrepare << {
   copy { // bundles the gems inside this artifact
     from gemFiles
     into sourceSets.main.output.resourcesDir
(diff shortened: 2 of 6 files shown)


mojavelinux:
:+1: 

I suppose in the future we could test against a remote Arquillian container running in Java 6...but then again, in the future, who cares about Java 6 ;)

robertpanzer:
So merge and disable the Java6 build?
I'm a bit afraid of that because using some APIs only available in Java 7 goes just too fast.
Avoiding use of Java 7 language options is easy, but the APIs... I just made these errors myself too often.

abelsromero:
> Avoiding use of Java 7 language options is easy, but the APIs... I just made these errors myself too often.

It happened to me right now using PathMatcher.

In the end I think it's a matter of fully commiting to Java 7 at some point and ignoring any Java 6 support (it's 2015! (╯°□°)╯︵ ┻━┻). Having a mid-term scenario where we offer a Java 6 release but we're not testing it, is a bit weird to me and has the mentioned risks. So, I'm would not consider it.

However, going fully v7 is a big change and I'm not sure that doing so for a building tool is a reason good enough. Also it would be a shame not to offer all the changes in the extension/AST API to Java 6 users.
What about moving this to 1.6.1 or even 1.7.0? 1.6.0 could be focused on polishing all those changes to offer a stable extension API compatible with Java6. Once it's done, we can announce the change to Java 7 and see how community reacts. wdyt?

mojavelinux:
I'm pretty sure the community is on board with Java 7. We've asked several times and we've never gotten anyone that said they'd drop AsciidoctorJ if we didn't have Java 6 (that I can remember). I'm not trying to push this forward, just clarifying that I'm not voting to hold back.

This is also where we have to play the community card. If someone really wants Java 6 support, then they can be the Java 6 support person :) Otherwise, we're really just guessing about use cases.

robertpanzer:
I guess this is no longer relevant, we're already using jruby-gradle.base 1.4.0.

PR #380, opened 2015-08-20
Fixes #201. Added methods createSection() to Processor.

### LINKED ISSUE #201: Cannot create Sections blocks in Block extensions
When writing an extension block with Java you can create any Block as a result of the extension and an Inline, but not a Section. (see here http://discuss.asciidoctor.org/How-to-generate-sections-in-extension-td2142.html) 

We need to create a createSection method as well that returns a Section.

robertpanzer:


mojavelinux:
:+1:

PR #390, opened 2015-10-24
Changed dependency on JRuby to 9.0.3.0.

### LINKED ISSUE #230: Use "org.jruby:jruby" as dependency instead of "org.jruby:jruby-complete"
#217 and #219 did a real great job on reducing the size of the distributed artifact! But it now depends on "org.jruby:jruby-complete" which is an uber-jar containing several 3rd-party artifacts, e.g. ASM 4.0.

The problem: the application I'm working on (http://jqassistant.org) uses asciidoctorj but requires a newer version of ASM (5.0) for scanning Java 8 bytecode. It would be great if I could just override the ASM version in the dependency management section of Maven. Due to the uber-jar this is currently not possible but could be achieved if asciidoctorj would depend on "org.jruby:jruby".

Current workaround: I'm excluding the jruby dependency from the asciidoctorj artifact and defining an explicit dependency of my artifact to org.jruby:jruby.

robertpanzer:
Bumped JRuby version to 9.0.3.0.
As JRuby only works with Java >= 7 I also bumped the source and target version of AsciidoctorJ to 1.7.

Additionally changed the dependency from `ruby-complete` to `jruby` which fixes #230.
With this the lib directory of the distribution looks like this:

```
asciidoctorj-1.6.0-SNAPSHOT.jar
asciidoctorj-epub3-1.5.0-alpha.4.jar
asciidoctorj-pdf-1.5.0-alpha.9.jar
bytelist-1.0.13.jar
dirgra-0.3.jar
high-scale-lib-1.0.6.jar
invokebinder-1.5.jar
jcodings-1.0.13.jar
jcommander-1.35.jar
jffi-1.2.9-native.jar
jffi-1.2.9.jar
jnr-constants-0.9.0.jar
jnr-enxio-0.9.jar
jnr-netdb-1.1.4.jar
jnr-posix-3.0.19.jar
jnr-unixsocket-0.8.jar
jnr-x86asm-1.0.2.jar
joda-time-2.8.2.jar
joni-2.1.8.jar
jruby-9.0.3.0.jar
jruby-core-9.0.3.0.jar
jruby-stdlib-9.0.3.0.jar
jzlib-1.1.3.jar
nailgun-server-0.9.1.jar
options-1.1.jar
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ jdk:
   - oraclejdk8
   - oraclejdk7
   - openjdk7
-  - openjdk6
 before_script: unset GEM_PATH GEM_HOME JRUBY_OPTS
 install: ./gradlew assemble
 script: ./gradlew -S check && bash test-asciidoctor-upstream.sh
--- a/asciidoctorj-core/build.gradle
+++ b/asciidoctorj-core/build.gradle
@@ -1,7 +1,7 @@
 apply plugin: 'osgi'
 
 dependencies {
-  compile "org.jruby:jruby-complete:$jrubyVersion"
+  compile "org.jruby:jruby:$jrubyVersion"
   compile "com.beust:jcommander:$jcommanderVersion"
   gems "rubygems:asciidoctor:$asciidoctorGemVersion"
   gems "rubygems:coderay:$coderayGemVersion"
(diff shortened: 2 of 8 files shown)


mojavelinux:
:+1: Moving forward!

Ideally, we want to use the most minimal JRuby possible. I think we had the dependency on jruby-complete because some part of AsciidoctorJ was using something outside of JRuby core. But it's much better to use as little of it as we can get away with.

PR #394, opened 2015-10-31
Bump version of Asciidoctor to 1.5.3 and fixes #393.

### LINKED ISSUE #393: ContentNode.hasRole() always returns false if multiple roles are present on a node.
ContentNode.hasRole() unfortunately maps to `role?` which returns true if the given string exactly matches the `role` attribute.
Instead it should map to `has_role?` which returns true if the given role is part of the attribute.

robertpanzer:
Also added new methods addRole() and removeRole() that became part of Asciidoctor 1.5.3.
Adapted to change of DocinfoProcessor option for adding Docinfo to the header (:header -> :head).

When this PR is merged I'd like to cut an Asciidoctor 1.6.0-alpha release.
Then we do not only make a difference internally to 1.5.2 but also by the numbers:
Asciidoctor 1.5.2 -> 1.5.3
JRuby 1.7.x -> 9.0.3.0
Java 6 -> 7

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/ast/ContentNode.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/ast/ContentNode.java
@@ -53,6 +53,11 @@ public interface ContentNode {
      */
     String role();
     List<String> getRoles();
+
+    void addRole(String role);
+
+    void removeRole(String role);
+
     boolean isReftext();
     String getReftext();
     String iconUri(String name);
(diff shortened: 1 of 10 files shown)


mojavelinux:
:+1: It will be awesome to have a new alpha available for announcement and hacking at Devoxx.

Btw, what I plan to do for core is branch 1.5.x and make master 1.6.0.dev. That way, 1.6.0 core doesn't get stuck behind 1.5.4. And 1.5.4 will only be if necessary.

robertpanzer:
:beers: :+1:

PR #396, opened 2015-11-01
Upgrade to Asciidoctor 1.5.3 and JRuby 1.7.22

robertpanzer:
Upgraded the dependency on Asciidoctor for the master! branch.
This is to prepare a release of AsciidoctorJ 1.5.3.


mojavelinux:
Excellent!

As I mentioned in the other thread, let me update Asciidoctor PDF and EPUB3
so we can use new versions across the board. I'll ping when they're ready.

PR #399, opened 2015-11-03
FIXED #398 - Compile for JDK 1.6

### LINKED ISSUE #398: AsciidoctorJ 1.5.3 classes are compiled for Java 8
Somehow gradle looses the information that it should compile the classes for Java 6, as specified in gradle.properties and the root build.gradle file.

Debug output shows that no target version is passed to the Java compiler, so that the classes are created for Java 8 if a JDK 8 is used.

It would be great if someone with some more gradle expertise could have a look at that, maybe @ysb33r or @melix 

I'll try to get a 1.5.3.1 published asap.

ysb33r:


robertpanzer:
Thank you!!!!

Do you have an idea, why setting the properties source- and targetCompatibility does not work?
Is the multi-project setup too complex?

ysb33r:
Since @melix is on the Gradle team now, he can possibly commetn on the issue. All I know that it is probably a bug in Gradle and that is the fix. (Kudos to @aalmiray for showing that fix in a conference presentation).

PR #408, opened 2015-11-08
WIP: Fixes #404. Add AST types for definition lists.

### LINKED ISSUE #404: Asciidoctorj-1.6.0 Definition lists: Don't know what to do with a Array
Set up a document that contains a definition list.

```
= Title

== Chapter

first term:: definition of first term
section term:: definition of second term
```

This stack trace pops up

```
Don't know what to do with a Array
java.lang.IllegalArgumentException: Don't know what to do with a Array
	at org.asciidoctor.ast.NodeConverter$NodeType.getNodeType(NodeConverter.java:69)
	at org.asciidoctor.ast.NodeConverter.createASTNode(NodeConverter.java:97)
	at org.asciidoctor.internal.RubyBlockListDecorator.get(RubyBlockListDecorator.java:84)
	at org.asciidoctor.internal.RubyBlockListDecorator.get(RubyBlockListDecorator.java:14)
	at java.util.AbstractList$Itr.next(AbstractList.java:358)
```

The problem seems to be access to `ListImpl.getItems()`. 

This happily works for ther kinds of lists, but not for definition lists.

robertpanzer:
I created two new classes in the AST package:

`DefinitionList` is a `StructuralNode` and contains `DefinitionListItem`s.

`DefinitionListItem` has a list of `definitionTerm`s, that are `ListItem`s, and a `definition`, that is also a `ListItem`.
But a `DefinitionListItem` is not a `ContentNode`.
And that's maybe not so nice, because `DefinitionList.getBlocks()` returns a list of `StructuralNode`s per its signature.
In fact the members are `DefinitionListItem`s and therefore no `StructuralNode`s.

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/ast/DescriptionList.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/ast/DescriptionList.java
@@ -1,0 +1,14 @@
+package org.asciidoctor.ast;
+
+public interface DescriptionList extends StructuralNode {
+
+    java.util.List<DescriptionListEntry> getItems();
+
+    boolean hasItems();
+
+    @Deprecated
+    public String render();
+
+    public String convert();
+
+}
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/ast/DescriptionListEntry.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/ast/DescriptionListEntry.java
@@ -1,0 +1,9 @@
+package org.asciidoctor.ast;
+
+public interface DescriptionListEntry {
+
+    java.util.List<ListItem> getTerms();
+
+    ListItem getDescription();
+
+}
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/ast/NodeConverter.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/ast/NodeConverter.java
@@ -3,13 +3,17 @@ package org.asciidoctor.ast;
 import org.asciidoctor.ast.impl.BlockImpl;
 import org.asciidoctor.ast.impl.CellImpl;
 import org.asciidoctor.ast.impl.ColumnImpl;
+import org.asciidoctor.ast.impl.DescriptionListImpl;
+import org.asciidoctor.ast.impl.DescriptionListEntryImpl;
 import org.asciidoctor.ast.impl.DocumentImpl;
 import org.asciidoctor.ast.impl.ListImpl;
 import org.asciidoctor.ast.impl.ListItemImpl;
 import org.asciidoctor.ast.impl.PhraseNodeImpl;
 import org.asciidoctor.ast.impl.SectionImpl;
 import org.asciidoctor.ast.impl.TableImpl;
+import org.asciidoctor.internal.RubyObjectWrapper;
 import org.jruby.Ruby;
+import org.jruby.RubyArray;
 import org.jruby.RubyClass;
 import org.jruby.RubyModule;
 import org.jruby.java.proxies.RubyObjectHolderProxy;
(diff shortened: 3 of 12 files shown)


robertpanzer:
What about changing `List.getBlocks()` to always return `null`?
I think it's more logical to require calling the method `List.getItems()` or `DefinitionList.getItems()` respectively.

LightGuard:
What did it return before, was it a collection? If it did, it should return
an empty collection.

On Sunday, November 8, 2015, Robert Panzer <notifications@github.com> wrote:

> What about changing List.getBlocks() to always return null?
> I think it's more logical to require calling the method List.getItems()
> or DefinitionList.getItems() respectively.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/asciidoctor/asciidoctorj/pull/408#issuecomment-154871728>
> .
>


-- 
Sent from Gmail Mobile

LightGuard:
Also, wouldn't getChildren be better?

On Sunday, November 8, 2015, Jason Porter <lightguard.jp@gmail.com> wrote:

> What did it return before, was it a collection? If it did, it should
> return an empty collection.
>
> On Sunday, November 8, 2015, Robert Panzer <notifications@github.com
> <javascript:_e(%7B%7D,'cvml','notifications@github.com');>> wrote:
>
>> What about changing List.getBlocks() to always return null?
>> I think it's more logical to require calling the method List.getItems()
>> or DefinitionList.getItems() respectively.
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/asciidoctor/asciidoctorj/pull/408#issuecomment-154871728>
>> .
>>
>
>
> --
> Sent from Gmail Mobile
>


-- 
Sent from Gmail Mobile

robertpanzer:
At the moment `getItems()` is just an alias for `getBlock()` and returns a collection of blocks, that are now called `StructuralNode`s: 
```java
List<StructuralNode> getBlocks()
```

Now the problem with a `DefinitionList` is that its items, the `DefinitionListItem`s are not nodes in asciidoctor core, nor in HTML5.
So I cannot make `StructuralNode`s from them that support methods like `getStyle()`, `getContent()` etc.

That's why I think it might make sense to distinguish `item`s from `block`s

> What did it return before, was it a collection? If it did, it should return
an empty collection.

Yes, I think that's a good idea.

> Also, wouldn't getChildren be better?

Why not. HTML5 also calls it children and says in http://www.w3.org/TR/html5/dom.html#palpable-content that a `dl` is palbable content '_if the element's children include at least one name-value group_'.
But that still leaves us with the question about the type of a `DefinitionListItem` which is in HTML5 a name-value-group and what this `getChildren()` methods returns.

LightGuard:
Sounds like we should ask the community again.

On Mon, Nov 9, 2015 at 12:53 AM, Robert Panzer <notifications@github.com>
wrote:

> At the moment getItems() is just an alias for getBlock() and returns a
> collection of blocks, that are now called StructuralNodes:
>
> List<StructuralNode> getBlocks()
>
> Now the problem with a DefinitionList is that its items, the
> DefinitionListItems are not nodes in asciidoctor core, nor in HTML5.
> So I cannot make StructuralNodes from them that support methods like
> getStyle(), getContent() etc.
>
> That's why I think it might make sense to distinguish items from blocks
>
> What did it return before, was it a collection? If it did, it should return
> an empty collection.
>
> Yes, I think that's a good idea.
>
> Also, wouldn't getChildren be better?
>
> Why not. HTML5 also calls it children and says in
> http://www.w3.org/TR/html5/dom.html#palpable-content that a dl is
> palbable content '*if the element's children include at least one
> name-value group*'.
> But that still leaves us with the question about the type of a
> DefinitionListItem which is in HTML5 a name-value-group and what this
> getChildren() methods returns.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/asciidoctor/asciidoctorj/pull/408#issuecomment-154990138>
> .
>



-- 
Jason Porter
http://en.gravatar.com/lightguardjp

mojavelinux:
I'm looking forward to this addition!

I'd like to suggest a couple of name changes in the terminology. First, it's important that we use the term "description" instead of "definition" everywhere (e.g., `DescriptionList`). That name change was introduced in HTML5 and I'd like to follow it.

See http://www.w3.org/TR/html-markup/dl.html

I'd also like to use the term "entry" instead of "item" for the child elements (i.e., `DescriptionListEntry`).

roughly:

```
DescriptionList(items: DescriptionListEntry[])

DescriptionListEntry(terms: ListItem[], description: ListItem)
```

mojavelinux:
Something I noticed today that I want to point out. The ListItem for the description node can be nil in the Ruby API. You might need to guard against that if you are walking the tree.

mojavelinux:
I'm fine if you hijack the `getItems()` to give you more control. I agree that `getBlocks()` should return null (or throw an exception) since I'm trying to move away from that in core too. (All the converters use items instead of blocks. See https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/converter/html5.rb#L503).

robertpanzer:
Thanks for the feedback!!!

I renamed `DefinitionList` to `DescriptionList` and `DefinitionListItem` to `DescriptionListEntry`.
Also rebased on top of branch asciidoctorj-1.6.0 again.

Another thing that is missing at the moment is factory methods for `DescriptionList` and `DescriptionListEntry` in the `Processor` base class.
But these methods are also missing for `List` and `ListItem` at the moment, so I'd like to resolve this with another PR.

mojavelinux:
Nice work!

robertpanzer:
Thank you! 
BTW there is a spectacular crash on travis with OpenJDK7 in the native code of InetAddress.getLocalHost().
Not that boring Rubygem proxy timeout stuff ;-)

Issue #401, opened 2015-11-03
asciidoctorj-diagram 1.3.0 does not work on Windows with JRuby 1.7.22


robertpanzer:
When trying to render the test document from the asciidoctorj-diagram on Win7 via the distribution rendering fails with an exception:
```
c:\work\asciidoctorj-1.5.3.1\bin>asciidoctorj -r asciidoctor-diagram test.adoc
io/console not supported; tty will not be manipulated
Exception in thread "main" org.jruby.exceptions.RaiseException: (Errno::ENOENT) C:\work\asciidoctorj-1.5.3.1\lib\asciidoctorj-diagra
m-1.3.0.jar!\gems\asciidoctor-diagram-1.3.0\lib\asciidoctor-diagram\blockdiag.rb
        at org.jruby.RubyFile.realpath(org/jruby/RubyFile.java:840)
        at RUBY.require_relative(file:/C:/work/asciidoctorj-1.5.3.1/lib/jruby-complete-1.7.22.jar!/jruby/kernel19/kernel.rb:11)
        at RUBY.(root)(C:/work/asciidoctorj-1.5.3.1/lib/asciidoctorj-diagram-1.3.0.jar!/gems/asciidoctor-diagram-1.3.0/lib/asciidoct
or-diagram/blockdiag.rb:2)
        at org.jruby.RubyKernel.require(org/jruby/RubyKernel.java:1040)
        at RUBY.(root)(/C:/work/asciidoctorj-1.5.3.1/lib/jruby-complete-1.7.22.jar!/META-INF/jruby.home/lib/ruby/shared/rubygems/cor
e_ext/kernel_require.rb:1)
        at RUBY.require(/C:/work/asciidoctorj-1.5.3.1/lib/jruby-complete-1.7.22.jar!/META-INF/jruby.home/lib/ruby/shared/rubygems/co
re_ext/kernel_require.rb:54)
        at org.jruby.RubyKernel.require(org/jruby/RubyKernel.java:1040)
        at RUBY.(root)(C:/work/asciidoctorj-1.5.3.1/lib/asciidoctorj-diagram-1.3.0.jar!/gems/asciidoctor-diagram-1.3.0/lib/asciidoct
or-diagram.rb:1)
        at RUBY.(root)(/C:/work/asciidoctorj-1.5.3.1/lib/jruby-complete-1.7.22.jar!/META-INF/jruby.home/lib/ruby/shared/rubygems/cor
[...]

PR #410, opened 2015-11-20
Fixes #405. Adds integration tests to asciidoctorj-distribution.

### LINKED ISSUE #405: Add integration tests for the distribution
Apparently JRuby behaves differently when loading gems that are exploded in the file system or when they are packaged in a jar.
The current unit tests test the pdf converter and the diagram extension only with exploded gems.

Tests should be added that render a PDF and diagrams using the jars built by asciidoctorj-diagram and asciidoctorj-pdf.

robertpanzer:
#401 has shown that JRuby behaves differently when loading gems from a jar or from an exploded directory.
The current test cases for diagram and pdf only refer to the exploded gems which does not reflect the way gems are loaded by a user, i.e. "in production".

I added a test case to asciidoctorj-distribution that creates a ditaa diagram, very similar to the test in asciidoctorj-diagram.
But as asciidoctorj-diagram is referenced as a dependency, the gem will be loaded from a jar.

#401 can be reproduced with this PR when setting the JRuby version to 1.7.22.

### DIFF
--- a/asciidoctorj-distribution/build.gradle
+++ b/asciidoctorj-distribution/build.gradle
@@ -2,6 +2,9 @@ dependencies {
   compile project(':asciidoctorj')
   compile project(':asciidoctorj-epub3')
   compile project(':asciidoctorj-pdf')
+  compile project(':asciidoctorj-diagram')
+
+  testCompile project(':asciidoctorj-arquillian-extension')
 }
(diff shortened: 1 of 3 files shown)


mojavelinux:
:+1: Real tests FTW!

PR #420, opened 2016-01-10
resolves #419 isolate asciidoctorj cli from system gems by default

### LINKED ISSUE #419: asciidoctorj cli should run in isolation by default
The asciidoctorj cli command should run in isolation from the system Ruby. Currently, it can see system-wide gems (or gems visible to the active gemset in RVM).

If no load paths are explicitly specified, AsciidoctorJ is initialized using:

```java
asciidoctor = JRubyAsciidoctor.create();
```

This causes AsciidoctorJ to see system-wide gems. If these gems were installed with MRI and use native extensions, this can lead to warnings and unexpected behavior.

The invoker should instead be initialized using:

```java
asciidoctor = JRubyAsciidoctor.create((String) null);
```

This will allow the asciidoctorj cli to only see the gems available on the classpath.

mojavelinux:


mojavelinux:
I'm not sure how to write a test for this, unless we use mocks and verify that JRubyAsciidoctor.create is invoked with a null argument. We could also access the JRubyRuntimeContext, but my understand is that is going away in 1.6.0.

robertpanzer:
JRubyRuntimeContext isn't gone on 1.6.0, it just got better!:)
Instead of holding a singleton Ruby instance that is used only be the first Asciidoctor instance you now can get the Ruby runtime used by any Asciidoctor instance via: 
```java
Ruby rubyInstance = JRubyRuntimeContext.get(asciidoctor);
```

I am pretty sure you already tested that the change works and has the desired effect.
I am just wondering about this: https://github.com/asciidoctor/asciidoctorj/blob/asciidoctorj-1.6.0/asciidoctorj-core/src/main/java/org/asciidoctor/internal/EnvironmentInjector.java#L19-L39
It gets the System environment and creates a `GEM_HOME` configuration variable in case a `GEM_PATH` is defined.
And this is called from `JRubyAsciidoctor.createJRubyAsciidoctorInstance()` independent of the `gemPath` method parameter.

But maybe I am interpreting this wrongly (didn't sleep well :wink:)

mojavelinux:
> Instead of holding a singleton Ruby instance that is used only be the first Asciidoctor instance you now can get the Ruby runtime used by any Asciidoctor instance 

Aha! Now I remember.

This change is still necessary for 1.5.x due to the (somewhat mixed up) way that we instantiate the Asciidoctor instance. If you don't pass any arguments to create, the environment injector is never used. This may have changed in 1.6.0, I didn't check. If so, great! It really should do this by default and only use the system gems if explicitly requested.

robertpanzer:
I think I'd prefer to to test via the associated Ruby runtime, as we cannot tell if JRuby possibly picks up environment vars itself.
A way to test could be to set a GEM_PATH environment variable in gradle, thus testing in an "evil" environment and then check `ENV['GEM_PATH']`.
Depending on if `Asciidoctor.create()` or `Asciidoctor.create((String) null)` was called I see different results.
And in case someone sees this error again for example on JRuby 9.0.4.0 we could simply retest with the upgraded JRuby version.

If you're ok with that I could prepare a PR for your PR tomorrow.

mojavelinux:
:+1: Sounds good. As long as we ensure JRuby is not looking at system gems (which, honestly, I just don't understand what the use case is for that)...then we're solid!

mojavelinux:
Isolation FTW!

robertpanzer:
I'll pull that over into the 1.6.0 branch.

mojavelinux:
Excellent!

PR #422, opened 2016-01-10
update version of artifacts in README

mojavelinux:


robertpanzer:
Thank you.
Sorry for missing this.

mojavelinux:
Where possible, we should see if we can get the release task to handle this. That's part of the "automate all the things" goal in 2016 :)

PR #425, opened 2016-01-10
resolves #424 don't set doctype by default in asciidoctorj cli

### LINKED ISSUE #424: asciidoctorj cli should not hardcode the doctype
The asciidoctorj cli should not hardcode the doctype to article by default. The default should be unset.

By setting the doctype attribute from the cli automatically, this prevents the document from being able to define the doctype. As a result, it's not possible to use asciidoctorj on a mixed set of documents (documents with different doctypes).

To demonstrate this problem, run the AsciidoctorInvoker with no flags defined on a document that defines its doctype as book:

```
= Title
:doctype: book

= Part 1

== Chapter 1

It was a dark and stormy night.
```

This will result in the following warning:

```
asciidoctor: ERROR: book.adoc: line 4: only book doctypes can contain level 0 sections
```

mojavelinux:


robertpanzer:
Thanks!
I'll also cherry-pick this into the 1.6.0 branch.

robertpanzer:
Spectacular crashes on Travis at the moment btw....

mojavelinux:
I noticed the crashes on Travis. Not sure what's going on there.

Thanks for pulling it into the 1.6.0 branch. I'm only submitting to master what applies to both 1.5.x and 1.6.0. If it's 1.6.0 only, then I won't send it to master.

robertpanzer:
> Thanks for pulling it into the 1.6.0 branch. I'm only submitting to master what applies to both 1.5.x and 1.6.0. If it's 1.6.0 only, then I won't send it to master.

That's perfect! :+1:

PR #426, opened 2016-01-14
Fix for Issue 419

### LINKED ISSUE #419: asciidoctorj cli should run in isolation by default

robertpanzer:
Fixes #419 for the 1.6.0 branch.
Contains the same commits as #420.
Made a new PR as the tests revealed a new error in the 1.6.0 branch that I want tested on the CI system before merging: 
A null gemPath passed to `Asciidoctor.Factory.create((String) null)` was completely ignored so that a `GEM_PATH` still leaked into the JRuby instance!

### DIFF
--- a/asciidoctorj-core/build.gradle
+++ b/asciidoctorj-core/build.gradle
@@ -56,3 +56,49 @@ jar {
       '*'
   }
 }
+
+test {
+  useJUnit {
+    excludeCategories 'org.asciidoctor.categories.Polluted'
+  }
+}
+
+task pollutedTest(type: Test) {
+  useJUnit {
+    includeCategories 'org.asciidoctor.categories.Polluted'
+  }
+  forkEvery = 10
+  minHeapSize = '128m'
+  maxHeapSize = '1024m'
+  if (JavaVersion.current().isJava8Compatible()) {
+    jvmArgs '-XX:-UseGCOverheadLimit'
+  }
+  else {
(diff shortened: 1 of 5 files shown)


mojavelinux:
> A null gemPath passed to Asciidoctor.Factory.create((String) null) was completely ignored so that a GEM_PATH still leaked into the JRuby instance!

This might be when we get in touch with the JRuby team and see if we can get a proper API for isolating from the system gems. This is a critical use case and I think JRuby should support it natively.

robertpanzer:
This error was more in our code:
When passing a null gemPath to `Asciidoctor.Factory.create()` the `GEM_PATH` was not set to `null` in the environment for the `RubyInstanceConfig` but it stayed unset, thereby falling back to the default behavior.

mojavelinux:
Ah, gotcha. I still think it's a pretty big hack that we have to do to mess with the environment to isolate. I really think JRuby should handle this for us (in an ideal world).

Glad to here you got to the bottom of the issue, though!

PR #427, opened 2016-01-25
Upgraded JRuby dependency to 1.7.24 and added integration test to asc…

robertpanzer:
…iidoctorj-distribution.

When setting the JRuby dependency to 1.7.23 the integration test will fail.
But it runs successfully with the version previously used, 1.7.21, and also with the brand new 1.7.24.

### DIFF
--- a/asciidoctorj-distribution/src/test/java/org/asciidoctor/WhenDitaaDiagramIsRendered.java
+++ b/asciidoctorj-distribution/src/test/java/org/asciidoctor/WhenDitaaDiagramIsRendered.java
@@ -1,0 +1,86 @@
+package org.asciidoctor;
+
+import org.junit.ClassRule;
+import org.junit.Test;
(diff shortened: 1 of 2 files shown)


mojavelinux:
\o/

robertpanzer:
:beers: 
Maybe this can save us some roundtrips we had to go in the past with the last 1.5.x releases.

Also tested locally in an integration test that plantuml diagrams work.
Unsure yet how to integrate this test so that it can be run on the CI system. 
Maybe install the Windows or Linux graphviz binaries in the gradle scripts and set the PATH appropriately?

mojavelinux:
> Maybe install the Windows or Linux graphviz binaries in the gradle scripts and set the PATH appropriately?

Another option is to use the Travis CI / AppVeyor configuration to install additional libraries.

PR #431, opened 2016-02-16
Fixes #430. Preserve attribute insertion order.

### LINKED ISSUE #430: Support to unset attributes in AttributesBuilder
AttributesBuilder has the `arguments` [method](https://github.com/asciidoctor/asciidoctorj/blob/master/asciidoctorj-core/src/main/java/org/asciidoctor/AttributesBuilder.java#L578) that allows to pass a chain of command line attributes.
It allows to update attributes that have been set previously. But when you want to unset an attribute it just adds a new entry in the internal map instead of updating the current one. 
Even if Asciidoctor (Ruby core) deals with it properly, I think it should be better to remove the attribute and add the new unset one.

As example, the next code
```
AttributesBuilder builder = AttributesBuilder.attributes()
builder.arguments("stylesheet=my-theme.css toc").print()
builder.arguments("stylesheet=my-theme2.css toc=left").print()
builder.arguments("data-uri mysample.adoc").print()
builder.arguments("stylesheet!").print()
builder.unsetStyleSheet().print()
```
generates the following map of attributes, you can see that stylesheet appears twice.
```
stylesheet -> my-theme2.css
stylesheet! -> ""
mysample.adoc -> ""
data-uri -> ""
toc -> left
```

robertpanzer:


robertpanzer:
Took about 3 attemps to make the appveyor build pass because the graphviz site only works randomly.

Will cherry-pick this into the 1.6.0 branch.

PR #463, opened 2016-05-29
update gradle wrapper to 2.14

ancho:
* fixed publishing with local signing (jars and pom) and upload to bintray

  This is a way to complicated. nebula publishing dropped the signing stuff after
  bintray makes it possible to sign the artifacts automatically with their key or
  with a private/public keypair linked and stored within the repository.

  See https://github.com/nebula-plugins/nebula-publishing-plugin/issues/61 and
  https://bintray.com/docs/usermanual/uploads/uploads_gpgsigning.html

* add singned pom to publication only if not skipped.

* set name and description for pom

* fallback to buildscript block to apply plugins only to the subprojects


ancho:
oh no...the assemble task does not run on the ci server because it has no configured signatory. i'll fix that.

ancho:
This time it's the test-asciidoctor-upstream script that fails. The tests succeeded.

ancho:
The failures with openjdk may relate to https://github.com/travis-ci/travis-ci/issues/5227

build.gradle
```diff
@@ -74,31 +77,10 @@ subprojects {
...
-    }
-    maven {
-      name 'rubygems-prerelease'
-      url 'http://rubygems-proxy.torquebox.org/prereleases'
-    }
-  }
 
-  plugins.withType(JavaPlugin) {
-
-    project.tasks.withType(JavaCompile) { task ->
```
robertpanzer: I think this part has to stay.
I did a clean build and Gradle still seems to ignore this:
```
  sourceCompatibility = targetCompatibility = JavaVersion.VERSION_1_6
```
If I invoke this the output after building with Java 8 the class still seems to be built for Java 8 even with Gradle 2.13:
```
$ hexdump asciidoctorj-core/build/classes/main/org/asciidoctor/Asciidoctor.class|head
0000000 ca fe ba be 00 00 00 34 00 52 07 00 4e 07 00 4f
```

robertpanzer:
I think the failing OpenJDK test can be ignored, they are already failing for a couple of months now.
Crazy that this still isn't fixed. Thanks for finding this issue though!!!

Regarding the upstream test I also tested on the current master which worked recently.
Maybe a asciidoctor 1.5.5.dev is really not available atm.

robertpanzer:
I am wondering about the signing process.
If I read the documentation correctly automatic signing by Bintray has to be activated for the whole Maven repository of the Asciidoctor organization, which already contains 13 different packages.
I don't know if we can simply activate this for all packages in there like asciidoclet, etc.
(I don't seem to have the rights for that)

@mojavelinux Do you have some thoughts on this?

I would definitely love if the signing process could be simplified.

robertpanzer:
I would definitely love to see an upgrade of Gradle!!!!

robertpanzer:
Maybe I got the description wrong, when doing a release build I get valid signatures :-)

ancho:
Hmm...the compatibility stuff drives me crazy. Don't know why it doesn't work for subprojects.
I think it's best to rollback to the workaround and hope the best for gradle 2.14 though.
Very weird....

And yes. The local signing works as expected. The only change is the switch from nebula-publish to the build-in maven-publish, combined with the maven and signing plugin.

robertpanzer:
I just checked that the classes are now compiled for Java 1.6 again.

@mojavelinux Could you take a look at this? You spent so much time on the publishing plugin that you can probably judge better.

robertpanzer:
Squashed and cherry-picked into the 1.6.0 branch and made an alpha.2 release: 9ac7714a4fc4d5ee70eae0bbbdcaa25c13ec0105.

Would be great if you could have a look at it.

What I found a bit strange is that the release goes immediately live. 
For the past releases I always had to confirm the new artifacts.
Could this be due to the configuration of the publishing plugin or could this be a change at Bintray?

robertpanzer:
Just seen that in the build that I moved over to the 1.6.0 branch a pom.xml got generated for the distribution which overwrites the original asciidoctorj pom.xml.
Therefore asciidoctorj has deps now on asciidoctorj-diagram, asciidoctorj-pdf and asciidoctorj-epub3.
Will take a look tomorrow what's gone wrong.

robertpanzer:
@ancho Would be great if you could help out on this and see what I've done wrong on the AsciidoctorJ 1.6.0 branch. 
My knowledge about Gradle publishing is too limited.
Major differences to the current master are the support for Snapshot deployments and the documentation module which is not published.

I can only guess the hell @mojavelinux went through when crafting the initial publishing for AsciidoctorJ.

ancho:
Sorry for answering so late. I'll have a look at it to find a solution. It is strange that the jar for the distribution project is not created, as the jar task is disabled but the pom get generated anyways.

The auto publish is configured in the [deploy.gradle file](https://github.com/ancho/asciidoctorj/blob/12f7696c98e54469b01b8bb2196d29dcc27a7ebc/gradle/deploy.gradle#L13)
Maybe it didn't work before. I didn't change that part of configuration.

robertpanzer:
Thanks! That would be cool.
It is very likely that I've done sth wrong, but I'm not really proficient with this.

You can also see the result when you take a look at the pom here: https://jcenter.bintray.com/org/asciidoctor/asciidoctorj/1.6.0-alpha.2/asciidoctorj-1.6.0-alpha.2.pom

ancho:
I found the problem for the override of the asciidoctorj pom with the pom form the distrbution.
To prevent it I do not apply the publishing stuff to the distribution package.
There is no need anyway, as the distribution package is not a jar with dependencies on it.

But I couldn't figure out what is wrong with the asciidoctorj-1.6.0 configuration yet ;(

I will continue to spot the differences in the evening.

ancho:
I tested it the following way:

* created a maven repsotory at bintray  called experiment
* `export BINTRAY_REPO=experiment; ./gradlew -PbintrayPersonal bU`

See: https://bintray.com/ancho/experiment/asciidoctorj/1.5.5/view#files/org/asciidoctor/asciidoctorj/1.5.5

It is not synchronized with jcenter or mavencentral. Don't panik :)

robertpanzer:
Hehe, no panic :-)

Yep, for the master branch the pom looks correct.
Only for the stuff that I probably broke on the 1.6.0 branch the pom is apparently overwritten by a pom created for the distribution.

mojavelinux:
Keep up the great work!

robertpanzer:
Thanks for all the hard work on that! 🍻 
This is really awesome!
I will also upgrade the 1.6.0 branch to Gradle 2.14 and the Bintray plugin 1.7.

Strange thing though is it seems like artifacts are uploaded twice: https://gist.github.com/robertpanzer/4c020ec0a2ef5e2fed63f596a2ddf59c
(And it also takes the time to upload the files)
But shouldn't be a problem as long as in the end the right files are in the repo.

ancho:
That's really strange and should be avoided. o.O 
I'll check that an see what I can do...
Thanks for the appreciation, I'm just glad to help out and get the chance to give something back ;)

PR #469, opened 2016-06-06
Asciidoctorj 1.6.0 - pom generation fix

ancho:
I think I found a solution to make the pom generation work again.
Something is strange. I should not have to add the publishing configuration block into afterEvaluate.
But it works that way with the current configuration in the 1.6.0 branch.
Before the pom where generated too, but without the dependencies block o.O


robertpanzer:
Thanks!!!!

This did the trick!

robertpanzer:
Thanks, POM generation seems to be good now!

Unfortunately we have another problem now :-) #470 
@ancho If you could have another look at this.

mojavelinux:
You're beginning to discover why we had such a unique combination of plugins in the build ;) It's an art of sidestepping bugs.

PR #482, opened 2016-07-14
fix pom generation for release versions to prevent adding a pom packaging tag

### LINKED ISSUE #481: Packaging type is set to pom instead of jar in generated POM
The POM file generated by the publishing plugin incorrectly sets the packaging type to `pom` instead of `jar`. We previously had a patch to reverse the incorrect behavior in Gradle, but that patch seems to have been removed.

See the difference between the following two POM files (before and after the change):

* before: https://repo1.maven.org/maven2/org/asciidoctor/asciidoctorj/1.5.4/asciidoctorj-1.5.4.pom
* after: https://repo1.maven.org/maven2/org/asciidoctor/asciidoctorj/1.5.4.1/asciidoctorj-1.5.4.1.pom

This also affects the 1.6.0 branch.

ancho:
fixes #481

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,11 @@ jdk:
   - openjdk6
 before_script: unset GEM_PATH GEM_HOME JRUBY_OPTS
 before_install:
+  - sudo hostname "$(hostname | cut -c1-63)"
+  - (cat /etc/hosts ; echo 127.0.0.1 $(hostname)) | sudo tee /etc/hosts
   - sudo apt-get -qq update
   - sudo apt-get install -y graphviz
-install: ./gradlew assemble
+install: ./gradlew -S assemble
 script: ./gradlew -S check && bash test-asciidoctor-upstream.sh
 notifications:
   email: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,8 @@ before_install:
   - (cat /etc/hosts ; echo 127.0.0.1 $(hostname)) | sudo tee /etc/hosts
   - sudo apt-get -qq update
   - sudo apt-get install -y graphviz
-install: ./gradlew -S assemble
-script: ./gradlew -S check && bash test-asciidoctor-upstream.sh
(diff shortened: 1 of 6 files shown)


robertpanzer:
I am inclined to merge this.
The Java7 build failed on Appveyor for no apparent reason, there are not even log files.
The OracleJDK builds passed on Travis (OpenJDK fails already for a longer time, there's a problem in the Travis machine setup).
The travis build tries to also build against the next dev version of asciidoctor, which would be 1.5.5.dev. As this is not available the second build fails after the first successful build.

ancho:
Pleas wait. There is a fileSpec variant to add custom files to the bintray repository. If it works we have a cleaner version without manipulating the pom.

ancho:
Here we go. This should do the trick. Test with my experimental maven repository on bintray.

abelsromero:
I think that the build process needs to be documented at some point.

abelsromero:
I also tested on MacOS and worked fine.

mojavelinux:
> The travis build tries to also build against the next dev version of asciidoctor, which would be 1.5.5.dev. As this is not available the second build fails after the first successful build.

See #484.

mojavelinux:
It's really important that we get to a green build. Otherwise, we're going to miss errors as they come into master.

robertpanzer:
Restarted the CI build on travis and the Oracle JDK builds are green now.

ancho:
travis has still problems with openjdk builds. See https://github.com/travis-ci/travis-ci/issues/5227

So maybe we can try the workaround and manipulate the /etc/hosts file.

```
before_install:
  - sudo hostname "$(hostname | cut -c1-63)"
  - (cat /etc/hosts ; echo 127.0.0.1 $(hostname)) | sudo tee /etc/hosts
```

ancho:
Hurray! :tada:
Everything is green again \o/

robertpanzer:
Thanks a lot! 🍻

PR #484, opened 2016-07-14
resolves #483 fix test against upstream

### LINKED ISSUE #483: configure travis to run only the check task for PR's
Currently every PR seems to fail on travis. In reality the check task is successful. But the `bash test-asciidoctor-upstream.sh` tasks fails.

See https://travis-ci.org/asciidoctor/asciidoctorj/jobs/144690217

Is it necessary to run the bash script for a pull request?

mojavelinux:
* update script to handle new gemspec contents
* only run upstream test on first job in build

### DIFF
--- a/test-asciidoctor-upstream.sh
+++ b/test-asciidoctor-upstream.sh
@@ -1,11 +1,11 @@
 #!/bin/bash
 
-# This script runs the AsciidoctorJ tests against the specified tag (or master) of the Asciidoctor RubyGem.
+# This script runs the AsciidoctorJ tests against the specified tag (or master) of the Asciidoctor Ruby gem.
 
 GRADLE_CMD=./gradlew
-#if [ ! -z $TRAVIS ]; then
-#  GRADLE_CMD=gradle
-#fi
+if [ ! -z $TRAVIS_JOB_NUMBER ] && [ "$TRAVIS_PULL_REQUEST" != "false" ] && [ "${TRAVIS_JOB_NUMBER##*.}" != '1' ]; then
+  exit 0
+fi
 # to build against a tag, set TAG to a git tag name (e.g., v1.5.2)
 TAG=master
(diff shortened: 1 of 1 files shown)


mojavelinux:
This pull request solves some errors, but reveals others. It will also need to be applied against the 1.6.0 branch.

robertpanzer:
Strange thing is that the upstream build worked for me locally also before this PR. 
Great that the travis is green again finally!

robertpanzer:
Cherry-picked into asciidoctorj-1.6.0 as f27691fe8c9537cf88a802c0a38166f2e624e394

mojavelinux:
Almost green ;)

I want to clarify that this was an issue related specifically to JRuby 1.7, which doesn't implement a method option I was using the the gemspec. However, I applied a fix that will work for all Rubies, so it should be very safe now. :+1:

mojavelinux:
It appears we get a test failure on AsciidoctorJ on AppVeyor. We should enable the flag that reports the test failures on the commandline so we can see them. The other option is to define the test reports as artifacts, which AppVeyor makes available in the workspace after the build.

For example:

```
artifacts:
  - path: **/build/reports/tests/**
```

robertpanzer:
It seems to fail here: https://github.com/asciidoctor/asciidoctorj/blob/asciidoctorj-1.6.0/asciidoctorj-distribution/src/test/groovy/org/asciidoctor/diagram/WhenDitaaDiagramIsRendered.groovy#L84

Seems like the first test or the when has failed so that `png` is not initialized. Wonder why Spock doesn't report that.
Could be this one: https://github.com/jruby/jruby/issues/3957
This test was specifically made to catch such errors :-)

mojavelinux:
> This test was specifically made to catch such errors :-)

At least we can take comfort in the fact that its a meaningful failure ;)

PR #485, opened 2016-07-16
Publishing fixes for the 1.6.0 branch 

robertpanzer:
This PR tries to get the recent changes/fixes to the publishing originally done by @ancho on the master branch also to the asciidoctorj-1.6.0 branch.
Instead of cherry picking the multiple commits I tried to apply them manually, would be great if somebody could review.

Also cherry-picked the checkVersion build steps.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,8 @@ before_install:
   - (cat /etc/hosts ; echo 127.0.0.1 $(hostname)) | sudo tee /etc/hosts
   - sudo apt-get -qq update
   - sudo apt-get install -y graphviz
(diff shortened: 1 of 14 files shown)


ancho:
I tested the configuration. It works as expected.

* bintrayUpload
  * with and without signing
* checkVersion
  * skip works
* generated pom is clean 
  * with and without SNAPSHOT version
* the build is obviously green for travis
* appveyor does not seem to have a problem with the build configuration, but has a failing test.
   so it's kind of ok in that case.

robertpanzer:
Thanks for reviewing this!

PR #489, opened 2016-07-18
fix parameter names in internal Ruby API

mojavelinux:
- use filename instead of content where appropriate
- replace tabs with spaces
- drop unnecesary empty brackets for methods that take no parameters


mojavelinux:
This change should not impact any behavior. It's just a technical correction.

robertpanzer:
Thanks! :-)

PR #490, opened 2016-07-19
Fixes 486. Added methods ContentNode.hasAttr to check if a node has a…

robertpanzer:
…n attribute with whatever value


asciidoctorj-core/src/main/java/org/asciidoctor/ast/ContentNode.java
```diff
@@ -40,6 +40,20 @@
     Object getAttr(Object name, Object defaultValue, boolean inherit);
     Object getAttr(Object name, Object defaultValue);
     Object getAttr(Object name);
+
+    /**
+     * @param name
+     * @return {@code true} if this node or one of its parent has an attribute with the given name
```
mojavelinux: Actually, there is only one level of inheritance. It is either the node or the document. Interim nodes are not consulted. This is true for all attribute-related methods.

robertpanzer:
Updated the javadoc

mojavelinux:
:+1:

PR #491, opened 2016-07-21
Make methods on JavaExtensionRegistry return itself instead of void t…

robertpanzer:
…o allow chaining of method calls


abelsromero:
Silly thing, but it always have puzzled me why docinfoProcessor uses Camel when the rest of processors don’t: preprocessor, postprocessor, treeprocessor.
Then we have blockMacro, inlineMacro.

mojavelinux:
These are silly English rules. Preprocessor is a single word, so it goes to all lowercase. You could argue the case for Postprocessor being a single word or two words...but some dictionaries have it as a single word. Treeprocessor is not a real word, so I followed the pattern. For certain, there is no word "docinfoprocessor", so we have to use camel case. The same for the others.

abelsromero:
I had the impression it was something like that, I know you give a lot of thought to those details.
It was obvious for macro method, but not so (for me) for the other ones.

PR #493, opened 2016-07-23
Added method Document.getCounter. Interaction with using counters in …

### LINKED ISSUE #488: Expose the AsciiDoc counter facility as an API
Asciidocotor has built-in support for counter attributes. Provide an API to allow the counter under these attributes to be used. This is particularly useful in extensions.

This method should be added to Document:

* counter(String name, Object seed) - Get the next number in the sequence for this counter, initializing it with the seed value if not already set

We should verify that after a counter is retrieved from the API, it continues to get used when resolved in the AsciiDoc source.

```
{counter:tab-index}
```

There are methods on the Ruby API that can be used to control the internal state.

robertpanzer:
…the Document not possible yet

Fixes #488.

Added methods `Document.getCounter(counterName)` and `Document.getCounter(counterName, initialValue)` to the API.

The interaction with using the {counter:mycounter} macro is not possible though.

If I've understood it correctly the counters are finally stored in an attribute having the name of the counter on the Document.
All Document attributes are restored after for example BlockMacros are called in Document#parse.
The counter macro however is invoked in the conversion phase by the Substitutor after the attributes have been reset. Therefore the counter restarts at 1 again.

The first test reflects this behavior.

With a TreeProcessor it works a bit better in so far as the document can continue using the counters as the reset happens before the invocation of the Treeprocessors.
But I'd guess that when having a counter in the document, then reusing the same counter in a TreeProcessor will result in the counter sequence 2, 1.

When using counters both in a processor and the document, I think it would be better to make the processor insert the text `{counter:foocounter}` instead of invoking the method on the Document.

To wrap it up I wonder if it makes sense at all to invoke this API from an extension and if it has already been used by Ruby extensions.
(Maybe I completely missed a point, that's also perfectly possible)

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/ast/Document.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/ast/Document.java
@@ -58,4 +58,23 @@ public interface Document extends StructuralNode {
      * @return options defined in document.
      */
     Map<Object, Object> getOptions();
+
+    /**
+     * Gets the current counter with the given name and increases its value.
+     * At the first invocation the counter will return 1.
+     * After the call the value of the counter is set to the returned value plus 1.
+     * @param name
+     * @return
+     */
+    int getAndIncrementCounter(String name);
+
+    /**
+     * Gets the current counter with the given name and increases its value.
+     * At the first invocation the counter will return the given initial value.
+     * After the call the value of the counter is set to the returned value plus 1.
+     * @param name
+     * @param initialValue
+     * @return
+     */
+    int getAndIncrementCounter(String name, int initialValue);
 }
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/ast/ContentModel.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/ast/ContentModel.java
@@ -53,32 +53,32 @@ public @interface ContentModel {
     /**
      * Predefined constant to let Asciidoctor know that this BlockProcessor creates simple paragraph content.
      */
-    public static final String SIMPLE =":simple";
+    public static final String SIMPLE = ":simple";
 
     /**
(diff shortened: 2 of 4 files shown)


asciidoctorj-core/src/main/java/org/asciidoctor/ast/Document.java
```diff
@@ -58,4 +58,20 @@
...
+    /**
+     * Gets the current counter with the given name and increases its value.
+     * @param name
+     * @return
+     */
+    int getCounter(String name);
+
+    /**
+     * Gets the current counter with the given name and increases its value.
+     * At the first invocation
```
mojavelinux: It seems like the rest of this sentence is missing.

asciidoctorj-core/src/main/java/org/asciidoctor/ast/Document.java
```diff
@@ -58,4 +58,20 @@
      * @return options defined in document.
      */
     Map<Object, Object> getOptions();
+
+    /**
+     * Gets the current counter with the given name and increases its value.
+     * @param name
+     * @return
+     */
+    int getCounter(String name);
```
mojavelinux: I think these methods should be named getAndIncrementCounter so we clearly communicate that this is a mutable operation.

robertpanzer:
Renamed `getCounter()` to `getAndIncrementCounter()` and fixed javadoc.

mojavelinux:
I think it's understood that these counters won't affect parsing/conversion since they are being used in a different phase. The API can leverage the counter mechanism for its own purposes. In some cases, extensions may be able to influence the attribute in the document, but as you have pointed out, it's highly dependent on when it is invoked. I'm okay with that as the API current stands. That's how it works in Ruby too. In the future, we may evolve this.

PR #494, opened 2016-07-25
Fixes #492. Deprecate Processor.CONTENT_MODEL_* and moved @ContentMod…

### LINKED ISSUE #492: Deprecate content model constants on Processor
Now that the content model constants are defined and organized in the ContentModel class, I don't think we need the equivalent constants on the Processor class anymore. We should either remove the constants on the Processor class or at least mark them as deprecated.

For example, the constant for the the compound content model is defined in these two places:
- Processor.CONTENT_MODEL_COMPOUND
- ContentModel.COMPOUND

I also think that ContentModel should be moved to the ast package.

robertpanzer:
…el to org.asciidoctor.ast.

Also fixed a minor formatting bug in the integrator guide.

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/extension/Processor.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/extension/Processor.java
@@ -34,48 +34,65 @@ public class Processor {
      * <pre>
      * <verbatim>
      * Map&lt;String, Object&gt; config = new HashMap&lt;&gt;();
-     * config.put(CONTENT_MODEL, CONTENT_MODEL_COMPOUND);
+     * config.put(CONTENT_MODEL, ContentModel.COMPOUND);
      * BlockProcessor blockProcessor = new BlockProcessor("foo", config);
      * asciidoctor.javaExtensionRegistry().block(blockProcessor);
      * </verbatim>
      * </pre>
      * </p>
+     *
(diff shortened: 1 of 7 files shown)


asciidoctorj-core/src/main/java/org/asciidoctor/extension/Processor.java
```diff
@@ -40,42 +40,59 @@
      * </verbatim>
      * </pre>
      * </p>
+     *
+     * @deprecated Use the annotation {@link ContentModel} instead to describe this Processor
      */
+    @Deprecated
     public static final String CONTENT_MODEL = "content_model";
```
mojavelinux: I don't think there's an equivalent for this constant, it there? Perhaps we can move it to ContentModel.KEY? or something.

robertpanzer: Totally right.
I was probably thinking that the whole manual configuration should be deprecated and happen via annotations where also deprecating this made sense.
I will "undeprecate" it again :-)

mojavelinux: In most cases I agree annotations are the preferred approach. It comes in
when constructing a tree manually and building options for the node. (That
could probably be improved too, but a separate issue).

robertpanzer:
@mojavelinux Dou you think it is ok like this now?

mojavelinux:
Yep, I really like this approach. :+1:

PR #496, opened 2016-08-02
added sdkman verndors plugin

ancho:
Configure your local gradle.properties file with the following credentials.

* sdkman_consumer_key=_some_key_
* sdkman_consumer_token=_some_token_

Adds the following Tasks to the configuration if project status is 'release'.
See project.statusIsRelease.

* _sdkAnnounceVersion_ - Announce a Release on SDKMAN.

* _sdkDefaultVersion_ - Make an existing Candidate Version the new Default on SDKMAN.

* _sdkReleaseVersion_ - Release a new Candidate Version on SDKMAN.

* _sdkMajorRelease_ - Convenience task performs a Major Release consisting of Release, Default and Announce combined.

* _sdkMinorRelease_ - Convenience task performs a Minor Release consisting of Release and Announce combined

See https://github.com/sdkman/sdkman-vendor-gradle-plugin for more information.

### DIFF
--- a/asciidoctorj-distribution/build.gradle
+++ b/asciidoctorj-distribution/build.gradle
@@ -1,5 +1,9 @@
 apply plugin: 'application'
 
+if (project.statusIsRelease) {
+    apply from: rootProject.file('gradle/sdkman.gradle')
+}
+
 dependencies {
   compile project(':asciidoctorj')
   compile project(':asciidoctorj-diagram')
--- a/build.gradle
+++ b/build.gradle
@@ -7,10 +7,14 @@
 buildscript {
     repositories {
         jcenter()
+        maven {
+            url "https://plugins.gradle.org/m2/"
+        }
     }
     dependencies {
         classpath "com.github.jruby-gradle:jruby-gradle-plugin:0.1.17"
         classpath "com.jfrog.bintray.gradle:gradle-bintray-plugin:1.7"
+        classpath "gradle.plugin.io.sdkman:gradle-sdkvendor-plugin:1.1.0"
     }
 }
 
@@ -71,7 +75,7 @@ subprojects {
     }
     apply from: rootProject.file('gradle/deploy.gradle')
   }
-  
+
   status = _status
 
   // NOTE sourceCompatibility & targetCompatibility are set in gradle.properties to meet requirements of Gradle
--- a/gradle/sdkman.gradle
+++ b/gradle/sdkman.gradle
@@ -1,0 +1,10 @@
+apply plugin: "io.sdkman.vendors"
+
+sdkman {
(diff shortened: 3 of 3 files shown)


ancho:
What do you think about a _release_ task that runs the bintrayUpload task and the sdkMajorRelease task afterwards if successful?

robertpanzer:
I am unsure about that.
Most of the times I have to manually publish the artifacts in Bintray, which is not too bad as it allows me to check them again.
I think I'd prefer to do that step first and once I am sure this was fine do the sdkman publish afterwards.

ancho:
Well....then...this is it. All you need is the token and the key.
But now I'm curious what exactly makes you unsure about the bintray publishing process. Is there anything you check manually that couldn't be automated?

robertpanzer:
It's more about checking again that the files are correct, like the pom etc. 
Too many surprises recently that you phenomenally resolved:)

ancho:
Ok. Fair enough. 
Maybe we can collect some common trust issues with the build that we can transform in a pre-check that the publish task depends on to gain trust, or something like that.

Don't get me wrong. I don't think everything needs to be automated. On the other hand these two tasks could go hand in hand with a single operation. But I'm with you. They don't need to.

robertpanzer:
The URL seems to not be 100% correct, but I can fix that myself.
This will give the correct URL
```
 url = "https://dl.bintray.com/asciidoctor/maven/org/asciidoctor/asciidoctorj/${rootProject.version}/asciidoctorj-${rootProject.version}-bin.zip"
```
This gives asciidoctorj-distribution-1.5.4.1.zip which is not correct:
```
   sdkman.url = "https://dl.bintray.com/asciidoctor/maven/org/asciidoctor/asciidoctorj/$rootProject.version/$distZip.archiveName"
```

Thanks a lot for pushing and helping out on this!!!
AsciidoctorJ 1.5.4.1 should be already available on sdkman.

robertpanzer:
Just to keep it noted somewhere:
I had a misunderstanding of how sdkman works: My idea was that it is an own repository storing the binaries somewhere.
Instead it only links to the zip file on jcenter, so trying to release sth on sdkman that isn't successfully released on jcenter doesn't work anyway.

ancho:
Hmm. I don't understand what happened with the url. My test with older archives were all successful.
Using the original solution and printing out the url in _distributionAvailable_ with version 1.5.4.1 prints 

> :asciidoctorj-distribution:distributionAvailable
> https://dl.bintray.com/asciidoctor/maven/org/asciidoctor/asciidoctorj/1.5.4.1/asciidoctorj-1.5.4.1-bin.zip

Which is the expected and correct url. Strange. But anyways...if it works it works....

PR #498, opened 2016-08-12
Fixes #497. Patch Asciidoctor::AbstractBlock#section? in abstract_blo…

robertpanzer:
…ck.rb to return context by string instead of Symbol because Java blocks always return strings.

There's a problem currently in the way AsciidoctorJ 1.5.x sits on top of Asciidoctor:
It replaces all blocks in the AST once touched by a processor by their Java counterparts, which is necessary.
This makes it impossible for public methods exposed via the Java API to return Ruby specifics like for example Ruby Symbols.

There's a recent change in Asciidoctor 1.5.4 that uses the result of `Asciidoctor::AbstractBlock#context` and compares it to `:section`.
As the Java object returns a String instead of a Symbol this test always fails and therefore the error described in asciidoctor/asciidoctor-maven-plugin#258 and filed in #497 occurs.

This PR fixes this by 
1. patching abstract_block.rb to compare the String representation of the symbol instead of the symbol
2. Implementing the methods in SectionImpl and AbstractBlockImpl necessary to build a table of contents.

Also tested locally building the asciidoctor-maven-plugin.

Issue #497, opened 2016-08-12
Using a Processor makes rendering documents with a TOC fail


robertpanzer:
As first observed in asciidoctor/asciidoctor-maven-plugin#258 when using a Processor that iterates over the blocks of a document, rendering that Document with a table of contents will fail.

That means with this Processor:
```java
public class MyTreeprocessor extends Treeprocessor {
    public TerminalCommandTreeprocessor(Map<String, Object> config) {
        super(config);
    }

    @Override
    public Document process(Document document) {
        final List<AbstractBlock> blocks = this.document.blocks();
        return document;
    }
```
this code will fail:
```java
        JavaExtensionRegistry javaExtensionRegistry = this.asciidoctor.javaExtensionRegistry();
        javaExtensionRegistry.treeprocessor(MyTreeprocessor.class);
        String content = asciidoctor.renderFile(
            classpath.getResource("sample-with-sections.ad"),
            options().toFile(false)
                .attributes(AttributesBuilder.attributes().tableOfContents(true))
                .get());
```

The exception is this:
```
org.asciidoctor.internal.AsciidoctorCoreException: org.jruby.exceptions.RaiseException: (NoMethodError) undefined method `level' for nil:NilClass
	at org.asciidoctor.internal.JRubyAsciidoctor.renderFile(JRubyAsciidoctor.java:345)
	at org.asciidoctor.internal.JRubyAsciidoctor.renderFile(JRubyAsciidoctor.java:434)
	at org.asciidoctor.extension.WhenJavaExtensionIsRegistered.when_using_a_tree_processor_a_toc_should_still_be_created(WhenJavaExtensionIsRegistered.java:401)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
[...]

PR #499, opened 2016-08-13
Add missing method implementations to Java AST classes and fixed some…

robertpanzer:
… misalignments.

As a continuation of #498 this adds more methods missing in the Java AST classes that are required if nodes are touched once by a Treeprocessor.

Additionally all occurrences of .context are now replaced with .context.to_sym.
The contexts of content created by Blockmacros could for example never be handled correctly by the parser.

This PR should also fix the second issue mentioned in #497 .

### DIFF
--- a/asciidoctorj-core/build.gradle
+++ b/asciidoctorj-core/build.gradle
@@ -29,10 +29,9 @@ jrubyPrepareGems << {
     from gemFiles
     eachFile {
       // See https://github.com/asciidoctor/asciidoctorj/issues/497
-      if (it.path ==~ /gems\/asciidoctor-.+\/lib\/asciidoctor\/.*\.rb/) {
+      if (it.path ==~ /gems\/asciidoctor-.+\/lib\/asciidoctor\/abstract_block.rb/) {
         it.filter { line ->
-          line.replaceAll(/\.context /, '.context.to_sym ')
-                  .replaceAll(/\.context\)/, '.context.to_sym)')
+          line.replaceAll(/block.context == :section/, 'block.context.to_sym == :section')
         }
       }
     }
--- a/asciidoctorj-core/build.gradle
+++ b/asciidoctorj-core/build.gradle
@@ -29,9 +29,10 @@ jrubyPrepareGems << {
     from gemFiles
     eachFile {
       // See https://github.com/asciidoctor/asciidoctorj/issues/497
-      if (it.path ==~ /gems\/asciidoctor-.+\/lib\/asciidoctor\/abstract_block.rb/) {
+      if (it.path ==~ /gems\/asciidoctor-.+\/lib\/asciidoctor\/.*\.rb/) {
         it.filter { line ->
-          line.replaceAll(/block.context == :section/, 'block.context.to_sym == :section')
+          line.replaceAll(/\.context /, '.context.to_sym ')
+                  .replaceAll(/\.context\)/, '.context.to_sym)')
         }
       }
     }
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/ast/AbstractBlockImpl.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/ast/AbstractBlockImpl.java
@@ -32,6 +32,10 @@ public class AbstractBlockImpl extends AbstractNodeImpl implements AbstractBlock
         return delegate.getTitle();
     }
 
+    public boolean isTitle() {
+        return RubyUtils.invokeRubyMethod(delegate, "title?", new Object[0], Boolean.class);
+    }
+
     @Override
     public String style() {
         return getStyle();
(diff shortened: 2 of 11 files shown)


mojavelinux:
It might be a lot cleaner to simply monkeypatch the AbstractNode (in src/main/resources/org/asciidoctor/internal/asciidoctorclass.rb) to override the context getter method.

```ruby
module Asciidoctor
    class AbstractNode
        def context
            @context.to_sym
        end
    end
end
```

That still won't fix internal references to `@context`, but it at least addresses all the public uses.

I really don't feel good about modifying the Ruby source code from the build. This eliminates that need.

robertpanzer:
Thanks for the tip!
I'll try that out as soon as I'm sitting in front of a big screen again.
Nevertheless I'd intuitively say that it won't be called for Java nodes as
they don't really care about this class (my Ruby knowledge is very poor
though)

I don't feel good about monkeypatching either.
At the same time I am wondering why we don't have a lot more issues when
working with Extensions.
Many comparisons simply seem to return false always as soon as
Treeprocessors, Blockmacros or Block extensions come into play.

Am Samstag, 13. August 2016 schrieb Dan Allen :

mojavelinux:
> I don't feel good about monkeypatching either.

I don't mind monkeypatching. I just don't like that we are modifying the code in core directly instead of extending/overridding the Ruby code.

> Nevertheless I'd intuitively say that it won't be called for Java nodes as
they don't really care about this class

This is probably my misunderstanding of how 1.5.x works, but as far as Ruby is concerned, the underlying `context` method will have changed. As long as the Java is hitting the `context` method to get the value, this will work.

> Many comparisons simply seem to return false always as soon as
Treeprocessors, Blockmacros or Block extensions come into play.

That's why I advocate so strongly for anyone using extensions to be using AsciidoctorJ 1.6.x. I don't think we want to spend a lot of time trying to fix 1.5.x when we know that the model is wrong. We should only address the high priority issues and keep encouraging people to switch.

robertpanzer:
The thing with the Java nodes is that they sit directly between Asciidoctor
Ruby and the original Ruby node, aka the "delegate", as soon as they are
touched once by Java.
These Java nodes have nothing in common with the Ruby classes except
methods that are magically duck typed against the called methods, so every
call of `context` that is on a Java node will return a string.

That's the major difference between Asciidoctorj 1.5 and 1.6: in 1.6 this
is no longer the case: Everything that is in Ruby stays in Ruby, the Ruby
world should never see Java implementations.





Am Sonntag, 14. August 2016 schrieb Dan Allen :

mojavelinux:
>  in 1.6 this is no longer the case: Everything that is in Ruby stays in Ruby, the Ruby world should never see Java implementations.

That means the patch I proposed may not work. If that's the case, then we'll have to use the solution in the PR.

robertpanzer:
Just tried adding the script adding some debug outputs without luck

As soon as a Treeprocessor touches a node, the Java version of AbstractNodeImpl.getContext() is called instead of the Ruby AbstractNode#context.
So I fear that as long as we don’t find another way to fix this, monkeypatching all calls of context to context.to_sym is the only way to make AsciidoctorJ behave the same way as Asciidoctor.

Just to make it explicit:
The additional global monkeypatching is *NOT* necessary to fix the exception Alexander reported in #497!
It was more that this additional example made me think again about how the current integration works, and that we have differences here.
To fix this error it was sufficient to implement the additional methods in the Java objects. (Which is also a sign that the original Ruby classes are completely out of the game after the conversion)
If we feel that we can live with these differences it would be sufficient to only add the missing Java methods and reduce the patching again.
We already live so long with these differences, maybe we can continue living with them until 1.6 when these problems should be fixed conceptually. (Waiting for errors to come:-) )

Cheers
Robert

mojavelinux:
> (Which is also a sign that the original Ruby classes are completely out of the game after the conversion)

That's the clarification I was looking for. Thank you.

> We already live so long with these differences, maybe we can continue living with them until 1.6 when these problems should be fixed conceptually. (Waiting for errors to come:-) )

That's the strategy I support. I think we need to be looking ahead to 1.6 instead of backwards at 1.5. That does put some more pressure on the timeline, but I think that's a good thing.

robertpanzer:
OK, then I’ll go back to monkeypatching only the comparison in abstract_node.rb that we currently have on master.

mojavelinux:
Cool. Technically, that's not monkey patching. That's hot patching. Monkey patching refers to loading a Ruby file that modifies a type already defined (such as replacing a method).

robertpanzer:
> Technically, that's not monkey patching. That's hot patching. Monkey patching refers to loading a Ruby file that modifies a type already defined (such as replacing a method).

Thanks for the clarification! I didn't know that difference yet.
Heard the term monkey patching in Java in the past when classes with the same name as a class in an imported library were defined to override behavior. 

If you're ok with the current state of this PR I would merge it.

mojavelinux:
:+1:

PR #503, opened 2016-08-19
Improve the toFile documentation about file vs. string output

### LINKED ISSUE #502: Output to HTML string omit header / footer
I'm using AsciidoctorJ 1.5.4.1 and recently switched from calling `convertFile` with

    OptionsBuilder.options()
      .backend(backend)
      .docType(DOCTYPE)
      .eruby(ERUBY)
      .safe(SafeMode.UNSAFE)
      .baseDir(basedir)
      .toFile(outFile);

to calling it with

    ...
      .toFile(false);

I.e. I simply switch to outputting to a `String` instead an output file. While investigating the HTML output String I noticed it lack the HTML header / footer, meaning there is no `<!DOCTYPE html>` and `<head></head>` tag in there, and no `<div id="footer-text">` etc. in the end. I don"t believe this is expected as I did not see that behavior documented anywhere.

sschuberth:
Closes #502.

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/Options.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/Options.java
@@ -49,6 +49,13 @@ public class Options {
         this.options.put(ATTRIBUTES, attributes);
     }
 
+    /**
+     * Toggle including header and footer into the output.
+     *
+     * @param headerFooter If <code>true</true>, include header and footer into the output,
+     *                     otherwise exclude them. This overrides any output-specific defaults.
+     *
+     */
     public void setHeaderFooter(boolean headerFooter) {
         this.options.put(HEADER_FOOTER, headerFooter);
     }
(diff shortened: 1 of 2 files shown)


asciidoctorj-core/src/main/java/org/asciidoctor/Options.java
```diff
@@ -67,10 +67,26 @@ public void setTemplateEngine(String templateEngine) {
...
+     *
+     */
     public void setToFile(String toFile) {
         this.options.put(TO_FILE, toFile);
     }
 
+    /**
+     * Toogle writing output to a file.
+     *
+     * @param toFile If <code>true</true>, overwrite the input file with the output,
```
robertpanzer: Would be great to have in here that this just changes the default of the headerFooter option and that it still can be changed.
Like this it reads that it's a fixed behavior creating an embeddable document with toFile=false and a full doc with toFile=true.

mojavelinux: It's also not true that it "overwrite(s) the input file with the output". It writes to a file adjacent to the input file which has the same root name and a file extension as specified by the converter.

sschuberth:
@mojavelinux Sorry for the delay, I've updated the PR to address your comments.

mojavelinux:
Thanks! Looks good to me!

robertpanzer:
Thank you!

Cherry-picked into asciidoctorj-1.6.0 as 07fed407fce84dc9c0f7cf203eb8266255ebad7b

PR #504, opened 2016-08-22
Fixes 501: Added method to get and remove substitutions

robertpanzer:
This PR adds methods to 
- get the active substitutors of a node `getSubstitutions()`
- to query if a substitutor is active `isSubstitutionEnabled()`
- and to remove a substitute `removeSubstitution()`

I could not find a method though to prepend or append a substitutor in the current Ruby API and directly adding the symbol to `@subs` did not work.
I guess to add a substitution at the moment it is necessary to pass the corresponding option to the node.

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/ast/StructuralNode.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/ast/StructuralNode.java
@@ -6,6 +6,42 @@ import java.util.Map;
 public interface StructuralNode extends ContentNode {
 
     /**
+     * Constant for special character replacement substitution like {@code <} to {@code &amp;lt;}.
+     * @see <a href="http://asciidoctor.org/docs/user-manual/#special-characters">Asciidoctor User Manual</a>
+     */
+    public static final String SUBSTITUTION_SPECIAL_CHARACTERS = "specialcharacters";
+
+    /**
+     * Constant for quote replacements like {@code *bold*} to <b>{@code bold}</b>.
+     * @see <a href="http://asciidoctor.org/docs/user-manual/#quotes">Asciidoctor User Manual</a>
+     */
+    public static final String SUBSTITUTION_QUOTES             = "quotes";
+
+    /**
+     * Constant for attribute replacements like {@code {foo}}.
+     * @see <a href="http://asciidoctor.org/docs/user-manual/#attributes-2">Asciidoctor User Manual</a>
+     */
+    public static final String SUBSTITUTION_ATTRIBUTES         = "attributes";
+
+    /**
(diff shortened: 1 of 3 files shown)


mojavelinux:
> I could not find a method though to prepend or append a substitutor in the current Ruby API 

Those methods do not currently exist.

>  directly adding the symbol to @subs did not work.

Hmm. That should work. That's how subs are modified in the Ruby code.

> I guess to add a substitution at the moment it is necessary to pass the corresponding option to the node.

That is the preferred way to set the subs on a block. Modifying the subs after the fact should be the exception.

robertpanzer:
I'll try again, I probably missed sth and I didn't fully follow the code.
It was — at least for me ;)— already pretty late.

Am Montag, 22. August 2016 schrieb Dan Allen :

robertpanzer:
No idea what I was doing yesterday. Retried this morning and it immediately worked like a charm.
That was absolutely not my day yesterday.

mojavelinux:
Excellent! I was amazed at how fast you got to this one. Moving forward at light speed!

robertpanzer:
Thanks! :-)
Atm I feel more like a snail though ;-)

I wonder if we should add constants for the substitutions (and where, i.e. on Processor or on StructuralNode – probably rather Processor)

mojavelinux:
Trust me, we're about to "turn a corner". There are a lot of things still up in the air, but they are going to land and give us a big boost forward. I'm starting to focus again on core so we can get 1.5.5 out and eventually 1.6.0. That will unblock a lot of things, I'm sure of it.

> I wonder if we should add constants for the substitutions 

I can definitely support that. Substitution are part of the AST, so they belong there. They aren't just an extension thing.

robertpanzer:
The snail thing was more a personal observation of myself, not related to Asciidoctor at all :-D

Will add constants to StructuralNode, makes absolute sense.

mojavelinux:
> The snail thing was more a personal observation of myself, not related to Asciidoctor at all :-D

I gotcha...though it's true that Asciidoctor does have a snail problem around this 1.6.0 transition. I need to face it head on so we can get beyond it.

> Will add constants to StructuralNode, makes absolute sense.

:+1:

robertpanzer:
Added constants for the substitution names to StructuralNode as I pulled them from substitutors.rb.
Left out `highlight` and `callouts` as they are also not documented in the asciidoctor user manual.

mojavelinux:
callouts should be documented. That's an oversight. That substitution only applies to verbatim blocks, but it is public. highlight is indeed an internal option (for now).

I wonder whether these constants should be moved to an interface named Substitution (or Substitutions). We should probably make it an annotation, even though it's not yet used as one.

robertpanzer:
Squashed and fixed my stupid codestyle error.

PR #509, opened 2016-09-06
Multiple small changes

kduske-n4:
Sorry to submit multiple changes in one pull request (and not using branches), but I had to make these changes in quick succession and could not isolate them. They are independent of each other, however. There are three changes:

**Convert nested Java maps to Ruby hashes in `RubyHashUtil#toRubyObject`**
This prevents a `ClassCastException` when, after creating a new block in a block processor, you attempt to access the attributes of that block later in a tree processor. When asking the block for its attributes, AsciidoctorJ assumes that what it gets from the asciidoctor AST is an instance of `RubyHash`, when in this case what it really gets is an instance of ` MapJavaProxy` that was injected by JRuby into the asciidoctor AST when the block was created originally.

I'm not sure about this change, since I had to choose one of several different methods to convert Java maps to Ruby hashes in `RubyHashUtil`.

**Return `null` if the inner document of a table cell is `nil`**
This prevents a `NullPointerException` when attempting to access the inner document of a table cell in a processor when that cell doesn't have an inner document in the asciidoctor AST.

**Add getter for `@content_model`**
This adds a getter for the `@content_model` attribute of an asciidoctor `AbstractBlock` to `StructuralNode`.

I hope these changes are useful for you and that you can apply them. I have run the tests in the asciidoctorj-asciidoctorj and asciidoctorj-distribution projects and there were no failures.

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/ast/impl/CellImpl.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/ast/impl/CellImpl.java
@@ -70,7 +70,10 @@ public class CellImpl extends ContentNodeImpl implements Cell {
 
     @Override
     public Document getInnerDocument() {
-        return (Document) NodeConverter.createASTNode(getRubyProperty("inner_document"));
+        IRubyObject innerDocument = getRubyProperty("inner_document");
+        if (innerDocument.isNil())
+        	return null;
+		return (Document) NodeConverter.createASTNode(innerDocument);
     }
 
     @Override
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/ast/impl/CellImpl.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/ast/impl/CellImpl.java
@@ -71,8 +71,9 @@ public class CellImpl extends ContentNodeImpl implements Cell {
     @Override
     public Document getInnerDocument() {
         IRubyObject innerDocument = getRubyProperty("inner_document");
-        if (innerDocument.isNil())
+        if (innerDocument.isNil()) {
         	return null;
+        }
 		return (Document) NodeConverter.createASTNode(innerDocument);
     }
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+bin/
 build/
 target/
 .classpath
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/ast/StructuralNode.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/ast/StructuralNode.java
@@ -80,6 +80,15 @@ public interface StructuralNode extends ContentNode {
     int getLevel();
 
     /**
+     * Returns the content model.
+     *
+     * @see ContentModel
+     *
+     * @return the content model
+     */
+    String getContentModel();
+
+    /**
      * Returns the source location of this block.
      * This information is only available if the {@code sourcemap} option is enabled when loading or rendering the document.
      * @return the source location of this block or {@code null} if the {@code sourcemap} option is not enabled when loading the document.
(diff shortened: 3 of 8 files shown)


asciidoctorj-core/src/main/java/org/asciidoctor/ast/impl/CellImpl.java
```diff
@@ -70,7 +70,10 @@ public void setVerticalAlignment(Table.VerticalAlignment valign) {
 
     @Override
     public Document getInnerDocument() {
-        return (Document) NodeConverter.createASTNode(getRubyProperty("inner_document"));
+        IRubyObject innerDocument = getRubyProperty("inner_document");
+        if (innerDocument.isNil())
+        	return null;
+		return (Document) NodeConverter.createASTNode(innerDocument);
```
robertpanzer: I haven't tested it, but I don't think this will compile.

robertpanzer:
Thanks a lot for your great contribution, Kristian!

Would you mind adding tests for the three problems you fixed?
That should be 
- a test that simply invokes getContentModel (Maybe like here: https://github.com/asciidoctor/asciidoctorj/blob/07fed407fce84dc9c0f7cf203eb8266255ebad7b/asciidoctorj-core/src/test/java/org/asciidoctor/WhenAsciiDocIsRenderedToDocument.java#L149)
- a test that uses two extensions, where one creates a node and another that tries to access the attributes? (Maybe copy this test and accommodate to the needs: https://github.com/asciidoctor/asciidoctorj/blob/07fed407fce84dc9c0f7cf203eb8266255ebad7b/asciidoctorj-core/src/test/groovy/org/asciidoctor/extension/WhenABlockMacroProcessorCreatesATable.groovy#L51)
- A test that tries to access the inner document of a table cell when it doesn't have the Asciidoctor style. That should fit in here: https://github.com/asciidoctor/asciidoctorj/blob/asciidoctorj-1.6.0/asciidoctorj-core/src/test/groovy/org/asciidoctor/WhenATableIsLoaded.groovy

That would help us a lot and would avoid that we later stumble over these problems again.

asciidoctorj-core/src/main/java/org/asciidoctor/ast/impl/CellImpl.java
kduske-n4: It does, though. Why shouldn't it?

kduske-n4:
Added test cases as requested.

asciidoctorj-core/src/main/java/org/asciidoctor/ast/impl/CellImpl.java
robertpanzer: Gotcha! :-)
Could you please add braces around the `return null`.

I probably automatically put them in there mentally around both returns that have the same indentation.

kduske-n4: Done.

robertpanzer:
Thanks a lot! Such tests really help us.
Executed them locally and they look good, if I revert the main code, I can also see the errors that you fixed.

There are a couple of code style problems that codenarc complains.
Would you mind fixing them?
While it might seem pedantic it helps us to keep a common code style across the groovy code.
When doing a `./gradlew clean build` codenarc will create a report that shows you what it doesn't like.

Thanks a lot!

Cheers
Robert

kduske-n4:
Sure, will do!

kduske-n4:
Done ;-)

kduske-n4:
Some things still seem broken in your CI. I have run the tests locally and it was fine. Any hints?

robertpanzer:
Yes, there are some problems with Asciidoctor PDF and JRuby on Windows.
That's why this test is in. :-)
Will upgrade the JRuby version soon and look if it works better now (even though I don't have much hope)
Unix build is green, so I'm fine to merge.

Thanks!

Issue #513, opened 2016-09-15
Copying attribute_entries when creating a new block


kduske-n4:
I'm trying to create a new block in a tree processor. This block should retain the `attribute_entries` of another block, but when I pass a copy of the original block's attributes to `Processor#createBlock`, they are not stored. This leads to document variables being replaced with incorrect values for the newly created block.

The first problem appears to be in `RubyAttributesMapDecorator#createJavaMap` which doesn't handle keys that are Ruby symbols. Adding

     } else if (rubyKey instanceof RubySymbol) {
          key = ((RubySymbol) rubyKey).asJavaString();

Puts the `attribute_entries` into the Java map. But then the second problem kicks in, which is that this key must be converted back to a Ruby symbol when the attributes are stored in the options. In method `Processor#createBlock(StructuralNode, String, List<String>, Map<String, Object>, Map<Object, Object>)`, the attributes get put into the options:

    options.put(Options.SOURCE, content);
    options.put(Options.ATTRIBUTES, new HashMap<String, Object>(attributes));

They are then passed on to `Processor#createBlock(StructuralNode, String, Map<Object, Object>)`, where the options are converted to a Ruby hash:

    RubyHash convertMapToRubyHashWithSymbols = RubyHashUtil.convertMapToRubyHashWithSymbolsIfNecessary(rubyRuntime, options);

But the key `attribute_entries` in the nested attributes is not treated specially, so it remains a string rather than a symbol. If I convert the attributes map to a Ruby hash myself before they are put into the options, it works, but then other keys get lost (e.g. `role`) since those are converted to symbols when they aren't expected to be.

So, long story short, how to I store things which are expected to be symbols in the attributes map and have them show up correctly in the Ruby AST?

robertpanzer:
@kduske-n4 Thank you for such a detailed issue report, this is really awesome.
I tried to reproduce the error you mentioned but I couldn't find it.
I added a test case in #514, maybe you can take a look at that and see if I missed sth in in your description.

What I realized though when creating the test was, that there is no way yet to set the style of a node (Or I forgot it due to vacation mode in my brain). Therefore I added that setter.

kduske-n4:
Hi @robertpanzer, thank you for your reply. I will create a pull request with a test that demonstrates the issue tomorrow.

mojavelinux:
[...]

PR #517, opened 2016-10-01
Add Cell.getRawText() and ListItem.getRawText() to get unsubstituted …

robertpanzer:
…text of table cells and list items.

This PR goes for the first point in https://github.com/asciidoctor/asciidoctorj/issues/513#issuecomment-247933943 :
Add accessors to get the raw, unsubstituted text from a Table Cell or a ListItem.
The methods are each called `getRawText()` and they get the field `@text` of the associated nodes instead of invoking the accessor method `text` on the respective Ruby nodes, as these apply the substitutions.

I am not really happy with the naming of the methods yet, as we have `String getContent()` and `List<String> getLines()` on Blocks, but `String getText()` and `String getRawText()` on Cells and ListItems.

But the thing that we're even able now to directly access members of the Ruby objects is already a great thing! This approach wouldn't work on 1.5.x without reimplementing all the functionality in the Java nodes in the same way as it works in the Ruby nodes.

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/ast/Cell.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/ast/Cell.java
@@ -16,7 +16,7 @@ public interface Cell extends ContentNode {
     /**
      * @return The text of the cell without substitutions being applied.
      */
-    String getRawText();
+    String getSource();
 
     Object getContent();
 
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/ast/Cell.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/ast/Cell.java
@@ -8,7 +8,15 @@ public interface Cell extends ContentNode {
 
     int getRowspan();
 
+    /**
+     * @return The text of the cell including substitutions being applied.
+     */
     String getText();
+
+    /**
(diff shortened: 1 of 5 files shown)


mojavelinux:
I agree we need to think about the naming. That's one of the reasons I haven't added these methods in core yet. I'd prefer either `getSourceText` or just `getSource`. I don't like the term "raw".

robertpanzer:
Absolutely right. 
I renamed `getRawText()` to `getSource()`.
Then the methods also align with their pendants in `org.asciidoctor.ast.Block`.

Wonder if it would make sense to put an interface on it containing `getContent()` and `getSource()`?

mojavelinux:
Those methods probably belong on ContentNode (though we may still have some children that don't yet implement them).

robertpanzer:
Will merge it now then.

Regarding where to put those methods: Don't these methods only make sense for nodes that are leafs in the AST? Sections or List don't have an own source or content, do they? 
This only under the premise that it isn't defined as the source or content collected from the contained subtree.
This would also create problems with the corresponding setSource() methods.

mojavelinux:
That's a good point. They do only apply for leafs at the moment. Though it's reasonable to assume, I suppose, that all nodes should be able to access their source. It's a bit of a fuzzy area in the AST right now.

PR #518, opened 2016-10-03
Isolate the JRuby instance of Asciidoctor from a system gem path by d…

robertpanzer:
…efault.

This PR changes the behavior of `Asciidoctor.Factory.create()` to isolate from a system gem path.


mojavelinux:
That calls for a :taco:!

:tada:

PR #523, opened 2016-10-09
Added methods setSource to Block, Cell and ListItem

robertpanzer:
This PR tackles points 2 & 3 of #513 

It adds the method `setSource(String)` to Block, ListItem and Cell.
There's also a new method `Block.setLines(List<String>)` to set the source lines.


jiangrij:
Hi robertpanzer,
I tried to use setSource(String) to a Cell, but it does not work for generating HTML(the cell source was not updated). Is there anything else I missed? Could you please advise?

The following is the code which extends from Treeprocessor:
----------------- code part ---------------------
public Document process(Document document) {
List<StructuralNode> blockslist1 = document.getBlocks();
.......
List<StructuralNode> blockslist = section.getBlocks();
for (StructuralNode abstractBlock : blockslist) {
    if ((abstractBlock instanceof Table)) {
	 Table table = (Table) abstractBlock;
	  for (Row row : table.getBody()) {
		List<Cell> cells = row.getCells();
		for (int cellindex = 0; cellindex < cells.size(); cellindex++) {
			Cell cell = cells.get(cellindex);
			cell.setSource("**New Source**");
--------------------------------------

Thanks.

robertpanzer:
It looks like we're missing a method `setText()` at the moment.
For now you will have to create a new Cell with `Processor.createTableCell(column, "New Source")` and replace it in the `cells`.

robertpanzer:
I just tried an example like this and it worked for me.
Are you sure that the your conditions found the table?

robertpanzer:
@jiangrij I just added a test in #638 that shows that updating the source of a cell works.

PR #525, opened 2016-10-21
Upgrade distribution to use JRuby 9.1.5.0 instead of 1.7.xx

robertpanzer:
If we want to deliver the distribution with the latest and greatest asciidoctor(j)-pdf soon, we will have to deliver it with JRuby 9xxx.

This PR simply pulls up the version for the asciidoctorj-distribution to 9.1.5.0, so that it contains jruby-complete-9.1.5.0.jar instead of jruby-complete-1.7.25.jar.


robertpanzer:
I wonder if I should extend this PR to also update asciidoctorj-pdf to the latest version as it would use mixed versions of JRuby in this project anyway.

PR #537, opened 2016-12-07
Fixes and Extensions to the IncludeProcessor and Reader classes

mmews-n4:
This PR includes the following three changes:

### Fix of IncludeProcessor#handles method
The handles(String) method in the IncludeProcessor is not called, because the annotation on the IncludeProcessorProxy#handles method lacks a question mark. Test case added.

### Extended IncludeProcessor to accept positional attributes
Before, the IncludeProcessor ignored any positional attributes. Now they can be used separated by commas. Internally, they are key value pairs where the keys are numbered accordingly to their order in the squared macros. The fix is only that these key value pairs are also evaluated when constructing the attributes map. Test case added.

### Extended Reader and ReaderImpl
The reader now provides the methods getFile and getDir which are delegated to 'file' and 'dir' of asciidoctor.

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/extension/Reader.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/extension/Reader.java
@@ -6,7 +6,7 @@ public interface Reader {
 
     /**
      * Get the 1-based offset of the current line.
-     * 
+     *
      * @return 1-based offset.
      * @deprecated Please use {@link #getLineNumber()}
      */
@@ -20,40 +20,40 @@ public interface Reader {
     int getLineNumber();
 
     /**
-	 * Get the name of the file of the current line.
-	 *
-	 * @return file name
-	 */
-	public String getFile();
+     * Get the name of the file of the current line.
+     *
+     * @return file name
+     */
+    public String getFile();
 
-	/**
-	 * Get the name of the directory of the current file
-	 *
-	 * @return directory name
-	 */
-	public String getDir();
+    /**
+     * Get the name of the directory of the current file
+     *
+     * @return directory name
+     */
+    public String getDir();
 
     /**
      * Check whether there are any lines left to read. If a previous call to this method resulted in a value of false,
      * immediately returned the cached value. Otherwise, delegate to peek_line to determine if there is a next line
      * available.
-     * 
+     *
      * @return True if there are more lines, False if there are not.
      */
     boolean hasMoreLines();
 
     /**
      * Peek at the next line and check if it's empty (i.e., whitespace only)
-     * 
+     *
      * This method Does not consume the line from the stack.
-     * 
+     *
      * @return True if the there are no more lines or if the next line is empty
      */
     boolean isNextLineEmpty();
 
     /**
      * Get the remaining lines of source data joined as a String. Delegates to Reader#read_lines, then joins the result.
-     * 
+     *
      * @return the lines read joined as a String
      */
     String read();
@@ -62,7 +62,7 @@ public interface Reader {
      * Get the remaining lines of source data. This method calls Reader#read_line repeatedly until all lines are
      * consumed and returns the lines as a String Array. This method differs from Reader#lines in that it processes each
      * line in turn, hence triggering any preprocessors implemented in sub-classes.
-     * 
+     *
      * @return the lines read as a String Array
      */
     List<String> readLines();
@@ -112,13 +112,13 @@ public interface Reader {
 
     /**
      * Advance to the next line by discarding the line at the front of the stack
-     * 
+     *
      * @return a Boolean indicating whether there was a line to discard.
      */
     boolean advance();
 
     /**
-     * 
+     *
      * Public: Advance to the end of the reader, consuming all remaining lines
      */
     void terminate();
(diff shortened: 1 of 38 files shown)


robertpanzer:
Awesome, thank you!
Great that you discovered that the handles() method wasn't actually called for IncludeProcessors.

I am personally ok with mapping the positional args to the strings "1", "2", ...

Could you please check why the tests are failing on Travis with Java 7?

Another – maybe pedantic – note: Could you please check the indentation of the source code? Looks like you're using Tabs instead of Spaces which makes the code look a bit weird on Github.

mojavelinux:
> I am personally ok with mapping the positional args to the strings "1", "2", ...

I believe that will solve some other type issues we were having. I was using numbers here to avoid collision with attributes that have the same name, but we could just say that numeric keys (1=value) are reserved for another purpose and should not be declared explicitly.

> Extended IncludeProcessor to accept anonymous attributes

Do you mean positional attributes (value only)? If so, could you update the terminology? (See http://asciidoctor.org/docs/user-manual/#setting-attributes-on-an-element)

If I try using an IncludeProcessor written in Ruby, I get these attributes:

```
include::other.adoc[foo,bar,a=b]
```

gives me:

```
{1=>"foo", 2=>"bar", "a"=>"b"}
```

So effectively this fix translates those numeric keys into string keys, correct?

mojavelinux:
Btw, thanks for this fix!

mojavelinux:
:+1:

mmews-n4:
Thanks for your feedback!
- I replaced the tabs by spaces, as you preferred, @robertpanzer.
- @mojavelinux, you're right, this fix translates those numeric keys into string keys.
- I also replaced the term 'anonymous' by 'positional' to fit in to the terminology.
- I tried to reproduce the errors in the Java7 environment, but it seems to work with JDK7 on my machine. Do you have some advice on how I could reproduce the failure?

robertpanzer:
Sorry for responding so late!

I just tried out the failing test.
I guess this is due to the extension iterating over the attributes *Map* and concatenating the results.
The map is not ordered in any way and the order of the elements given by a HashMap definitely differs between Java 7 and 8. (Stumbled myself over this change a couple of times)
So for me the test fails with this error:
```
java.lang.AssertionError: 
Expected: a string starting with "My,Positional,Attribute List"
     but: was "Attribute List,Positional,My"
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.junit.Assert.assertThat(Assert.java:956)
	at org.junit.Assert.assertThat(Assert.java:923)
	at org.asciidoctor.extension.WhenJavaExtensionIsRegistered.a_include_processor_can_handle_positional_attrs(WhenJavaExtensionIsRegistered.java:430)
```
You might simply want to either sort the entries by key ("1", "2" and "3") or just test if the expected substrings are present.

asciidoctorj-core/src/main/java/org/asciidoctor/extension/Reader.java
```diff
@@ -20,6 +20,20 @@
     int getLineNumber();
 
     /**
+	 * Get the name of the file of the current line.
+	 *
+	 * @return file name
+	 */
+	public String getFile();
```
robertpanzer: Also, could you please change these tabs to spaces?
(Sorry, but I am pretty opinionated on this 😁

asciidoctorj-core/src/main/java/org/asciidoctor/extension/ReaderImpl.java
```diff
@@ -31,10 +31,20 @@ public int getLineno() {
...
     @Override
     public int getLineNumber() {
-        return getInt("lineno");
+		return getInt("lineno");
     }
 
-    @Override
+	public String getFile() {
+		IRubyObject rObj = getRubyProperty("file");
+		return rObj.toString();
```
robertpanzer: Same here.

asciidoctorj-core/src/test/java/org/asciidoctor/extension/WhenJavaExtensionIsRegistered.java
```diff
@@ -395,6 +396,42 @@ public void a_include_instance_processor_should_be_executed_when_include_macro_i
     }
 
     @Test
+	public void a_include_processor_should_only_handle_its_handles() {
```
robertpanzer: Here as well, tabs.

mmews-n4:
@robertpanzer thanks a lot for your advice! Hopefully the tests will be green now!

mmews-n4:
@robertpanzer, I don't know the cause of the error. The message says, I should look at 
`file:///C:/projects/asciidoctorj/asciidoctorj-pdf/build/reports/tests/index.html`
to get information about the error. Do you know the cause of the error?

robertpanzer:
Awesome!
Thank you for this great contribution!

abelsromero:
> Do you know the cause of the error?

@mmews-n4 The full report is not accessible now, but the output shows this `SEVERE: (RuntimeError) asciidoctor: FAILED: required gem 'rouge' is not installed. Processing aborted.`

Highly likely, to be related to this: https://github.com/robertpanzer/asciidoctorj-pdf/issues/3

Issue #526, opened 2016-10-23
Align test support modules with build tool conventions on the 1.6.0 branch


robertpanzer:
At the moment the test support modules asciidoctorj-test-support and asciidoctorj-arquillian-extension contain their sources under src/test even though these sources are exported to other modules.

While the Gradle build is configured correctly to reflect this IntelliJ 16 doesn't seem to play nicely with that and the project doesn't build any longer for me, while it was working with IntelliJ 14.

Goal of this issue to either produce a description how to get the project usable with IntelliJ 16 or to move the sources to src/main which makes more sense imo.

ancho:
Hey Robert, is this still a problem for you? I tried to build the asciidoctorj-1.6.0 branch with intellij 2016.3 and it is working as expected for me.
But I try to update the gradle build to the latest version 3.2.1 and update the plugins. Maybe that increase the build time.

robertpanzer:
Hi Frank,

I just retried and it still doesn't work.
I updated IntelliJ to 2016.03.1, cleaned the repo directory and newly imported the project.
Still classes are ClasspathResources or Unshared are not found.
Also the project structure indicates that the test sources of the testutil and arquillian-extension modules are not imported into the test dependencies of the asciidoctorj-core module.

Actually this behavior really sucks as I have already lost several hours trying to get the stuff working in IntelliJ after switching between master and asciidoctorj-1.6.0 (I do this a lot)

(Actually I still don't understand the reasoning for moving these sources to `src/test`. Imo for the same reason the sources of JUnit or Arquillian should reside in `src/test`)

<img width="1132" alt="bildschirmfoto 2016-12-18 um 19 30 42" src="https://cloud.githubusercontent.com/assets/1163662/21295592/c219b1da-c558-11e6-91f3-578620be4a64.png">

ancho:
Ahh. Now I get it. Unfortunately intellij didn't show me any dependency problems after switching the branch. I had to open a test in the core module to see them.
I'm with your suggestion of moving the classes to src/main. pull-request follows.

PR #539, opened 2016-12-19
fix dependency problems with intellij

ancho:
Moved asciidoctorj-test-support and asciidoctorj-arquillian-extension test classes to src/main as discussed in #526 

I replaced provided configurations with compileOnly a while ago. It is not part of the problem but I think it's still a good idea.


robertpanzer:
Thank you!
Finally able to work again! :-)

PR #542, opened 2017-01-20
Bump asciidoctor-diagram to 1.5.4

robertpanzer:


robertpanzer:
Build with Java6 failed because the TLS certificate could not be validated when downloading gradle.
I think Java6 is really dead now.

PR #545, opened 2017-03-05
Bump asciidoctor-pdf to 1.5.0-alpha.14

robertpanzer:
This PR upgrades asciidoctor-pdf to 1.5.0-alpha.14.
As this requires Ruby 2 the following changes are made:

- asciidoctorj still depends on jruby-complete 1.7.26 
- asciidoctorj-pdf depends on jruby-complete 9.1.7.0
- asciidoctorj-distirbution depends on jruby-complete 9.1.7.0

Therefore a lib that directly requires asciidoctorj will still be able to run with Java 6.
As soon as the new asciidoctorj-pdf or a new distribution would be used JRuby 9.1.7.0 and therefore Java 7 would be required.

This is an alternative take to get a current asciidoctorj-pdf out without using an external repository for it.


robertpanzer:
Travis build failed on Java 6 which is clear. 

I'd love to get some feedback if this would be an acceptable approach.

robertpanzer:
Upgraded the dependency on JRuby for asciidoctorj-pdf and the distribution to the latest 9.1.8.0.

@mojavelinux, @abelsromero, @lordofthejars I would be happy to get feedback if this is sth we would like to merge and release. (And I would like to push out a release, we're already far behind)

abelsromero:
> And I would like to push out a release, we're already far behind

I feel the same way.

I'd go for the option that pushes the release and does not block us. If it means having asciidoctorj-pdf require a different version of Java and JRuby, so be it.
Also I think that maintaining Java 6 compatibility today doesn't make the same sense as when we decided to keep it up to v 1.6.0,

robertpanzer:
If there are no objections until next Wednesday (3/15) I'd go for a release with this on Thursday.

dcoraboeuf:
Will it allow a fix for asciidoctor/asciidoctor-pdf#6 ?

abelsromero:
> Will it allow a fix for asciidoctor/asciidoctor-pdf#6 ?

Yes. This release will contain asciidoctor-pdf 1.5.0-alpha.14 as seen in the last change.

mojavelinux:
I'm in favor of getting the release out however you can. At the same time,
I'm going to see that the new repository is set up in the Asciidoctor
organization. I'd like to release from that repository, but I don't want to
hold up this release. We'll see where we stand after this weekend to make
that call one way or the other.

abelsromero:
Now that the release is life, during this week I will test it in the maven plugin and proceed to release the 1.5.5 version too this week.
I will skip the 1.5.4 version and go directly to 1.5.5 if no one has any comment on it.

mojavelinux:
:+1: 

Nice work!

mojavelinux:
I'm currently working on Asciidoctor 1.5.6, so that's going to happen soon, FYI.

I'm also going to be releasing another version of Asciidoctor PDF to address compatibility with Prawn 2.2.0.  When we upgrade AsciidoctorJ PDF to Prawn 2.2.0, let's make sure we align all the gem dependency versions at that time. The current release of AsciidoctorJ PDF wasn't affected since it locks the Prawn version at 2.1.0.

abelsromero:
@mojavelinux as discussed time ago I have run some test on my own repo, and finished the release manual https://github.com/abelsromero/asciidoctor-maven-plugin/wiki/Maven-release-steps. Oddly I had to set SSH authentication to push tags, I did not do that last time.

Tomorrow I'll prepare the release notes and release 🎉

abelsromero:
Not possible to release today.
After upgrading jRuby to 1.7.26, and testing on Windows 7 I found that this test https://github.com/asciidoctor/asciidoctor-maven-plugin/blob/master/src/test/groovy/org/asciidoctor/maven/test/AsciidoctorMojoTest.groovy#L920 fails.
This test in fact validated that remote github includes failed, so it's good news. But now under some circumstances they still fail.
Running from shell, the include works, but from IntelliJ it fails saying `Unresolved directive in github-include.adoc`. I am not very worried because I have branch in my repo and everything worked fine in Travis (for Java 6, 7 and 8), but I have to run some tests on my Mac to see if it's only Windows to document it.

robertpanzer:
@abelsromero I can also run that test tomorrow morning, I have a Windows box here.

robertpanzer:
@abelsromero I ran the AsciidoctorMojoTest and indeed after updating jruby to 1.7.26 and AsciidoctorJ to 1.5.5 the test fails.
I am just wondering if the test is correct. It contains an include of a pom.xml on Github and set `allow-uri-read=true`.
Then it checks that the generated document does *not* contain the included pom.xml.
Isn't it just working right? Why should Asciidoctor not include the file? Maybe I am missing sth.

abelsromero:
Exactly, it was done to keep track of an issue, I recall it had something to do with specific GitHub urls.
If it's fixed, we should change the assert or remove the test. I already tested the change yesterday in a branch and it's working fine https://travis-ci.org/abelsromero/asciidoctor-maven-plugin/builds/211284707.
Maybe I am just over-reacting and the error has something to do with my IntelliJ? I run it on my work machine, I still have to run in on my Windows PC.

robertpanzer:
Ah ok, so you're saying that the current code in this commit does not run in your IntelliJ on your Windows box?
Then I can say that it runs on my Windows box in my IntelliJ :-)

abelsromero:
> Ah ok, so you're saying that the current code in this commit does not run in your IntelliJ on your Windows box?

Exactly!

> Then I can say that it runs on my Windows box in my IntelliJ :-)

Thanks, in that case I can breath :+1: I'll release on Friday, today is beer Thurday :beers: and I' don't think I should release when I get home.

robertpanzer:
Very good idea, 🍻!  I'll do that too today!

PR #552, opened 2017-05-07
Fix upstream build and remove submodule for asciidoctorj-pdf

robertpanzer:
Similar to #550 this PR fixes the build against upstream.
(Just tested that the upstream build also fails on current master)

Also this PR removes the submodule for asciidoctorj-pdf and imports it as an external dependency as we moved it into its own repo at https://github.com/asciidoctor/asciidoctorj-pdf.
While it looks like we're removing a test for asciidoctorj-pdf against *this* version of AsciidoctorJ I want to mention that we still have an integration test with asciidoctorj, asciidoctorj-pdf and asciidoctorj-diagram that uses the jars instead of the exploded archives and is therefore much more realistic anyways.


asciidoctorj-core/src/main/resources/org/asciidoctor/internal/asciidoctorclass.rb
```diff
@@ -2,6 +2,9 @@ module AsciidoctorJ
     include_package 'org.asciidoctor'
     module Extensions
         include_package 'org.asciidoctor.extension'
+        # Treeprocessor was renamed in to TreeProcessor in https://github.com/asciidoctor/asciidoctor/commit/f1dd816ade9db457b899581841e4cf7b788aa26d
+        # This is necessary to run against both Asciidoctor 1.5.5 and 1.5.6
+        TreeProcessor = Treeprocessor
```
mojavelinux: I really want to know why the alias I have in core isn't working here.

robertpanzer: The tests fail with this error: (see https://travis-ci.org/asciidoctor/asciidoctorj/jobs/229627476)
```
org.asciidoctor.extension.WhenJavaExtensionIsRegistered > a_treeprocessor_instance_should_be_executed_in_document FAILED
    org.asciidoctor.internal.AsciidoctorCoreException: org.jruby.exceptions.RaiseException: (NameError) asciidoctor: FAILED: /home/travis/build/asciidoctor/asciidoctorj/asciidoctorj-core/build/resources/test/sample-with-terminal-command.ad: Failed to load AsciiDoc document - TreeProcessor not found in packages org.asciidoctor.extension; last error: cannot load Java class org.asciidoctor.extension.TreeProcessor
```

I don't fully understand what happens here in https://github.com/asciidoctor/asciidoctor/blob/48120dc8d56028268515882e84ac6ca165f76e7e/lib/asciidoctor/extensions.rb#L1197:
```ruby
kind_java_class = (defined? ::AsciidoctorJ) ? (::AsciidoctorJ::Extensions.const_get kind_class_symbol) : nil
```
but it seems to expect a Java class under `::AsciidoctorJ::Extensions::TreeProcessor`.
And this is sth we don't have at the moment without this assignment.
(Or is it necessary to distinguish if the extension was registered as a class or an Object?)

I don't know the reasons for a special handling of AsciidoctorJ extensions and I've actually not known before today that there is one, but I guess we could drop that as soon as we're 1.6.x-only, as these Extensions should be proper Asciidoctor extensions that aren't distinguishable from "normal" Asciidoctor extensions written in Ruby.

mojavelinux: Aha! Thanks for pointing that out. You are absolutely correct. We do need the alias there. That's a different namespace (and thus different alias) from the one in core.

You are correct that we may not need this special case for handling Java types in the future. But we'll have to test it. After all, Ruby is using the Java native type directly in this case.

PR #553, opened 2017-05-07
Fixes assertions for some test that failed on  Windows 7 and Windows 8.1

abelsromero:
Fixes the first and third issues mentioned in https://github.com/asciidoctor/asciidoctorj/pull/550#issuecomment-299573561.

Fixes only affect test assertions so they work both for *nix and Windows.

### DIFF
--- a/asciidoctorj-core/src/test/java/org/asciidoctor/WhenAsciiDocIsRenderedToDocument.java
+++ b/asciidoctorj-core/src/test/java/org/asciidoctor/WhenAsciiDocIsRenderedToDocument.java
@@ -235,7 +235,7 @@ public class WhenAsciiDocIsRenderedToDocument {
         Document document = asciidoctor.load(DOCUMENT, options);
         File inputFile = classpath.getResource("rendersample.asciidoc");
         String content = document.readAsset(inputFile.getAbsolutePath(), new HashMap<Object, Object>());
-        assertThat(content, is(IOUtils.readFull(new FileReader(inputFile))));
+        assertThat(content, is(IOUtils.readFull(new FileReader(inputFile)).replaceAll("\r","\\")));
     }
 
     @Test
--- a/asciidoctorj-core/src/test/java/org/asciidoctor/WhenAsciiDocIsRenderedToDocument.java
+++ b/asciidoctorj-core/src/test/java/org/asciidoctor/WhenAsciiDocIsRenderedToDocument.java
@@ -235,7 +235,7 @@ public class WhenAsciiDocIsRenderedToDocument {
         Document document = asciidoctor.load(DOCUMENT, options);
         File inputFile = classpath.getResource("rendersample.asciidoc");
         String content = document.readAsset(inputFile.getAbsolutePath(), new HashMap<Object, Object>());
-        assertThat(content, is(IOUtils.readFull(new FileReader(inputFile)).replaceAll("\r","\\")));
+        assertThat(content, is(IOUtils.readFull(new FileReader(inputFile))));
     }
 
     @Test
--- a/asciidoctorj-distribution/src/test/groovy/org/asciidoctor/diagram/WhenDitaaDiagramIsRendered.groovy
+++ b/asciidoctorj-distribution/src/test/groovy/org/asciidoctor/diagram/WhenDitaaDiagramIsRendered.groovy
@@ -1,6 +1,5 @@
 package org.asciidoctor.diagram
 
-import jnr.ffi.Platform
 import org.asciidoctor.Asciidoctor
 import org.asciidoctor.AttributesBuilder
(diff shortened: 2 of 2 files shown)


asciidoctorj-core/src/test/java/org/asciidoctor/WhenAsciiDocIsRenderedToDocument.java
```diff
@@ -235,7 +235,7 @@ public void should_be_able_to_read_asset() throws FileNotFoundException {
         Document document = asciidoctor.load(DOCUMENT, options);
         File inputFile = classpath.getResource("rendersample.asciidoc");
         String content = document.readAsset(inputFile.getAbsolutePath(), new HashMap<Object, Object>());
-        assertThat(content, is(IOUtils.readFull(new FileReader(inputFile))));
+        assertThat(content, is(IOUtils.readFull(new FileReader(inputFile)).replaceAll("\r","\\")));
```
mojavelinux: Actually, we really want to avoid doing this type of hack, even in tests. I looked at the implementation of IOUtils.readFull and it looks very strange to me. Java should be reading the file the same on all systems, but it's not.

asciidoctorj-distribution/src/test/groovy/org/asciidoctor/diagram/WhenDitaaDiagramIsRendered.groovy
```diff
@@ -58,7 +59,10 @@ Hello World
 
 
         then:
-        result.contains("""src="${imagesOutDir.absolutePath}/${imageFileName}.png""")
+        if (Platform.getNativePlatform().OS == Platform.OS.WINDOWS)
+            result.contains("""src="${imagesOutDir.absolutePath.replaceAll("\\\\", "//")}/${imageFileName}.png""")
+        else
+            result.contains("""src="${imagesOutDir.absolutePath}/${imageFileName}.png""")
```
mojavelinux: I don't think we should be checking for absolute paths in the output. We would never want the HTML to point to an absolute path. But this is caused by the fact that we're passing an absolute path to imagesdir. I think imagesdir needs to be set to a relative path and then we don't have to worry about replacing any system path separators.

abelsromero: At some point the usage of 'imagesdir 'was added. We could completely remove it or just set any value for the sake of ensuring imagesdir works as espected.

mojavelinux: :+1:

asciidoctorj-core/src/test/java/org/asciidoctor/WhenAsciiDocIsRenderedToDocument.java
abelsromero: Upgraded to last version of JDK (131) and error banished :man_shrugging: I'll rollback the change.

asciidoctorj-distribution/src/test/groovy/org/asciidoctor/diagram/WhenDitaaDiagramIsRendered.groovy
```diff
@@ -27,8 +27,7 @@ class WhenDitaaDiagramIsRendered extends Specification {
 
         given:
         String imageFileName = UUID.randomUUID()
-        File imagesOutDir = testFolder.newFolder()
-        File createdImage = new File(imagesOutDir, "${imageFileName}.png")
+        File imagesOutDir = new File(testFolder.root, "images-dir")
```
robertpanzer: Without looking at the codenarc report I guess the build failed because this could be a string with single quotes instead of double quotes.
Also I could imagine that codenarc prefers `.name` over `.getName()` but I am not sure about that.

abelsromero: I was checking right now. Really this is my worst PR in years XD but I take it with humor.

robertpanzer: I know that feeling, also happens to me very often at the moment :-D

abelsromero: It complains about the GString only, but I will change the `getName` anyway

robertpanzer:
Appveyor also made it to our known rouge error.
For some reason the Appveyor integration is missing in the project settings.
Nevertheless it still seems to get the PR notifications and build them, strange.

PR #559, opened 2017-06-30
Fix upstream build

robertpanzer:
The current upstream version of asciidoctor core (1.5.6.dev) has some changes that break our tests.
(Requiring an alt text for images and different section names)

This PR fixes the tests so that they run against asciidoctor 1.5.5 and the current 1.5.6.dev


mojavelinux:
:+1:

PR #560, opened 2017-06-30
Fix tests against upstream

robertpanzer:
Fixes tests against the current asciidoctor 1.5.6.dev


mojavelinux:
We're now storing proper names for section types so we can do things like formal cross reference text.

PR #561, opened 2017-07-01
First proposal to unregister an extension.

robertpanzer:
This is a first proposal for unregistering dedicated extensions.
The registration methods now return an `ExtensionRegistration` object that can be used to unregister the extension again with `Asciidoctor.unregisterExtension(ExtensionRegistration)`.
By changing the signature like this we have to give up on chaining calls to the registration methods, but I think that this is not really an issue, we also didn't have that yet in 1.5.x – it was a 1.6.x-only feature.

The group name is the symbol of a random UUID, I assume that this will not clash.

The tests are likely to fail against the current asciidoctor 1.5.5, but it should succeed against upstream.

What do you think about it?
If you are principally ok with it I would add more tests, there is only one test atm.

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/Asciidoctor.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/Asciidoctor.java
@@ -12,6 +12,7 @@ import org.asciidoctor.ast.Document;
 import org.asciidoctor.ast.DocumentHeader;
 import org.asciidoctor.ast.StructuredDocument;
 import org.asciidoctor.converter.JavaConverterRegistry;
+import org.asciidoctor.extension.ExtensionRegistration;
 import org.asciidoctor.extension.JavaExtensionRegistry;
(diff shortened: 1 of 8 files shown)


mojavelinux:
This is a good start.

> By changing the signature like this we have to give up on chaining calls to the registration methods, but I think that this is not really an issue

That seems reasonable to me. I suppose if we wanted to still allow chaining, we'd just have to get creative with method overloading. Perhaps what we could do is chain methods that accept an explicit group name. If you know the group name up front, you can pass that in to be unregistered.

> The group name is the symbol of a random UUID, I assume that this will not clash.

It should definitely be an option to pass a custom group name. And, as I mentioned above, it could solve our chaining / no chaining problem.

Now that 1.5.6 is out, let's proceed. We definitely want to propagate this feature up to AsciidoctorJ 1.5.6 since that was the motivation for the chain to core.

robertpanzer:
No longer relevant. 
We found another approach.

PR #564, opened 2017-07-19
Fixes #359. Explicit unregistration of extensions.

### LINKED ISSUE #359: Please add a way to unregister a specific extension (make extension API symmetric)
When registering a extension class, e.g. via the `JavaExtensionRegistry`, there is no way to unregister the same extension besides unregistering all. In `ExtensionRegistry` there is no unregister support at all.

To implement extension support in OSGi via the Whiteboard Pattern (see #297) a way to unregister an extension is needed.

robertpanzer:
A second approach to unregistering extensions.
Now it is completely explicit.
Instead of getting a registration handle from the registration, a registration name has to be passed to the registration methods explicitly.

Basically all registration methods are overloaded with an additional version that takes a string as the registration name.

Note that this PR is for master.

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/Asciidoctor.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/Asciidoctor.java
@@ -638,7 +638,13 @@ public interface Asciidoctor {
      * Unregister all registered extensions.
      */
     void unregisterAllExtensions();
-    
+
+    /**
+     * Unregisters the extension represented by the given registration name.
+     * @param registrationName
+     */
+    void unregisterExtension(String registrationName);
+
     /**
      * This method frees all resources consumed by asciidoctorJ module. Keep in mind that if this method is called, instance becomes unusable and you should create another instance.
      */
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/Asciidoctor.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/Asciidoctor.java
@@ -652,12 +652,6 @@ public interface Asciidoctor {
     void unregisterAllExtensions();
 
     /**
-     * Unregisters the extension represented by the given registration name.
-     * @param registrationName
-     */
-    void unregisterExtension(String registrationName);
-
-    /**
      * This method frees all resources consumed by asciidoctorJ module. Keep in mind that if this method is called, instance becomes unusable and you should create another instance.
      */
     void shutdown();
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/Asciidoctor.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/Asciidoctor.java
@@ -13,6 +13,7 @@ import org.asciidoctor.ast.DocumentHeader;
 import org.asciidoctor.ast.DocumentRuby;
 import org.asciidoctor.ast.StructuredDocument;
 import org.asciidoctor.converter.JavaConverterRegistry;
+import org.asciidoctor.extension.ExtensionGroup;
 import org.asciidoctor.extension.JavaExtensionRegistry;
 import org.asciidoctor.extension.RubyExtensionRegistry;
 import org.asciidoctor.internal.JRubyAsciidoctor;
@@ -633,7 +634,18 @@ public interface Asciidoctor {
      * @return Converter Registry object.
      */
     JavaConverterRegistry javaConverterRegistry();
-    
+
+    /**
+     * Creates an ExtensionGroup that can be used to register and unregister a group of extensions.
+     * @return
+     */
+    ExtensionGroup createGroup();
+
+    /**
+     * Creates an ExtensionGroup that can be used to register and unregister a group of extensions.
+     * @return
+     */
+    ExtensionGroup createGroup(String groupName);
     /**
      * Unregister all registered extensions.
      */
--- a/asciidoctorj-core/src/main/resources/org/asciidoctor/internal/asciidoctorclass.rb
+++ b/asciidoctorj-core/src/main/resources/org/asciidoctor/internal/asciidoctorclass.rb
@@ -25,50 +25,50 @@ class AsciidoctorModule
         Asciidoctor::Extensions.unregister name
     end
 
-    def docinfo_processor(extensionName, registrationName = nil)
-        Asciidoctor::Extensions.register (registrationName != nil ? registrationName.to_sym : nil) do
+    def docinfo_processor(extensionName, groupName = nil)
+        Asciidoctor::Extensions.register (groupName != nil ? groupName.to_sym : nil) do
             docinfo_processor extensionName
         end
     end
 
-    def treeprocessor(extensionName, registrationName = nil)
-        Asciidoctor::Extensions.register (registrationName != nil ? registrationName.to_sym : nil) do
+    def treeprocessor(extensionName, groupName = nil)
+        Asciidoctor::Extensions.register (groupName != nil ? groupName.to_sym : nil) do
             treeprocessor extensionName
         end
     end
     
-    def include_processor(extensionName, registrationName = nil)
-        Asciidoctor::Extensions.register (registrationName != nil ? registrationName.to_sym : nil) do
+    def include_processor(extensionName, groupName = nil)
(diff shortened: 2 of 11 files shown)


mojavelinux:
Great work!

I saw the problem you reported in core regarding the type coercion of the group name, which I'll address in the 1.5.6.1 release.

My one suggestion is to use the term "group name" instead of "registration name". That would mean changing the unregisterExtension method to:

```java
void unregisterExtensionGroup(String name);
```

Although right now each group has one extension (when using this API), in the future we may merge subsequent registrations against the same group so the group holds multiple extensions.

That reads a bit better to me. Aside from that, I think this looks great.

mojavelinux:
There's one possible alternative to consider (or add). We may want to introduce the concept of an extension group, to which you'd assign a name, then register one or more extensions. This could be an alternative to adding the group name to each registration method. Instead, you'd only get to use a group name if you create a group, register extensions on that group, then register the group. That's really what is happening under the covers anyway.

In other words, group registration would be a separate chain in the API, started by initiating a group.

mojavelinux:
Actually, I really like this idea of starting with the group API. We'd use a createGroup method to create a group. That method would accept an option group name. If a name is not specified, one is generated. Either way, the group name is accessible via the ExtensionGroup#getName. When you're done adding extensions to the group, you'd pass it to the register method (or perhaps registration is implicit by creating the group).

We could then add the "a la carte" groups like you've done in this pull request later (or, if you really want them, we can leave them too).

wdyt?

robertpanzer:
Sounds good. 
With a Java 6 codebase I'd prefer to register eagerly ;), but lazy registration should still be feasible. 
I'll try to go for it as soon as i find the time, shouldn't be too much. 
Your idea is sth like this, right?
```java
ExtensionGroup g = asciidoctor.createGroup().block("yell").preprocessor(...);
g.register();
asciidoctor.convert(...);
g.unregister();
```

(On a small screen, please excuse any typos)

robertpanzer:
^^ The call to .block() certainly also has to take a class or instance.

mojavelinux:
Precisely.

I'm okay with registering eagerly. That's up to you. But the typed ExtensionGroup for register/unregister certainly has a nice feel to it.

robertpanzer:
@mojavelinux I updated the PR with the ExtensionGroup.
The extensions are now registered lazily.

It certainly needs more test cases.

asciidoctorj-core/src/main/java/org/asciidoctor/Asciidoctor.java
```diff
@@ -633,12 +634,29 @@ void convert(Reader contentReader, Writer rendererWriter,
...
      * Unregister all registered extensions.
      */
     void unregisterAllExtensions();
-    
+
+    /**
+     * Unregisters the extension represented by the given registration name.
+     * @param registrationName
+     */
+    void unregisterExtension(String registrationName);
```
mojavelinux: Should this be groupName?

asciidoctorj-core/src/main/resources/org/asciidoctor/internal/asciidoctorclass.rb
```diff
@@ -4,7 +4,9 @@ module Extensions
         include_package 'org.asciidoctor.extension'
         # Treeprocessor was renamed in to TreeProcessor in https://github.com/asciidoctor/asciidoctor/commit/f1dd816ade9db457b899581841e4cf7b788aa26d
         # This is necessary to run against both Asciidoctor 1.5.5 and 1.5.6
-        TreeProcessor = Treeprocessor
+        if !defined? TreeProcessor
```
mojavelinux: This is better written as:

```ruby
TreeProcessor = Treeprocessor unless defined? TreeProcessor
```

asciidoctorj-core/src/main/resources/org/asciidoctor/internal/asciidoctorclass.rb
```diff
@@ -19,50 +21,54 @@ def unregister_all_extensions()
...
     end
 
-    def docinfo_processor(extensionName)
-        Asciidoctor::Extensions.register do
+    def unregister_extension name
+        Asciidoctor::Extensions.unregister name
+    end
+
+    def docinfo_processor(extensionName, groupName = nil)
+        Asciidoctor::Extensions.register (groupName != nil ? groupName.to_sym : nil) do
```
mojavelinux: Don't put a space before the opening bracket. That causes a warning in some versions of Ruby.

You no longer need to coerce to a symbol as Asciidoctor 1.5.6.1 does that for you. Just pass the groupName as is.

build.gradle
```diff
@@ -45,7 +45,7 @@ ext {
   xmlMatchersVersion = '1.0-RC1'
 
   // gem versions
-  asciidoctorGemVersion = project.hasProperty('asciidoctorGemVersion') ? project.asciidoctorGemVersion : '1.5.5'
+  asciidoctorGemVersion = project.hasProperty('asciidoctorGemVersion') ? project.asciidoctorGemVersion : '1.5.6'
```
mojavelinux: Let's use 1.5.6.1.

mojavelinux:
This is looking really nice!

I made some comments inline.

asciidoctorj-core/src/main/java/org/asciidoctor/Asciidoctor.java
robertpanzer: Actually this method should go away completely.
Unregistration should only be possible for extension registered via the ExtensionGroup mechanism.
Otherwise I would also need to add the "named" registration methods to JavaExtensionRegistry and RubyExtensionRegistry and I wanted to avoid that and keep everything in the ExtensionGroup.

asciidoctorj-core/src/main/resources/org/asciidoctor/internal/asciidoctorclass.rb
robertpanzer: Good to know, thanks!
I put the space in there not for the method argument, but to wrap the conditional expression.
I will try to get around without it.

build.gradle
robertpanzer: Yes, that's part of the other PR, that's why I sticked with 1.5.6 here.

robertpanzer:
Updated the PR implementing the review comments.

build.gradle
mojavelinux: :+1:

asciidoctorj-core/src/main/resources/org/asciidoctor/internal/asciidoctorclass.rb
mojavelinux: In that case, you can write it as:

```
Asciidoctor::Extensions.register(groupName && groupName.to_sym)
```

or

```
Asciidoctor::Extensions.register(groupName ? groupName.to_sym : nil)
```

A method argument list is an implicit grouping expression.

asciidoctorj-core/src/main/java/org/asciidoctor/Asciidoctor.java
mojavelinux: :+1:

asciidoctorj-core/src/main/resources/org/asciidoctor/internal/asciidoctorclass.rb
robertpanzer: Completely overseen that I can remove the coercion. Will remove it, then I don't need the brackets anyway.

mojavelinux: :+1:

mojavelinux: It does require 1.5.6.1 though. To be clear, the coercion was fixed internally in 1.5.6.1.

robertpanzer:
Bumped to Asciidoctor 1.5.6.1 and removed the coercion.

robertpanzer:
I think it would make sense to squash this PR and put it on top of #563 
Otherwise I'll never get a green build without pulling the changes from that PR in here.

mojavelinux:
Seems reasonable to me!

robertpanzer:
Integrated into #563 
Therefore closing this PR.

PR #568, opened 2017-08-05
Add ExtensionGroup API and fix registration of multiple extensions

robertpanzer:
This PR brings the ExtensionGroup API to the 1.6.0 branch.

It seems that I got it wrong on the current master by assuming that I can do multiple executions of `Asciidoctor::Extensions.register` with the same groupName.
If I didn't get it wrong it overwrites a previous registration.
Therefore I hopefully fixed it here by collecting all extensions in list or maps per type and the registering them in one go in the Ruby code (asciidoctorclass.rb).

There was also an error with multiple executions of extensions that were registered as instances.
For these it was tried to call `setConfig` twice which was rejected with an IllegalStateException. I fixed this by adding a method `updateConfig` that can be executed multiple times and add to the existing config.
In such a case it is probably not clever to use the same Asciidoctor instance concurrently to render documents.

I copied many of the existing extension tests and migrated them to the group API to get a better coverage.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1374,6 +1374,30 @@ String content = asciidoctor.convertFile(new File(
 
 You can unregister all extensions by calling the `org.asciidoctor.Asciidoctor.unregisterAllExtensions()` method.
 
+Additionally, since AsciidoctorJ 1.5.6.1 it is possible to unregister specific extensions when using the ExtensionGroup API.
+This allows to register and unregister a group of extensions as a whole:
+
+[source,java]
+----
+BlockProcessor extension1 = ...
+Treeprocessor extension2 = ...
+ExtensionGroup extensionGroup =
+  asciidoctor.createGroup()
+    .block(extension1)
+    .treeprocessor(extension2);
+
+asciidoctor.convert(...) // <1>
+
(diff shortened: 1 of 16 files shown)


mojavelinux:
> It seems that I got it wrong on the current master by assuming that I can do multiple executions of Asciidoctor::Extensions.register with the same groupName.
If I didn't get it wrong it overwrites a previous registration.

Yes, it overwrites the previous registration. The groupName can only be used to hold a single set of extensions. In Asciidoctor core, and ExtensionGroup is used to bundle extensions together, which then get associated with a group name.

> Therefore I hopefully fixed it here by collecting all extensions in list or maps per type and the registering them in one go in the Ruby code (asciidoctorclass.rb).

Yes, that's correct.

> In such a case it is probably not clever to use the same Asciidoctor instance concurrently to render documents.

Yes, not unless the extensions are pre-activated. Though we could certainly consider trying to make this possible in the future.

robertpanzer:
Refactored ExtensionGroupImpl and asciidoctor_module.rb by moving the concrete registration into the java code again and avoid smearing the internals of ExtensionGroupImpl into the Ruby code.

PR #577, opened 2017-09-23
Add .travis-upstream-only.yml

robertpanzer:


mojavelinux:
```yml
install: true
script: bash test-asciidoctor-upstream.sh
```

The `install: true` line disables any action in this step of the build lifecycle.

robertpanzer:
My bad. 
I thought that the file was executed (seems like this is not possible) instead of being read by the rakefile and sent as a delta config. 
Will fix ASAP.

robertpanzer:
Changed the .travis-upstream-only.yml as commented.
That should be it, right?

mojavelinux:
Yep, that should do it. But we'll find out for sure soon enough :)

robertpanzer:
Thanks for reviewing! :)

mojavelinux:
I looks like it worked!

https://travis-ci.org/asciidoctor/asciidoctorj/builds/279391632

:tada:

robertpanzer:
Awesome, will cherry-pick it into the 1.6.0 branch today :+1:

PR #578, opened 2017-10-08
reenable CI notifications

mojavelinux:
At one time, the defaults for the notifications were annoying. Now they are reasonable and we depend on them being sent.

I reverted to the default, though I'm not sure if we need to set anything to get notifications in the gitter channel. Here's the template if we do need that:

```yml
notifications:
  webhooks:
    urls:
    - https://webhooks.gitter.im/e/3b53badb62267d16b33a
```


mojavelinux:
For more details, see https://docs.travis-ci.com/user/notifications/

robertpanzer:
Thanks :) :+1:

PR #585, opened 2017-10-22
Fix upstream build failing due to changes in behavior of creating dirs

robertpanzer:
The upstream build fails due after the changes to Asciidoctor wrt mkdir.
This PR fixes the test by explicitly setting the mkDirs option to true for that test.


mojavelinux:
This is good news. It actually means that the test was failing silently before (or, at least, it was not completing all the tasks, in this case writing the stylesheet).

mojavelinux:
Upstream builds FTW!

robertpanzer:
👍

PR #587, opened 2017-10-25
Fix upstream build 1.6.0

robertpanzer:
Same fix was already applied to master.
Also bump JRuby to 9.1.13.0.


robertpanzer:
Test on Windows fails due to Rouge incompatibility.

PR #590, opened 2017-11-10
Move "org.asciidoctor.ast" package to "asciidoctorj-api" project

jmini:
* convert classes to interfaces in api project
  * `Author`
  * `ContentPart`
  * `DocumentHeader`
  * `RevisionInfo`
  * `StructuredDocument`
* Create implementation in core project


asciidoctorj-api/build.gradle
```diff
@@ -0,0 +1,20 @@
+apply plugin: 'osgi'
+
+dependencies {
+}
+
+
+javadoc {
+  classpath = sourceSets.main.output + sourceSets.main.compileClasspath + project(':asciidoctorj-test-support').sourceSets.test.output
```
robertpanzer: I think we don't need the asciidoctorj-test-support project for the java docs, right?

jmini: This is probabelly a copy paste error (I am not really a gradle user). What would be the correct value? nothing? Do I need a javadoc section?

asciidoctorj-api/src/main/java/org/asciidoctor/ast/ContentPart.java
```diff
@@ -0,0 +1,27 @@
...
+
+    public String getTitle();
+
+    public Map<String, Object> getAttributes();
+
+    public String getContent();
+
+    public List<ContentPart> getParts();
+
+    public void setParts(List<ContentPart> parts);
```
jmini: I do not think that `setParts(List<ContentPart>)` should be on the API. The only usage is in [JRubyAsciidoctor line 182](https://github.com/jmini/asciidoctorj/blob/patch-1/asciidoctorj-core/src/main/java/org/asciidoctor/internal/JRubyAsciidoctor.java#L182) in `JRubyAsciidoctor.getContentPartFromBlock(StructuralNode, int, int)`. At this point the setter on `ContentPartImpl` is enough. 

This would be more consistent with the other method on this interface.

asciidoctorj-api/build.gradle
robertpanzer: I guess this line can be completely removed. For this project we shouldn't need any particular classpath.

We would need this module though to create a complete javadoc for the API in asciidoctorj-api and asciidoctorj itself, but we can check that later.

asciidoctorj-api/src/main/java/org/asciidoctor/ast/ContentPart.java
robertpanzer: :+1:

jmini:
Amended commit with the discussed modifications pushed...

asciidoctorj-api/src/main/java/org/asciidoctor/ast/Author.java
```diff
@@ -0,0 +1,28 @@
...
+
+    public void setMiddleName(String middleName);
+
+    public String getEmail();
+
+    public void setEmail(String email);
+
+    public String getInitials();
+
+    public void setInitials(String initials);
```
robertpanzer: I think all these setters should go be on the impls only.
This API is only to extract structural information about the document, there's no way to modify these objects and have an impact on the underlying document.

jmini:
Setters removed...

asciidoctorj-api/src/main/java/org/asciidoctor/ast/Author.java
```diff
@@ -0,0 +1,16 @@
+package org.asciidoctor.ast;
+
+public interface Author {
+
+    public String getFullName();
```
jmini: "public" is not necessary in interfaces

asciidoctorj-api/src/main/java/org/asciidoctor/ast/ContentPart.java
```diff
@@ -0,0 +1,25 @@
+package org.asciidoctor.ast;
+
+import java.util.List;
+import java.util.Map;
+
+public interface ContentPart {
+
+    public String getId();
```
jmini: "public" is not necessary in interfaces

asciidoctorj-api/src/main/java/org/asciidoctor/ast/DocumentHeader.java
```diff
@@ -0,0 +1,19 @@
+package org.asciidoctor.ast;
+
+import java.util.List;
+import java.util.Map;
+
+public interface DocumentHeader {
+
+    public List<Author> getAuthors();
```
jmini: "public" is not necessary in interfaces

asciidoctorj-api/src/main/java/org/asciidoctor/ast/RevisionInfo.java
```diff
@@ -0,0 +1,10 @@
+package org.asciidoctor.ast;
+
+public interface RevisionInfo {
+
+    public String getDate();
```
jmini: "public" is not necessary in interfaces

asciidoctorj-api/src/main/java/org/asciidoctor/ast/StructuredDocument.java
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * Structure document containing header and content chunks
+ * 
+ * @author marek
+ * 
+ */
+public interface StructuredDocument {
+
+    public List<ContentPart> getParts();
```
jmini: "public" is not necessary in interfaces

asciidoctorj-api/src/main/java/org/asciidoctor/ast/Row.java
```diff
@@ -4,6 +4,6 @@
 
 public interface Row {
 
-    public List<Cell> getCells();
+    List<? extends Cell> getCells();
```
robertpanzer: I think this is not going to work.
While the getters on StructuredDocument etc are read-only collections, these are meant to be mutated by extensions.
With this change extensions will no longer be able to add columns, headers, footers, and rows.
I guess we have to figure out how to make the Groovy based tests spot such things.
As these tests are not compiled static at the moment Groovy just adds values to these members, but Java wouldn't allow it.

So using wildcards for these 5 interfaces should be fine.
- Author
- ContentPart
- DocumentHeader
- RevisionInfo
- StructuredNode
Collections returned by all other AST interfaces should stay the same because they are intended to be mutated by extensions.

See for example https://github.com/asciidoctor/asciidoctorj/blob/5b8018792c787f59df8ee099d5f8393f68e7591b/asciidoctorj-core/src/test/groovy/org/asciidoctor/extension/GithubContributorsBlockMacro.groovy#L52

jmini:
I am not an expert in API design in Java, but here my ideas about it:

During my experimentations I have noticed that it would help me to have a relaxed definition of the generics parameter in `List` at API level. This is why I have proposed:

```
public interface Row {
    List<? extends Cell> getCells();
}
```

If you try to code a different implementation than the Ruby one (I use the prefix `My` in this example):
`MyRow implements org.asciidoctor.ast.Row` wants to return a list of `MyCell`. This implementation needs to control the implementation type of the `getCells()` list. Using `List<Cell>` in the implementation is wrong (for example it cannot accept `org.asciidoctor.ast.impl.CellImpl` or any other implementation that is out of control). 

The implementation in `MyRow` is also something like:
```
public List<MyCell> getCells() {
    return this.cells;
}
```

Maybe my idea was naive, but I have noticed that at API level this means that I need the wildcard form in order to be able to compile everything.

---

For this pull request, I will revert the wildcard modification as you suggested in [your review](https://github.com/asciidoctor/asciidoctorj/pull/590/files/827d4b41a3853ab21498999d337e72d6192a3da2#diff-ce9045f9d65ad9c89978f3ecf1bbf619).

Let move forward on the point that have a consensus. (The need for wildcard everywhere is only requested if my experiment is a success and maybe the vision I have for an alternative plain Java implementation of the Asciidoctor AST is wrong anyway).

robertpanzer:
The thing is that this shouldn't compile:

```java
Row row = createTableRow();
Cell cell = createTableCell(column, "Some content");
List<? extends Cell> cells = row.getCells();
cells.add(cell) 
```
The last statement shouldn't compile because `cells` could also be a `List<SomeStrangeCell>` if `SomeStrangeCell` implements `Cell`.
All the compiler knows about `cell` is that it implements the interface `Cell`, but not if it also extends `SomeStrangeCell`.

I see what you are aiming for and I agree that using plain lists also has its problems because if the user implements the AST interfaces himself without using our factory methods we fail as well.

asciidoctorj-api/src/main/java/org/asciidoctor/ast/Row.java
jmini: I guess you mean `StructuredDocument` (`StructuredNode` does not exist)

jmini:
Thank you for your feedback.

I had understood your point, but now we have it explicitelly written in this PR. I think it is valuable to write down the arguments (pro and contra), in order to be able to start this discussion again in the future (if necessary).

I have pushed a new version with the discussed changes.

asciidoctorj-api/src/main/java/org/asciidoctor/ast/Row.java
robertpanzer: Of course, you're right.
This should've been StructuredDocument.

asciidoctorj-core/src/test/java/org/asciidoctor/WhenAsciiDocIsRenderedToDocument.java
```diff
@@ -298,20 +298,20 @@ public void should_get_attributes() {
             "\n";
 
         Document document = asciidoctor.load(documentWithAttributes, new HashMap<String, Object>());
-        List<StructuralNode> blocks = document.getBlocks();
+        List<? extends StructuralNode> blocks = document.getBlocks();
```
robertpanzer: Please don't do this.
I tend to show these tests to answer questions.
`List<? extends StructuralNode>` hides the possibility that this list can be mutated.
But we encourage that, see for example here:
https://github.com/asciidoctor/asciidoctorj/blob/5b8018792c787f59df8ee099d5f8393f68e7591b/asciidoctorj-core/src/test/java/org/asciidoctor/extension/TerminalCommandTreeprocessor.java#L31-L39

jmini: Of course. Sorry I missed that one by reverting...

robertpanzer: Ah ok, no worries.
I thought this was intentional :-)

robertpanzer:
Thanks! 🎉

robertpanzer:
Merged as fe439317b198664ac4595ca14cd14ddbd32603ba

mojavelinux:
Should I file an issue, or would you rather do it?

robertpanzer:
It would be great if you could file an issue for it.
I would happily create a PR for it if nobody else wants to jumps in.

mojavelinux:
#694

PR #592, opened 2017-11-19
Fixes #588. Fix rendering to Java streams

### LINKED ISSUE #588: Writing generated doc to an output stream does not work (undefined method `<<')
If I try to write the generated document (in this case with backend PDF) to an output stream I get this exception:

```
Exception in thread "main" org.asciidoctor.internal.AsciidoctorCoreException: org.jruby.exceptions.RaiseException: (NoMethodError) undefined method `<<' for #<Java::JavaIo::ByteArrayOutputStream:0x7378ff35>
	at org.asciidoctor.internal.JRubyAsciidoctor.convert(JRubyAsciidoctor.java:460)
	at org.asciidoctor.internal.JRubyAsciidoctor.convert(JRubyAsciidoctor.java:486)
	at org.asciidoctor.internal.JRubyAsciidoctor.convert(JRubyAsciidoctor.java:481)
	at Test.main(Test.java:82)
Caused by: org.jruby.exceptions.RaiseException: (NoMethodError) undefined method `<<' for #<Java::JavaIo::ByteArrayOutputStream:0x7378ff35>
	at RUBY.render_header(uri:classloader:/gems/pdf-core-0.7.0/lib/pdf/core/renderer.rb:191)
	at RUBY.render(uri:classloader:/gems/pdf-core-0.7.0/lib/pdf/core/renderer.rb:164)
	at RUBY.render(uri:classloader:/gems/prawn-2.2.2/lib/prawn/document.rb:378)
	at RUBY.write(uri:classloader:/gems/asciidoctor-pdf-1.5.0.alpha.16/lib/asciidoctor-pdf/converter.rb:2826)
	at RUBY.write(uri:classloader:/gems/asciidoctor-1.5.6.1/lib/asciidoctor/document.rb:1083)
	at RUBY.convert(uri:classloader:/gems/asciidoctor-1.5.6.1/lib/asciidoctor.rb:1507)
	at RUBY.convert(<script>:85)
```

I call asciidoctorj like this:
```
ByteArrayOutputStream baos = new ByteArrayOutputStream();
Asciidoctor.Factory.create().convert(
		out,
		OptionsBuilder.options()
				.backend("pdf")
				.toStream(baos)
				.headerFooter(true)
				.safe(SafeMode.UNSAFE)));
```
The same happens for each other implementation of OutputStream.

Environment:
asciidoctorj 1.6.0 alpha 6
asciidoctorj-pdf 1.5.0 alpha 16
jruby 9.1.13.0

robertpanzer:
This PR tries to fix #588 as well as rendering HTML to OutputStreams.
The key is to wrap the OutputStream into sth that implements methods like `write` (used by Asciidoctor core) and `<<` and `size` (used by pdf-core).

Direct calls of `write` to an OutputStream don't work from Ruby because the method is overloaded and JRuby can't figure out which one to call.

I don't know if I messed up encodings, because I just get RubyStrings in and I simply convert them to bytes without using an encoding. Do I have to pass in an encoding that is available somewhere.

PDF generally works but in comparison to rendering to a File the title page is missing.
@mojavelinux Is there any chance that this is not an error in this PR but the normal behavior of Asciidoctor-PDF under certain circumstances?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,6 @@ target/
 *.ipr
 *.iws
 .idea/
+
+# Strange directories that IntelliJ seems to create now 
+out/
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/converter/ConverterProxy.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/converter/ConverterProxy.java
@@ -4,6 +4,7 @@ import org.asciidoctor.ast.ContentNode;
 import org.asciidoctor.ast.NodeConverter;
 import org.asciidoctor.internal.RubyHashMapDecorator;
 import org.asciidoctor.internal.RubyHashUtil;
+import org.asciidoctor.internal.RubyOutputStreamWrapper;
 import org.jruby.Ruby;
 import org.jruby.RubyClass;
 import org.jruby.RubyHash;
@@ -162,6 +163,8 @@ public class ConverterProxy<T> extends RubyObject {
         } else if (target instanceof RubyString) {
             File f = new File(((RubyString) target).asJavaString());
             out = new FileOutputStream(f);
+        } else if (target instanceof RubyOutputStreamWrapper) {
+            out = ((RubyOutputStreamWrapper)target).getOut();
         } else {
             throw new IllegalArgumentException("Can't write to a " + target);
         }
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/internal/JRubyAsciidoctor.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/internal/JRubyAsciidoctor.java
@@ -30,6 +30,7 @@ import org.jruby.runtime.builtin.IRubyObject;
 
 import java.io.File;
 import java.io.IOException;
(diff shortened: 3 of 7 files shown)


robertpanzer:
Seems like the property `document.notitle` is false when writing to a stream. Trying to find out why.

robertpanzer:
Ah found it.
When writing to a stream the `header_footer` option must be set. Then the title page is also rendered to the stream. Sounds fair.
Will do some more testing with rendering themes etc, but atm it looks like it's good to go. Dittaa diagrams already seemed to work fine.

robertpanzer:
Huh???
The Codenarc plugin failed because it suddenly loads a version of groovy that is only compatible with Java8?
Why??? I didn't change the version of Gradle

robertpanzer:
Seems to work now.
The test renders the same document to a PDF file and stream and checks that they contain the same byte sequence. So I think we're fine no.

Should we also try to fix this on master/1.5.x and possibly release a 1.5.6.1?

mojavelinux:
Is this something we should check for / support in core?

PR #600, opened 2017-12-03
Static analysis fixes

ham1:
- removed unused imports
- some JavaDoc fixes
- Utilise enhanced for loop
- removed unused assignments
- simplified some boolean logic
- replaced String concatenation in a loop with StringBuilder


robertpanzer:
Tests have failed on this PR.
Also instead of removing empty javadoc tags for parameters or return values, I'd rather prefer to have them filled with content. Keeping them empty makes it more obvious that sth is still missing here imo.

PR #603, opened 2017-12-05
Move "org.asciidoctor.converter" package to api project

jmini:


robertpanzer:
You also might want to move AbstractConverter and StringConverter.
These are base classes meant to be used when writing custom converters in Java.

jmini:
`AbstractConverter` and `StringConverter` moved to api project.

robertpanzer:
Looking at the code it would probably make sense to make JavaConverterRegistry an interface and move the implementation to `org.asciidoctor.converter.internal`.

But that can be done in a follow up PR.

PR #605, opened 2017-12-22
Remove wildcards in the generic type

jmini:


jmini:
I know we have discussed it in PR #590, but I have changed my mind about wildcard generic Parameter in the return type of methods returning a `List`.

Reasons:

1. Have a consistency between the two kind of classes:

* Wrapper classes not connected with Ruby internal model
(`StructuredDocument`, `DocumentHeader`, `ContentPart`...). => We have added the wildcard here.
* Classes doing a bridge with JRuby (`Document`, `StructuralNode`, `Title`, ...) => We decided not to add them there, because modifications on list to modify the AST are allowed, and working with wildcard as consumer of the API do not make it easy.

2) Working with wildcard Generic is really a pain. I have encountered problem with:
* [Mockito](https://stackoverflow.com/questions/47727114/mockito-when-for-list-with-wildcard)
* [AssertJ](https://stackoverflow.com/questions/47317425/assertj-containsexactly-assertion-on-list-with-wildcard)

robertpanzer:
:+1:

PR #606, opened 2017-12-22
Move part of 'org.asciidoctor' package to api project

jmini:
Split the `org.asciidoctor` package between api and core project.


robertpanzer:
Thanks! Looks good!

PR #615, opened 2018-01-11
Merge master into 1.6.0 branch

tisoft:
Updated the 1.6.0 branch with the changes made to master.
Mostly documentation updates, test cases and chocolaty packaging.


robertpanzer:
I am a bit reluctant to merge this PR.
My idea was rather that when there is a version 1.6.0 of Asciidoctor, that we would merge asciidoctorj-1.6.0 into master, and not the other way around.

tisoft:
I have no problem with that. I just wanted to see if there was anything missing in 1.6 that was already in 1.5, but since it is all only documentation and test cases, it is not really relevant. But since I already had done the merge, I just opened it as PR. 😁

PR #617, opened 2018-01-14
Fixes #616. Fix handling of showTitle attribute

### LINKED ISSUE #616: AttributesBuilder: showTitle(boolean) method
With the `showtitle` attribute, the document title visibility can be controlled. See section [17.1.3. of the user-manual](http://asciidoctor.org/docs/user-manual/#document-title-visibility).

According to the user manual, the value can be set with `showtitle` or unset with `showtitle!`.

The `org.asciidoctor.AttributesBuilder` class provides a `showTitle(boolean)` method.
In my opinion this doesn’t work as expected, because 
`AttributesBuilder.attributes().showTitle(true)` and `AttributesBuilder.attributes().showTitle(false)` have the same effect.

See this example:

```
org.asciidoctor.Asciidoctor asciidoctor;
org.asciidoctor.ast.Document document;
org.asciidoctor.OptionsBuilder optionBuilder;

String s = "= This is a title\n" +
        "\n" +
        "This is some text. showtitle value: ${showtitle}.\n";

asciidoctor = Asciidoctor.Factory.create();
optionBuilder = OptionsBuilder.options()
        .attributes(AttributesBuilder.attributes()
                .showTitle(true));
document = asciidoctor.load(s, optionBuilder.asMap());

System.out.println("--- With 'showtitle == true'");
System.out.println("-> attributes contains 'showtitle' key: " + document.getAttributes().containsKey("showtitle"));
System.out.println("-> 'showtitle' attribute value: " + document.getAttributes().get("showtitle"));
System.out.println(document.convert());
```

The output is (as expected):

```
--- With 'showtitle == true'
-> attributes contains 'showtitle' key: true
-> 'showtitle' attribute value: true
<h1>This is a title</h1>
<div class="paragraph">
<p>This is some text. showtitle value: $true.</p>
</div>
```

With:
```
asciidoctor = Asciidoctor.Factory.create();
optionBuilder = OptionsBuilder.options()
        .attributes(AttributesBuilder.attributes()
                .showTitle(false));
document = asciidoctor.load(s, optionBuilder.asMap());
```

The output is the same (this is not what a user would expect in my opinion):
```
--- With 'showtitle == false'
-> attributes contains 'showtitle' key: true
-> 'showtitle' attribute value: false
<h1>This is a title</h1>
<div class="paragraph">
<p>This is some text. showtitle value: $false.</p>
</div>
```

This is how you can unset the showtitle attribute with the `AttributesBuilder`:

```
asciidoctor = Asciidoctor.Factory.create();
optionBuilder = OptionsBuilder.options()
        .attributes(AttributesBuilder.attributes()
                .arguments("showtitle!"));
document = asciidoctor.load(s, optionBuilder.asMap());
```

And then it works:

```
--- With 'showtitle!'
-> attributes contains 'showtitle' key: false
-> 'showtitle' attribute value: null
<div class="paragraph">
<p>This is some text. showtitle value: ${showtitle}.</p>
</div>
```
What are the opinions?
Maybe other boolean attributes have the same problem...

robertpanzer:
Fixes #616 on master.
The handling of the showTitle attribute wasn't correct.
When showTitle is true, the `showtitle` attribute has to be set to any value.
When showTitle is false, the `notitle` attribute has to be set to any value. 
If this is ok I will propose another PR on asciidoctorj-1.6.0.

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/Attributes.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/Attributes.java
@@ -41,6 +41,7 @@ public class Attributes {
     public static final String LINK_ATTRS = "linkattrs";
     public static final String EXPERIMENTAL = "experimental";
     public static final String SHOW_TITLE = "showtitle";
+    public static final String NOTITLE = "notitle";
     public static final String ALLOW_URI_READ = "allow-uri-read";
     public static final String TOC_POSITION = "toc-position";
     public static final String TOC_2 = "toc2";
@@ -332,7 +333,13 @@ public class Attributes {
      *            value.
      */
     public void setShowTitle(boolean showTitle) {
-        this.attributes.put(SHOW_TITLE, showTitle);
+        if (showTitle) {
+            this.attributes.put(SHOW_TITLE, true);
+            this.attributes.remove(NOTITLE);
+        } else {
+            this.attributes.put(NOTITLE, true);
+            this.attributes.remove(SHOW_TITLE);
+        }
     }
 
     /**
--- a/asciidoctorj-core/src/test/java/org/asciidoctor/WhenAttributesAreUsedInAsciidoctor.java
+++ b/asciidoctorj-core/src/test/java/org/asciidoctor/WhenAttributesAreUsedInAsciidoctor.java
@@ -10,6 +10,7 @@ import static org.hamcrest.CoreMatchers.notNullValue;
 import static org.hamcrest.CoreMatchers.nullValue;
 import static org.hamcrest.CoreMatchers.startsWith;
 import static org.hamcrest.collection.IsArrayContaining.hasItemInArray;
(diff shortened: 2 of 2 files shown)


asciidoctorj-core/src/main/java/org/asciidoctor/Attributes.java
```diff
@@ -332,7 +333,11 @@ public void setTableOfContents(Placement placement) {
      *            value.
      */
     public void setShowTitle(boolean showTitle) {
-        this.attributes.put(SHOW_TITLE, showTitle);
+        if (showTitle) {
+            this.attributes.put(SHOW_TITLE, true);
+        } else {
+            this.attributes.put(NOTITLE, true);
```
robertpanzer: Maybe I have to remove the other attribute, so that the logic stays consistent if somebody calls `.setShowTitle(true).setShowTitle(false)`.

PR #618, opened 2018-01-15
Update travis build matrix

### LINKED ISSUE #566: Add MacOS build to TravisCI
While checking another project I saw they integrate macOs in their travis configuration and I think we should do too. Doesn't seem that hard.

The only issue I have detected is setting up 2 builds that install graphviz:
-  for linux we can try the new apt addon https://docs.travis-ci.com/user/installing-dependencies/#Installing-Packages-with-the-APT-Addon
- for mac we can try to use the included homebrew

I'd like to keep everything clean in the .travisci file, but if we can't, worst case scenario, we have to make a sh with some `if (macOs) do_this this, else do_that`.

One thing I don't get and may affect is this hostname thing here https://github.com/asciidoctor/asciidoctorj/blob/master/.travis.yml#L8

tisoft:
Closes #566

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,29 @@
-dist: precise
+os: linux
+dist: trusty
 language: java
-jdk:
-  - oraclejdk8
-  - oraclejdk7
-  - openjdk7
-before_script: unset GEM_PATH GEM_HOME JRUBY_OPTS
+matrix:
+  include:
+    - jdk: oraclejdk8
+    - jdk: oraclejdk7
+      dist: precise
+    - jdk: openjdk7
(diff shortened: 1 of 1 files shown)


robertpanzer:
Hmm... Looks like the OS X build doesn't want to start...

tisoft:
Yes. Looking at https://www.traviscistatus.com/ it seems that the queue is getting longer and longer for osx builds. I guess we just have to wait...

tisoft:
@robertpanzer the OS X build went through.

robertpanzer:
Thanks, @tisoft! This is awesome! 🎉

PR #628, opened 2018-02-21
Activate java 9 on master branch

jmini:


abelsromero:
It seems it won't be that easy... I re-launched the build twice and in both cases it failed with the same issue.
 But is this correct? version 1.5.7.dev?
```
Execution failed for task ':asciidoctorj:jrubyPrepare'.
> Could not resolve all files for configuration ':asciidoctorj:gems'.
   > Could not resolve rubygems:asciidoctor:1.5.7.dev.
     Required by:
         project :asciidoctorj
      > Skipped due to earlier error
      > Could not resolve rubygems:asciidoctor:1.5.7.dev.
         > Could not get resource 'http://rubygems-proxy.torquebox.org/prereleases/rubygems/asciidoctor/1.5.7.dev/asciidoctor-1.5.7.dev.pom'.
            > Could not GET 'http://rubygems-proxy.torquebox.org/prereleases/rubygems/asciidoctor/1.5.7.dev/asciidoctor-1.5.7.dev.pom'. Received status code 500 from server: Internal Server Error                                                                                                         
```

Also, in my Windows machine I get this with Java 9.0.4. Which makes me think we most likely must use jRuby 9.1.14.0 for Java9 builds.
```
> Task :asciidoctorj:jrubyPrepare                                                                                                                                                         
WARNING: An illegal reflective access operation has occurred                                                                                                                              
WARNING: Illegal reflective access by jnr.posix.JavaLibCHelper (file:/C:/Users/asalgadr/.gradle/caches/modules-2/files-2.1/org.jruby/jruby-complete/1.7.26/c09885af02af34266ed929f94cedcf8
7cc965f46/jruby-complete-1.7.26.jar) to method sun.nio.ch.SelChImpl.getFD()                                                                                                               
WARNING: Please consider reporting this to the maintainers of jnr.posix.JavaLibCHelper                                                                                                    
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations                                                                                     
WARNING: All illegal access operations will be denied in a future release                                                                                                                 
error reading or parsing C:\Users\asalgadr/.m2/settings.xml                                                                                                                               
ERROR:  Loading command: install (LoadError)                                                                                                                                              
        Unsupported platform: x86_64-windows                                                                                                                                              
ERROR:  While executing gem ... (NoMethodError)                                                                                                                                           
    undefined method `invoke_with_build_args' for nil:NilClass                                                                                                                            
```

robertpanzer:
I agree, we should upgrade to the latest version of JRuby. 
We should even go to 9.1.15.0 imo.

abelsromero:
I proposed 9.1.14.0 because is officially the first release that supports Java9.

jmini:
JRuby updated. 

The build is still failing, but I have the feeling that the Gradle part ends with `BUILD SUCCESSFUL in 3m 0s`

abelsromero:
> but I have the feeling that the Gradle part ends with BUILD SUCCESSFUL

It did :+1: 

The issue is with the bash script that tests agains upstream Asciidoctor master branch. https://github.com/jmini/asciidoctorj/blob/master/.travis.yml#L14

I run it locally and amogst other issues I found that `gem-maven-plugin` does not support J9. We need to update this to 1.1.5....this is never ending...

jmini:
`gem-maven-plugin` updated to version `1.1.5`

jmini:
Any idea what I need to do as next step to fix the build with Java 9?

robertpanzer:
I have no idea at the moment.
It looks like the build used Java 9.0.1.
It worked on 9.0.4 for me.
I retriggered the build, let's see if it will get a newer Java version.

robertpanzer:
Still on 9.0.1. :-(

AsciidoctorJ 1.6.0 is already on Java9 though and the .travis.yml looks a bit different there.
Maybe it works if we align the version on master with the one on the 1.6.0 branch.

jmini:
I have copied the two versions in a gist:

* [proposed change for master](https://github.com/jmini/asciidoctorj/blob/patch-7/.travis.yml) 
* [1.6.0 version](https://github.com/asciidoctor/asciidoctorj/blob/asciidoctorj-1.6.0/.travis.yml) 

Diff: https://gist.github.com/jmini/f8d089430d84e9b4d145cc04b1bca4bd/revisions?diff=split

The differences are:

*  `before_script:` line is present twice in `1.6.0`.
* `after_success:`, `notifications:` and `env:` addition.

I am not sure this will change anything.

robertpanzer:
The diffs look smaller than I thought.
Another interesting point though is that 1.6.0 is still running with the gem-maven-plugin at 1.0.6:
```
[INFO] --- gem-maven-plugin:1.0.6:initialize (default-initialize) @ asciidoctor ---
Downloading from central: http://repo.maven.apache.org/maven2/org/jruby/jruby-complete/9.1.14.0/jruby-complete-9.1.14.0.jar
```
(From https://api.travis-ci.org/v3/job/347633860/log.txt)

jmini:
I have tried to set the same versions on the 1.6 branch:

* JRuby: 9.1.15.0
* gem-maven-plugin: 1.1.5 

And I get the same problem there. => PR #636 

---

I have also seen that [JRuby 9.1.16.0](http://jruby.org/2018/02/21/jruby-9-1-16-0.html) was just released.

---

`gem-maven-plugin` tries to fetch a `1.5.7.dev` from `rubygems-proxy.torquebox.org`.
URL: http://rubygems-proxy.torquebox.org/prereleases/rubygems/asciidoctor/1.5.7.dev/asciidoctor-1.5.7.dev.pom

Does this version exists?

robertpanzer:
Yes, that's the idea of the test-asciidoctor-upstream script, to test AsciidoctorJ against the current snapshot of Asciidoctor to find errors as soon as possible.
Commits to master of Asciidoctor even trigger this script.

Let's try with 9.1.16.0 then, maybe that helps. 
If that doesn't work imo we should fall back to the versions that are currently in use on the 1.6.0 branch.

jmini:
I have updated JRuby, and I have also try to put the `Java 8` as first build and `Java 9` at the second position.

But now I have the feeling that travis is not running...

jmini:
The problem with `1.5.7.dev` is independent from Java 8 / Java 9.
I have raised https://github.com/asciidoctor/asciidoctor/issues/2590 to discuss it.

mojavelinux:
Something must have changed which broke how that script works. What the script is supposed to do is download the tarball from GitHub, build the gem, and install it into the local Maven repository. The Gradle build then turns around and uses that installed artifact. It should not be going out to the rubygems proxy to get the gem at all. If it is, something else is wrong.

robertpanzer:
It looks like the gem-maven-plugin appends a `-SNAPSHOT` when installing the `1.5.7.dev` gem so that the file `~/.m2/repository/rubygems/asciidoctor/1.5.7.dev/asciidoctor-1.5.7.dev.pom` contains `<version>1.5.7.dev-SNAPSHOT</version>`.
And this seems to make gradle look for the artifact on torquebox, which is understandable.

jmini:
From where does it install the `1.5.7.dev`? I can not follow...

For me the maven repo `http://rubygems-proxy.torquebox.org/prereleases/` (in reality a bridge to the ruby gem repo) is contacted. As explained by @mojavelinux in https://github.com/asciidoctor/asciidoctor/issues/2590 the SNAPSHOT are not publised as dev versions anyways.

---

One last point: why does the build try to build against a `1.5.7.dev` when a `1.5.6.2` was just released few days ago. It would have be logic to test against a `1.5.6.dev`, isn't it?

robertpanzer:
It is downloaded here:
https://github.com/asciidoctor/asciidoctorj/blob/6b5da5407ac6cf6d2c706c5e9d86867a3fc76d97/test-asciidoctor-upstream.sh#L17

This seems to be the evil point:
```
         > inconsistent module metadata found. Descriptor: rubygems:asciidoctor:1.5.7.dev-SNAPSHOT Errors: bad version: expected='1.5.7.dev' found='1.5.7.dev-SNAPSHOT'
```
asciidoctor-1.5.7.dev is in the local repo, but it contains a mismatching version 1.5.7.dev-SNAPSHOT.

I have the impression that it's somehow related to this logic:
https://github.com/torquebox/jruby-maven-plugins/blob/d835086c66e9f57d274047cd0505a278d8d54c36/gem-proxy/src/main/java/de/saumya/mojo/proxy/RubygemsApiVisitor.java#L43-L48

robertpanzer:
The thing is for the upstream build it shouldn't load the gem via torquebox.
Instead test-asciidoctor-upstream.sh gets the latest build result from the Asciidoctor Travis job, extracts the version from the file `lib/asciidoctor/version.rb` and installs it as such in the local Maven repo using the gem-maven-plugin.
The current asciidoctor master carries the version 1.5.7.dev, therefore this version.
And somehow during the installation, the generated pom gets the `-SNAPSHOT` appended to the `<version/>` but not to the filename.

robertpanzer:
I think I have it!
It should work if you change this line in test-asciidoctor-upstream.sh:
https://github.com/asciidoctor/asciidoctorj/blob/6ff1d1e34b4bc79285e0fb1fd98175510a99082d/test-asciidoctor-upstream.sh#L21
to
```sh
ASCIIDOCTOR_VERSION=`grep 'VERSION' ./lib/asciidoctor/version.rb | sed "s/.*'\(.*\)'.*/\1/" | sed "s/[.]dev$/.dev-SNAPSHOT/"`
```
Do you want to try that?

Currently I only see a lot of test error after that because the file `asciidoctor-1.5.7.dev/data/stylesheets/asciidoctor-default.css` doesn't seem to be there.

robertpanzer:
With the new version `1.5.7.dev-SNAPSHOT` the installation of the gem in the gradle part doesn't seem to be correct anymore...

So this line has to change as well:
https://github.com/asciidoctor/asciidoctorj/blob/6b5da5407ac6cf6d2c706c5e9d86867a3fc76d97/asciidoctorj-core/build.gradle#L24
to
```groovy
  include "gems/asciidoctor-*/data/**"
```

It also seems to work with Java9 then.

jmini:
I have added a commit in your name with the changes... Let see what the build does. (In case of sucess, I will do the same changes for `1.6.0` as part of the other pull request.

robertpanzer:
Whee! The build passed! \o/

Now I just wonder if we want to require JRuby >= 9 for asciidoctorj and therefore require Java 7.
Until now we/I always tried to keep the compatibility with Java 6 at least for asciidoctorj-core. (PDF requires Ruby 2, therefore JRuby 9 and therefore Java 7)
I know that in the ages of Java 10 nobody is probably really caring anymore about Java 6, but I'd like to get some feedback on this.

jmini:
> I'd like to get some feedback on this.

I have twitted the question: https://twitter.com/j2r2b/status/977997151967043584
And asked on the mailing list: http://discuss.asciidoctor.org/Asciidoctor-1-5-x-and-Java-6-td6232.html

Personally, I consider that Java 9 / 10 support is more important that Java 11.

jmini:
There was absolutelly no feedback about Java 6 and Asciidoctor 1.5.x... 

@mojavelinux have you any opinion? Are you managing the `@Asciidoctor` twitter account? Maybe you could propagate my twitt to improve visibility.

mojavelinux:
We're actually going to start laying down some policies for the project that make decisions like this automatic. But to foreshadow a bit, the Asciidoctor project is going to stop supporting platforms which no longer receive public updates. Here we'd look at OpenJDK. It's hard to find an authority, but according to https://access.redhat.com/articles/1299013 updates for 6 ended on December 2016. We're 2 years past that date and therefore expecting us to continue to support that version is unreasonable.

The reason for these policies will be to allow us to move faster, spending more time on improvements and less time on support for legacy platforms. This is consistent with the direction the software industry is moving in general (see Node, Ruby, etc).

mojavelinux:
(and I think we should end support for 7 this summer).

robertpanzer:
Cool, sounds like a good plan.
I just want to be sure that we have a common agreement on this.
I'll test the PR again and merge if all is fine.

Maybe we should also change the java source and target version to 7.
That way we're crystal clear that we don't support Java 6 anymore and not get pulled into discussions like "we could still run on Java 6 if we fix this or that".

robertpanzer:
Hmmm... this is super strange...
The Java 9 build on Travis doesn't seem to have executed the test-asciidoctor-upstream script.
I will trigger a rebuild.

robertpanzer:
Even the rebuild didn’t seem to run the test-asciidoctor-upstream.sh, or it exited immediately with 0.
But I ran it locally yesterday with J9, so it should be good. 
I’ll try to figure out what’s going on in another PR and merge this now.

robertpanzer:
But it looks like it's running test-asciidoctor-upstream.sh with Java 9  for the builds triggered by upstream changes: https://travis-ci.org/asciidoctor/asciidoctorj/jobs/363685502
It seems to fail though. I guess I have seen this error when building with Java 9.0.1, so I think it's ok for now.

Now that I am looking at the script one condition to skip it is if the current build is a PR...
I'll open a PR to figure out how the environment looks like.

Issue #632, opened 2018-02-27
Unit tests fetching included web resources are not working in a proxy environment


t-h-a-l-i-s:
The following Unit Tests of test class `WhenJavaExtensionIsRegistered` are failing due to a "`Connection refused`" within a proxied environment:

- a_include_processor_as_string_should_be_executed_when_include_macro_is_found
- a_include_instance_processor_should_be_executed_when_include_macro_is_found
- a_include_processor_should_be_executed_when_include_macro_is_found
- an_inner_class_should_be_registered
- an_inner_anonymous_class_should_be_registered

PR #633, opened 2018-02-27
#632: Set proxy system properties for unit tests

t-h-a-l-i-s:


robertpanzer:
Also runs without a proxy :-)
Thanks.

Issue #634, opened 2018-02-27
Unit test fails due to line ending issue


t-h-a-l-i-s:
Unit test `WhenAsciiDocIsRenderedToDocument#should_be_able_to_read_asset` fails on Windows due to `CRLF` line endings.

PR #635, opened 2018-02-27
#634: Set line endings for *.asciidoc files to LF to make the content…

t-h-a-l-i-s:
… comparing tests pass on Windows


robertpanzer:
Seems to make a lot of sense, thanks!

PR #640, opened 2018-04-08
Added echo messages to better see when upstream build is skipped

robertpanzer:


robertpanzer:
Ah, now I understand.
The upstream build only runs on the first job of the test matrix.
https://github.com/asciidoctor/asciidoctorj/blob/51c271cf8f55c8f8425e8981052ffd0cf3674198/test-asciidoctor-upstream.sh#L6-L8
It's the last subcondition that skips it for everything other than the first job.
As oraclejdk8 is the first entry in the matrix, it only builds the upstream build on Oracle Java 8.

PR #641, opened 2018-04-16
Redirect Asciidoctor logging to Java Util Logging

robertpanzer:
This PR anticipates the change to the loging in Asciidoctor core that is about to come in 1.5.7 and redirects all Asciidoctor logging to Java Util Logging.
So the PR will not build atm, except the upstream build. 

When creating a new Asciidoctor instance a Logger is installed that maps calls to its add() method to invocations of java.util.logging.Logger.log().

Currently this logging is always enabled. 
Should it be possible to circumvent it?
And if yes how? 

JUL might not be the most popular logging framework, but it allows to get it done without adding new dependencies and it should be possible to bridge it into slf4j or log4j.

Would it make sense to also add this to the master branch?


robertpanzer:
This PR relates to the changes as part of https://github.com/asciidoctor/asciidoctor/pull/2660

mojavelinux:
One of the most important parts of this change is an API to set the logger instance. While I agree JUL is the right default, the ability to use a custom logger is also very important. A custom logger could be used, for instance, to force the processor to fail fast when an error message occurs. Another case is to collect the errors per document and show them in a dedicated view.

robertpanzer:
My idea was that you could hook in any logger by creating a corresponding JUL handler.
Registering a Java Logger in Asciidoctor is not trivial because log levels have to be translated, possibly blocks (i.e. message suppliers) have to be invoked etc. That's why I thought providing JUL as an entry into the Java space would be a good idea.

Do you think it would make more sense to have a lower level abstraction to observe log entries?
I.e. just a simple Observable? 
(I am not suggesting any concrete implementation of an Observable here, just the concept) 

And then add a JUL Logger on top of that to have these entries published by default?

I updated the PR to set the org.asciidoctor.ast.Cursor as the first entry of the `LogRecord.parameters`.
So this should allow to capture the file and line for which sth was logged.

mojavelinux:
I have to admit I don't exactly know the best approach here, so I defer to you. I do like the idea of being able to provide a custom logger via JUL. That's certainly an option we want to support.

But that still leaves room for a custom logger when you aren't using JUL. I don't think this is any more difficult than a custom extension class. The Java class just needs to respond to the logger methods. I would assume in that case it would be up to the developer to send those messages somewhere. I wouldn't expect that to happen automatically.

The Observable-like approach would essentially be like adding a middleman. It's an interesting idea. In that case, it would make creating a custom logger easier since you'd only have to program to the Java API, and we could handle any necessary mapping to make the translation easier. It's something to explore. But if it turns out to require more design, then we might want to just offer the JUL integration first, then graduate into something more advanced.

asciidoctorj-core/src/test/groovy/org/asciidoctor/internal/RubyAttributesMapDecoratorSpecification.groovy
```diff
@@ -39,7 +39,7 @@ Lorem ipsum dolor
         !attributes.containsKey(1L)
         attributes.containsKey(ATTR_ONE)
         attributes.keySet().contains(ATTR_ONE)
-        attributes.get(ATTR_ONE).startsWith(blockStyle)
+        attributes.get(ATTR_ONE) == blockStyle
```
mojavelinux: Exactly. Now that 1.5.7 is out, we can make this switch.

mojavelinux:
I do have one more quick question, which is high level. Will it be possible for the Maven plugin to use this integration to determine when a warning or error is recorded? That's the primary use case we want to be sure to satisfy.

robertpanzer:
The nice thing about doing it via JUL is that I don't have to provide an API to register sth, it is already there!
In particular on the 1.5 line, where the JRuby instances are not managed properly (see JRubyRuntimeContext), it might be difficult to work properly with different observers per Asciidoctor instance.

Maybe we can try to add an API to register generic observers later on the 1.6 branch once use cases come up.
JUL would still work then and could be "rebased" to be a bridge from this observer to JUL.

The Maven plugin should be able to register an own Handler and capture all entries including the log level.
I assume that this also means that it would be highly appreciated to have on the 1.5 line, right? :-)

robertpanzer:
So let's try to get this into master first and then I'll try to do a 1.5.7 release of AsciidoctorJ.

abelsromero:
I found a bug while testing this. I wanted to see how errors in uri included documents are show and found that errors **in and after** a uri include are not returned. They are shown in the console however.
To reproduce:

* Add to `.attributes(AttributesBuilder.attributes().allowUriRead(true))` here https://github.com/robertpanzer/asciidoctorj/blob/loggers-master/asciidoctorj-core/src/test/java/org/asciidoctor/WhenAsciidoctorLogsToConsole.java#L45
* Edit `documentwithnotexistingfile.adoc` like this:
```
= My Document

include::unexistingdoc.adoc[]

include::unexistingdoc.adoc[]

include::https://raw.githubusercontent.com/asciidoctor/asciidoctor-maven-plugin/master/src/test/resources/src/asciidoctor/github-include.adoc[]

include::unexistingdoc.adoc[]
```

You'll see that the logger only returns 2 errors, if you move the uri include after the first one, you get 1, if you move it to the first line, you get 0...

Here is the console btw:
```
may 05, 2018 7:31:44 PM <script> convertFile
GRAVE: documentwithnotexistingfile.adoc: line 3: include file not found: C:/home/data/asalgadr/temp/asciidoctorj/asciidoctorj-core/out/test/resources/unexistingdoc.adoc
may 05, 2018 7:31:44 PM <script> convertFile
GRAVE: documentwithnotexistingfile.adoc: line 5: include file not found: C:/home/data/asalgadr/temp/asciidoctorj/asciidoctorj-core/out/test/resources/unexistingdoc.adoc
may 05, 2018 7:31:53 PM <script> convertFile
GRAVE: https://raw.githubusercontent.com/asciidoctor/asciidoctor-maven-plugin/master/src/test/resources/src/asciidoctor/github-include.adoc: line 1: level 0 sections can only be used when doctype is book
may 05, 2018 7:31:54 PM <script> convertFile
GRAVE: documentwithnotexistingfile.adoc: line 9: include file not found: C:/home/data/asalgadr/temp/asciidoctorj/asciidoctorj-core/out/test/resources/unexistingdoc.adoc
```

Also, `shouldBeRedirectToAsciidoctorJLoggerSystem` test fails in Windows because the `cursor.getDir()` returns the path with unix slash '/'. I think it should return it with `\`.

robertpanzer:
Thanks for the pointer on Windows! I didn’t think of it. 

I am not sure though what to do about the other problem. Iirc uri loading happens in the Ruby space, so I think Asciidoctor Core would have to handle that.

robertpanzer:
> I am not sure though what to do about the other problem. Iirc uri loading happens in the Ruby space, so I think Asciidoctor Core would have to handle that.

Seems to be sth else. The log entries hit the JavaLogger, but somehow don't reach the handler. I'll investigate. 
Thanks for finding this!

robertpanzer:
Duh, looks like Loggers can be GC-ed.
So I have to store them myself.
My motivation to not store them was that the log events could potentially come in with different `programes`.
Maybe it's best to use `asciidoctor` as a fix value. Wdyt?
I'll continue fixing on #643 to get this into the 1.5 line first.

robertpanzer:
I think it's unrelated to loading the uri or such, but more that the it just takes so long and consumes so much other memory that the logger is GC'ed.

abelsromero:
After proving the integration with `asciidoctor-maven-plugin` I only have two concerns. One major and one minor.

* (major) The translation of log levels is confusing and forces people to understand how Asciidoctor log levels translate into Java.
It's nice being able to jump from Ruby, to Java or even JS (Antora) when building docs. But having different levels in Java introduces a barrier to users.
Looking at `java.util.logging.Level` is seems to me we could create custom levels, the `Handler.isLoggable` should work provided we map the int values correctly.

* (minor) I think that in the long term we will need to add some abstraction layer or at least a default Handler to capture messages. Implementing a custom Handler to store the messages felt like re-inventing the wheel, and extracting the Cursor with `getParameters[0]` is not very intuitive.

Other that, @robertpanzer work is great as always!

robertpanzer:
The Ruby log levels are these:
```ruby
    # Low-level information, mostly for developers.
    DEBUG = 0
    # Generic (useful) information about system operation.
    INFO = 1
    # A warning.
    WARN = 2
    # A handleable error condition.
    ERROR = 3
    # An unhandleable error that results in a program crash.
    FATAL = 4
    # An unknown message that should always be logged.
    UNKNOWN = 5
```
Do you mean we should define own Level constants with these names that map to appropriate ints?

Or step back from the JUL based approach and have sth like this?

```java
asciidoctor.registerLogHandler(
  new AsciidoctorLogHandler() {
    public void handle(Severity severity, Cursor cursor, String message) {
      .... // If we only could use Lambdas...
    }
  }
);
```

(We would still need to register a JUL handler then, otherwise the message would get lost in /dev/null)

abelsromero:
> Duh, looks like Loggers can be GC-ed.

I confirmed that the Logger instance is not the same and the handlers are remover. But I find it hard to believe it due to a GC thing. If that were true, JUL would not a be a viable solution...and it's part of the "rt" 🤔. I am more inclined to believe is a classloader or jruby thing.

> Or step back from the JUL based approach and have sth like this?

Maybe because it's late and when I am exhausted I find it easier to think in functional terms, but I would reconsider what are the goals here:
1. Offer a plug-able mechanics to capture events during rendering?
2. Offer a "specific" mechanism to plug a custom logger?

*1* means we can make any custom interface and we don't really need to care about translating log events. Just create some `LogEvent` class with a custom level enum.
*2* means translating log levels and using some standard interface. If JUL is not working, maybe Sl4J?

does it make sense?

PR #643, opened 2018-05-05
Redirect Asciidoctor logging to Java Util Logging

robertpanzer:
This is the version of #641 for the master branch.
In addition to only the changes from that PR I had to also pull in some more classes that didn't exist yet on master, i.e. Cursor/CursorImpl and RubyObjectWrapper.

I also changed the test to programmatically configure and register a Handler, the Maven plugin should be able to do just that as well.

Additionally I bumped the version of Java to 7 "officially" so that I don't have to downgrade some of the imported code. We're already relying on Java 7 now anyway.


mojavelinux:
:+1:

I think this is an essential addition for 1.5.7, primarily for the build tools that use it.

abelsromero:
This works wonders! This really makes it really easy to handle the log from a build tool. And I can see the gradle plugin using `registerLogHandler` with a closure also.
But I found a bug in the `Cursor` class. `getFile` fails when he error is reported from an uri include. There's a class cast exception, I suppose cause the object is not a RubyString. I could fix it doing this `this.getProperty("file").toString()` instead of `getString("file")`.

Btw, I added a couple of silly comments to the code.

asciidoctorj-core/src/main/java/org/asciidoctor/log/internal/JavaLogger.java
```diff
@@ -84,68 +79,46 @@ public IRubyObject add(final ThreadContext threadContext, final IRubyObject[] ar
       progname = args[2].asJavaString();
```
abelsromero: this is marked as unused in my IDE

asciidoctorj-core/src/main/java/org/asciidoctor/log/LogHandler.java
```diff
@@ -0,0 +1,8 @@
+package org.asciidoctor.log;
+
+import org.asciidoctor.ast.Cursor;
```
abelsromero: unused import

asciidoctorj-core/src/main/java/org/asciidoctor/log/internal/JULLogHandler.java
```diff
@@ -0,0 +1,45 @@
...
+    private final static Logger LOGGER = Logger.getLogger("asciidoctor");
+
+    @Override
+    public void log(LogRecord logRecord) {
+        final java.util.logging.LogRecord julLogRecord =
+            new java.util.logging.LogRecord(
+                mapSeverity(logRecord.getSeverity()),
+                logRecord.getCursor() != null ? logRecord.getCursor().toString() + ": " + logRecord.getMessage() : logRecord.getMessage());
+
+        julLogRecord.setSourceClassName(logRecord.getSourceFileName());
```
abelsromero: cursor is not captured here.
I fixed it locally adding ` julLogRecord.setParameters(new Object[] { logRecord.getCursor() });`

robertpanzer:
Thanks for the feedback! This is exactly what I needed! :)
I was thinking about no longer storing the Cursor in the JUL LogRecord, but maybe it makes sense to still do that. 
I’ll implement your comments tomorrow. 

(And thanks a lot again!)

abelsromero:
> I was thinking about no longer storing the Cursor in the JUL LogRecord,

It's still not 100% user friendly. But it's nice to have it if it can be done easily. We just need to document it clearly.

robertpanzer:
@abelsromero Is this now in a usable state for you?
If yes I would merge it and then try to do a 1.5.7 release of AsciidoctorJ.

abelsromero:
There are some other conversations about possible features. Its fine for
me, but I'd wait a bit more
https://github.com/asciidoctor/asciidoctor-maven-plugin/issues/261#issuecomment-387643160

Wdyt?

mojavelinux:
@robertpanzer @abelsromero Truly great work on this you two! I'm impressed!

mojavelinux:
> There's a class cast exception, I suppose cause the object is not a RubyString. I could fix it doing this this.getProperty("file").toString() instead of getString("file").

Sorry about that. Sometimes file or dir is a URI. This is internal information that is leaking out. When I create the log records, I should probably `.to_s` them. For now, you can just do so on the Java side.

abelsromero:
> If you have some more requirements just let me know.

I only see interesting the option to register via SPI.

robertpanzer:
You mean to register an Observer automatically if it is declared in a file `META-INF/services/org.asciidoctor.log.LogHandler`?
Sounds good.

robertpanzer:
@mojavelinux Regarding https://github.com/asciidoctor/asciidoctor-gradle-plugin/issues/233#issuecomment-388483656 Do you think it would make sense to always run with the verbose option when not running from the CLI? This way you would only configure the amount of logging in your logging framework.

The CLI overrides `$VERBOSE` anyway to `true` or `nil`.

abelsromero:
Apologies for the silly question, but I don't get the purpose of verbose.

From the command info `Verbosely print processing information and configuration file checks to stderr.`, I assume messages are sent to stdout by default. But, they are always sent to stderr, with and without --verbose.
That's running directly the gem in Windows.

asciidoctorj-core/src/main/java/org/asciidoctor/log/internal/JULLogHandler.java
robertpanzer: I added this below.
(Just mentioning the obvious here so that I remember that I already handled this comment)

robertpanzer:
I think the default log device is stderr:
https://github.com/asciidoctor/asciidoctor/blob/d15bc65a942a8d3b186eab6e038bfc000c4d48f5/lib/asciidoctor/logging.rb#L85

In my opinion it would make sense to always log these warnings independently of a verbose setting, at least when running embedded in the Maven or Gradle plugin.
Wdyt?

robertpanzer:
Thanks for the clarification :)

So would it make sense then to always enable this verbose flag to see those warnings?

abelsromero:
> So would it make sense then to always enable this verbose flag to see those warnings?

I'd vote for yes.

On other topics, have you consider the option to register using SPI? (asking this I feel like when my manager comes in late Friday with a new requirement to implement)

robertpanzer:
:) No worries!

I considered adding an SPI and I already developed it (if I understood you correctly that it should be possible to register an Observer via the java.util.ServiceLoader mechanism).
I just forgot to push my changes :-( 
But now they should be here as well.

robertpanzer:
To go ahead I am planning to always set the `$VERBOSE` global variable.
In my opinion it really makes sense to always get information about invalid references.

When testing it I just stumbled over that the warning is logged without a context. 
@mojavelinux Would it make sense to add the context and give the user information about where the invalid reference is located in the sources?

abelsromero:
> Updated the PR to always set $VERBOSE=true and therefore always log invalid refs.
> Imo users will always be interested in seeing this.

I second that!

robertpanzer:
I'll merge for now.
Additional comments can be implemented in follow-up PRs.

mojavelinux:
> To go ahead I am planning to always set the $VERBOSE global variable.
In my opinion it really makes sense to always get information about invalid references.

We've decided to go the other way on this. $VERBOSE is not something that should be set by default. The fact that it is linked to the invalid references warning is only because those warnings are not always accurate. In that regard, we should think of it more like "debug mode". But $VERBOSE does reveal programming errors, so it can be useful for other reasons. But certainly no something we want to enable by default.

PR #644, opened 2018-05-05
Fix start error for asciidoctorj tool on OpenJDK 10

robertpanzer:
As reported by Russel Winder on Twitter the AsciidoctorJ tool crashes right after the start on OpenJDK 10 due to the JVM option `-XX:+UseFastAccessorMethods`.
This PR simply removes this flag.
While it might make AsciidoctorJ a bit slower when started from the command line, it's still better than not starting at all.
And it shouldn't hurt, because this flag isn't set anyway when running for example as part of the Maven plugin.


mojavelinux:
I'm totally fine with this change. While that flag made sense at the time, it no longer makes sense. :wave:

mojavelinux:
Thanks for reporting @russel!

robertpanzer:
There's a few things that I want to get in first for example support for loggers, where I am waiting for some feedback.
When that is in I'll do a new release.

PR #652, opened 2018-05-29
Bump JRuby to 9.2.0.0

robertpanzer:
This PR bumps JRuby to 9.2.0.0 for the 1.6. branch and removes use of some of the dynamic type mapping features that are used with the `AsciidoctorModule`, where we currently rely on JRuby automatically implementing it as a `org.asciidoctor.Asciidoctor`.

Before really merging this I'd like to add more tests for Ruby extensions, as not all extension types are covered yet.


robertpanzer:
Looks like Java 7 is no longer support by JRuby 9.2, therefore removed it from the travis build matrix.

robertpanzer:
Going ahead with the merge.

Issue #655, opened 2018-06-13
Upgrade to latest Tilt


msgilligan:
It looks like https://github.com/asciidoctor/asciidoclet/issues/45 has been fixed upstream in Tilt. See:

* https://github.com/rtomayko/tilt/pull/276
* https://github.com/rtomayko/tilt/blob/master/CHANGELOG.md#205-2016-06-02

I believe we need this fix made in asciidoctorj, not in Asciidoclet.

robertpanzer:
Would you like to create a PR for this?

msgilligan:
I'll try...

msgilligan:
PR created. We an discuss it further over [there](https://github.com/asciidoctor/asciidoctorj/pull/656).

robertpanzer:
Released with 1.5.7

PR #656, opened 2018-06-16
Update tiltGemVersion to 2.0.8

msgilligan:
For Issue #655

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -58,7 +58,7 @@ ext {
   openUriCachedGemVersion = '0.0.5'
   slimGemVersion = '3.0.6'
   threadSafeGemVersion = '0.3.6'
-  tiltGemVersion = '2.0.1'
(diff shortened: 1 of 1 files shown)


robertpanzer:
Awesome, thank you!
This looks good to me.
@mojavelinux Do you think that upgrade would have any impact? As far as I can tell, Asciidoctor core depends on tilt ~> 2.0.0 anyway and should therefore work with 2.0.8 as well, right?

msgilligan:
Sometimes I contribute to open source projects on byte at a time ;-)

robertpanzer:
If it’s the right byte you’re making OSS better one byte at a time :)

mojavelinux:
Yes, we should upgrade. Thanks for the bytes!

robertpanzer:
Thanks! I'll try to cherry-pick the commit into the 1.6.0 branch.

PR #657, opened 2018-06-16
Add plugin mechanism for Loggers

robertpanzer:
This PR ports the changes in #643 to the 1.6.0 branch.

Like in that PR there custom log receivers can be implemented as `org.asciidoctor.log.LogHandler` and by default these entries are also forwarded to JUL.
Also the `$VERBOSE` variable is always set to `true` to always notify messages for invalid references.


robertpanzer:
Merging for now.
Additional comments can be implemented in follow up PRs.

PR #658, opened 2018-06-19
Remove Java VM options from the distribution start script that are no…

robertpanzer:
…t support on OpenJDK10

This PR adds the fix for OpenJDK10 that was still missing.

Also removed the extra dependency on JRuby from the distribution as JRuby 9.2.0.0 is already in the lib path.


robertpanzer:
Oops, looks like the PDF integration test failed with JRuby 9.2.0.0.
I'll look into it.

robertpanzer:
Looks like there is some variable part in the output again even though I set the attributes `docdatetime` and `localdocdatetime`.

PR #670, opened 2018-08-05
Adds LogHandler documentation to README

abelsromero:
This PR adds a small documentation to explain how to register custom LogHandlers.

**This is WIP and before merge** I'd like to understand a couple of things I found in my tests in order to document them properly.

* paths in the `Cursor` are absolute always, I tried with relative Files and Strings. Is that expected?
* Some names are confusing to me: check the docs for the use of cursor.path, cursor.file and sourceFilename. Specially the latests. Is that also normal? If `sourceFilename` is always <script>, does it make sense to expose it?

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1501,6 +1501,105 @@ org.asciidoctor.extension.ArrowsAndBoxesExtension
 And that's all.
 Now when a `.jar` file containing the previous structure is dropped inside classpath of AsciidoctorJ, the `register` method will be executed automatically and the extensions will be registered.
 
+== Logs handling API
+
+[NOTE]
+====
+This API is inspired by Java Logging API (JUL).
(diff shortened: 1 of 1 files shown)


robertpanzer:
Thanks for this awesome contribution!
With regards to the properties of the Cursor I am also unsure how to fully define them.

Regarding support for Java 7, I would like to keep at least that one for AsciidoctorJ 1.5.x even though the support from the CI seems to be lacking.
We already kicked out Java 6 with a minor release.
But I would be fine to disable the CI build for Java 7 if it is no longer working.

abelsromero:
> With regards to the properties of the Cursor I am also unsure how to fully define them.

OK then, if it's fine for you, we can merge. 

> I would like to keep at least that one for AsciidoctorJ 1.5.x

I can have a look at it. If the issue is what I think it's easy

robertpanzer:
🙌 Thanks! 🎉

PR #673, opened 2018-08-10
Fix Java7 builds

abelsromero:
Same as with the maven plugin, the issue is related to the lack of TLS 1.2 by default in Java7. In this case, due to a change in jcenter SSL.

Sadly, setting `-Dhttps.protocols=TLSv1.2` did not work because gradle seems to ignore it. So, I upgraaded gradle to v 4.8.1 which is the official fix for it -> https://github.com/gradle/gradle/releases/tag/v4.8.1.


robertpanzer:
:+1: Thanks!

Issue #667, opened 2018-08-03
Does asciidoctorj v1.5.7 support asciidoctorj-pdf v1.5.0-alpha.16?


charleech:
I'm testing the asciidoctorj `v1.5.7` with [asciidoctor-pdf-example](https://github.com/asciidoctor/asciidoctor-maven-examples/tree/master/asciidoctor-pdf-example) by changing the `pom.xml `as the following: -

### First try

```xml
<asciidoctorj.version>1.5.7</asciidoctorj.version>
<jruby.version>9.1.16.0</jruby.version>
```
There are some error display at console

```
uri:classloader:/gems/prawn-svg-0.27.1/lib/prawn/svg/color.rb:212: warning: shadowing outer local variable - result
uri:classloader:/gems/prawn-svg-0.27.1/lib/prawn/svg/properties.rb:50: warning: `*' interpreted as argument prefix
uri:classloader:/gems/ttfunk-1.5.1/lib/ttfunk/table.rb:26: warning: instance variable @offset not initialized
...
uri:classloader:/gems/prawn-2.2.2/lib/prawn/font/ttf.rb:118: warning: instance variable @italic_angle not initialized
uri:classloader:/gems/ttfunk-1.5.1/lib/ttfunk/table.rb:26: warning: instance variable @offset not initialized
...
uri:classloader:/gems/prawn-2.2.2/lib/prawn/font/ttf.rb:118: warning: instance variable @italic_angle not initialized
uri:classloader:/gems/ttfunk-1.5.1/lib/ttfunk/table.rb:26: warning: instance variable @offset not initialized
...
```

### Second try

```xml
<asciidoctorj.version>1.5.7</asciidoctorj.version>
<jruby.version>9.1.16.0</jruby.version>

<sourceHighlighter>rouge</sourceHighlighter>
```
There are some error display at console

```
uri:classloader:/gems/prawn-svg-0.27.1/lib/prawn/svg/color.rb:212: warning: shadowing outer local variable - result
uri:classloader:/gems/prawn-svg-0.27.1/lib/prawn/svg/properties.rb:50: warning: `*' interpreted as argument prefix
uri:classloader:/gems/rouge-2.0.7/lib/rouge/lexers/bsl.rb:62: warning: Ambiguous first argument; make sure.
...
uri:classloader:/gems/rouge-2.0.7/lib/rouge/lexers/io.rb:28: warning: Ambiguous first argument; make sure.
...
[...]

PR #677, opened 2018-08-20
Remove setting  again to avoid JRuby warnings

robertpanzer:
Same as #676 for the 1.6.0 branch:

As discussed in #667 this PR removes setting the global variable `$VERBOSE` to true again.
This avoids the warnings produced by JRuby when loading the prawn and ttfunk gems:
```
uri:classloader:/gems/prawn-svg-0.27.1/lib/prawn/svg/color.rb:212: warning: shadowing outer local variable - result
uri:classloader:/gems/prawn-svg-0.27.1/lib/prawn/svg/properties.rb:50: warning: `*' interpreted as argument prefix
Aug 18, 2018 7:14:58 PM <script> convertFile
WARNUNG: invalid reference: foo
uri:classloader:/gems/ttfunk-1.5.1/lib/ttfunk/table.rb:26: warning: instance variable @offset not initialized
uri:classloader:/gems/ttfunk-1.5.1/lib/ttfunk/table.rb:26: warning: instance variable @offset not initialized
uri:classloader:/gems/ttfunk-1.5.1/lib/ttfunk/table.rb:26: warning: instance variable @offset not initialized
uri:classloader:/gems/prawn-2.2.2/lib/prawn/font/ttf.rb:118: warning: instance variable @italic_angle not initialized
uri:classloader:/gems/ttfunk-1.5.1/lib/ttfunk/table.rb:26: warning: instance variable @offset not initialized
uri:classloader:/gems/ttfunk-1.5.1/lib/ttfunk/table.rb:26: warning: instance variable @offset not initialized
uri:classloader:/gems/ttfunk-1.5.1/lib/ttfunk/table.rb:26: warning: instance variable @offset not initialized
uri:classloader:/gems/prawn-2.2.2/lib/prawn/font/ttf.rb:118: warning: instance variable @italic_angle not initialized
```

At the same time it also removes the warnings from Asciidoctor about invalid references because it is coupled to the same global variable.
To enable logging of invalid refs à la
```
Aug 18, 2018 7:14:58 PM <script> convertFile
WARNUNG: invalid reference: foo
```
again the user can simply require a file that sets this variable.
That means the user can create a file `verbose.rb` with this content:
```ruby
$VERBOSE=true
```
and then invoke asciidoctorj like this:
```
$ asciidoctorj --load-path <path to verbose.rb> -r verbose.rb file.adoc
```

For the asciidoctor-maven-plugin I am unsure atm about the best way to do this.
But it should work to package `verbose.rb` in an own jar and to add a dependency to this jar to the plugin configuration of the asciidoctor-maven-plugin and require verbose.rb.

### DIFF
--- a/asciidoctorj-core/src/main/resources/org/asciidoctor/internal/asciidoctorclass.rb
+++ b/asciidoctorj-core/src/main/resources/org/asciidoctor/internal/asciidoctorclass.rb
@@ -1,6 +1,3 @@
-# Set the global variable VERBOSE to true to get invalid refs into the log
-$VERBOSE=true
-
 module AsciidoctorJ
     include_package 'org.asciidoctor'
     module Extensions
--- a/asciidoctorj-core/src/test/java/org/asciidoctor/WhenAsciidoctorLogsToConsole.java
+++ b/asciidoctorj-core/src/test/java/org/asciidoctor/WhenAsciidoctorLogsToConsole.java
@@ -12,6 +12,7 @@ import org.jboss.arquillian.junit.Arquillian;
 import org.jboss.arquillian.test.api.ArquillianResource;
 import org.junit.After;
 import org.junit.Before;
+import org.junit.Ignore;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
 import org.junit.runner.RunWith;
@@ -48,6 +49,7 @@ public class WhenAsciidoctorLogsToConsole {
     @Before
     public void before() {
         asciidoctor = JRubyAsciidoctor.create();
+        TestLogHandlerService.clear();
     }
 
     @After
@@ -127,6 +129,7 @@ public class WhenAsciidoctorLogsToConsole {
     }
 
     @Test
+    @Ignore("Until logging of invalid refs is enabled by default")
     public void shouldLogInvalidRefs() throws Exception {
 
         final List<LogRecord> logRecords = new ArrayList<>();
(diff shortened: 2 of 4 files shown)


robertpanzer:
I am still struggling a bit with this, because it is a global setting.
That means when embedding AsciidoctorJ enabling verbose mode via options seems to be a bad choice.

Maybe I could embed a script that enables this flag in the AsciidoctorJ jar itself.
Then when enabling the `-v` flag this script would be required after creation of the instance.
And an integrator who embeds AsciidoctorJ would need to require this script.

Or we add such a method on the `Asciidoctor` interface, or overload all methods of `Asciidoctor.Factory` to provide additional methods that set this flag.
Both options have its advantages and drawbacks, if we get more options the number of methods would explode and a Builder pattern would make more sense, for only one additional option overloading is less intrusive.

mojavelinux:
> That means when embedding AsciidoctorJ enabling verbose mode via options seems to be a bad choice.

Yes, that's true, but that's because these are still experimental warnings.

https://github.com/asciidoctor/asciidoctor/issues/2722

Right now these warnings are really only accessible when using the CLI (unless you want to hack the magic global variable).

> Then when enabling the -v flag this script would be required after creation of the instance.

I agree that the CLI should hide the the assignment of the magic variable behind the `-v` flag. So it still may be necessary to toggle it explicitly, but it shouldn't be something the user has to worry about.

> Or we add such a method on the Asciidoctor interface, or overload all methods of Asciidoctor.Factory to provide additional methods that set this flag.

I don't think we should do that as the use of this magic variable is really only temporary (or something intended exclusively for the CLI).

mojavelinux:
I had thought about moving the verbose setting to the logger, which would certainly work, except that's not what "verbose" really means in this context. It's more like "pedandic" which means show warnings that may be overeager.

robertpanzer:
Ah ok, so you mean to only expose this setting publicly to the users of the CLI and integrators should set the `VERBOSE` variable themselves?

I'm ok with that, I just think that at least 90% of the users consume AsciidoctorJ via the maven or gradle plugin and that many are very keen on getting these warnings.

mojavelinux:
Yes, that's correct.

> many are very keen on getting these warnings.

I'm definitely aware of that. The problem is, in a lot of cases, these particular warnings are wrong. That's why they are hidden behind the flag. Kind of think of them like debugging warnings atm.

Issue #680, opened 2018-09-25
Cannot render more than N documents on the same Asciidoctor instance (fixed with jRuby 9.1.17.0)


abelsromero:
AsciidoctorJ: 1.5.7
jRuby: 9.1.16.0 (default)
OS: Windows 10
JDK: 1.8.0_181

_This is one of those weird issues..._

Since version 1.5.7, if I try to render a set of documents reusing the same Asciidoctor instance, the process fails stating that it cannot read a document (see below). The error is always the same and happens on the same document.
I have been able to reproduce the issue rendering the same document in a loop and I have found the following:
* The iteration in which it fails depends on the document. With the documents I have used for testing it sometimes fails on the loop 53, others on 62 🤷‍♂️ .
* Using JRuby `9.1.17.0` seems to fix it (500 loops without issues), jRuby 9.2.0.0 does not work at all.

```
sep 25, 2018 3:09:53 PM org.asciidoctor.internal.JRubyAsciidoctor renderFile
GRAVE: (NoMethodError) asciidoctor: FAILED: C:/Users/asalgadr/github/asciidoctorj-gradle/src/asciidoc/user-password-expiration.adoc: Failed to load AsciiDoc document - undefined method `start_with?' for nil:NilClass
Exception in thread "main" org.asciidoctor.internal.AsciidoctorCoreException: org.jruby.exceptions.RaiseException: (NoMethodError) asciidoctor: FAILED: C:/Users/asalgadr/github/asciidoctorj-gradle/src/asciidoc/user-password-expiration.adoc: Failed to load AsciiDoc document - undefined method `start_with?' for nil:NilClass
	at org.asciidoctor.internal.JRubyAsciidoctor.renderFile(JRubyAsciidoctor.java:371)
	at org.asciidoctor.internal.JRubyAsciidoctor.renderFile(JRubyAsciidoctor.java:484)
	at org.asciidoctor.internal.JRubyAsciidoctor.convertFile(JRubyAsciidoctor.java:585)
	at org.asciidoctor.demos.AsciidoctorRunner.main(AsciidoctorRunner.java:30)
Caused by: org.jruby.exceptions.RaiseException: (NoMethodError) asciidoctor: FAILED: C:/Users/asalgadr/github/asciidoctorj-gradle/src/asciidoc/user-password-expiration.adoc: Failed to load AsciiDoc document - undefined method `start_with?' for nil:NilClass
[...]

PR #682, opened 2018-09-30
added test to reproduce issue with jRuby 9.1.16.0

abelsromero:
Related to issue #680 

Trying to reproduce it in the CI


robertpanzer:
Tested locally that the test works with the current master.
Thanks for finding this! 👍

PR #689, opened 2018-10-10
Disable automatic logging of invalid refs and enable it via verbose m…

robertpanzer:
…ode from CLI

This PR disables the automatic activation of the verbose mode again and thus disables logging of invalid refs.

The `-v` argument for the CLI additionally enables this again.
Therefore integrators have to run the Ruby script `$VERBOSE=true` themselves before rendering documents.

@mojavelinux Is this what you had in mind?

### DIFF
--- a/asciidoctorj-core/src/test/java/org/asciidoctor/WhenAsciidoctorLogsToConsole.java
+++ b/asciidoctorj-core/src/test/java/org/asciidoctor/WhenAsciidoctorLogsToConsole.java
@@ -9,10 +9,12 @@ import org.asciidoctor.log.TestLogHandlerService;
 import org.asciidoctor.util.ClasspathResources;
 import org.junit.After;
 import org.junit.Before;
+import org.junit.FixMethodOrder;
 import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
+import org.junit.runners.MethodSorters;
 
 import java.io.File;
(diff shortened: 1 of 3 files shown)


mojavelinux:
Exactly.

asciidoctorj-core/src/test/java/org/asciidoctor/WhenAsciidoctorLogsToConsole.java
```diff
@@ -135,6 +136,9 @@ public void log(LogRecord logRecord) {
         };
         asciidoctor.registerLogHandler(logHandler);
 
+        // Asciidoctor currently only logs invalid refs if this global var is set
+        JRubyRuntimeContext.get().evalScriptlet("$VERBOSE=true");
```
mojavelinux: Doesn't this need to be unset again after the test is run? Or do tests not share the JRuby runtime context?

robertpanzer: It should be fine, hopefully. 
Every Asciidoctor instance gets its own JRuby instance and should have it's own global state if I am not wrong.
But definitely sth to retest.

robertpanzer:
Added a second test to ensure that invalid refs log messages are not created by default.

I don't really like it, but also ensured that this test runs after the first that shows that $VERBOSE results in these messages being created to prove that the $VERBOSE global variable doesn't leak into other Asciidoctor instances.

PR #692, opened 2018-10-14
Further split into api and jruby impl modules.

robertpanzer:
This PR continues the separation into the asciidoctorj-api and asciidoctorj (core) module.
For an integrator it should be sufficient to only depend on the api module.
At the same time it would be possible to implement the implementation interfaces with something else if somebody feels she/he wants to do it.

The split happens through the service interface `org.asciidoctor.extension.spi.AsciidoctorFactory`.
asciidoctorj-core contains an implementation for this interface that uses the traditional JRuby implementation.

Even though the interface should be language independent, the `Asciidoctor.Factory.create()` methods are still the same and refer to JRuby specific things like gem paths or load paths to stay API compatible.

Even though the tests still pass (at least locally) unchanged, binary compatibility might be broken with the PR as some classes were changed to interfaces.

I also tried to keep the packages disjoint between the two modules. Therefore I moved some of the classes to other implement or internal packages. 
The naming might not seem consistent what is implementation and what is called internal, I am so bad at naming things.


robertpanzer:
Closed in favour of #700 
We can still try to have disjoint packages between the asciidoctorj-api and asciidoctorj-core later.

PR #693, opened 2018-10-15
resolves #508 add method to change sourcemap option on Document at runtime

### LINKED ISSUE #508: Setting options when using AsciidoctorInvoker
My goal is to add line numbers to the error messages produced by my custom macro processors. Unless I missed something, I have to set the `:sourcemap` option when converting a document to get line numbers from the nodes.

I currently rely on AsciidoctorInvoker when calling AsciidoctorJ from the command line. There seems to be no way of setting that particular option (and many others) when using AsciidoctorInvoker. I would rather not resort writing my own CLI for AsciidoctorJ, so it would be great to be able to set all options using AsciidoctorInvoker.

mojavelinux:


mojavelinux:
This PR needs tests.

PR #695, opened 2018-10-16
Add timings command option and fix timings

robertpanzer:
This PR adds a `-t` command line option to align with the original `asciidoctor` command.
Previously the timings should've been printed when starting with `-v` (verbose mode).

But it looks like this didn't even work, as the timings are now managed by the `Asciidoctor::Timings` class which has to be present in the options.
I fixed this by explicitly creating such an instance, and then after the rendering calling this to print its report, so that the report should also look identically to what asciidoctor produces.

Along the way I also found an error in the tests regarding the destination dir option `-D`.
The test `destination_dir_should_render_files_to_ouput_directory()` also fails on the current master when running in isolation.
It only succeeded because the test `verbose_option_should_fill_monitor_map()` was executed before.

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/cli/AsciidoctorInvoker.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/cli/AsciidoctorInvoker.java
@@ -79,8 +79,10 @@ public class AsciidoctorInvoker {
     }
 
     private void setTimingsMode(AsciidoctorCliOptions asciidoctorCliOptions, Options options) {
-        options.setOption("timings",
-            JRubyRuntimeContext.get().evalScriptlet("Asciidoctor::Timings.new"));
+        if (asciidoctorCliOptions.isTimings()) {
+            options.setOption("timings",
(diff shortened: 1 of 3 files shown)


asciidoctorj-core/src/main/java/org/asciidoctor/cli/AsciidoctorInvoker.java
```diff
@@ -31,19 +32,13 @@ public void invoke(String... parameters) {
...
-            asciidoctor = buildAsciidoctorJInstance(asciidoctorCliOptions);
+            Asciidoctor asciidoctor = buildAsciidoctorJInstance(asciidoctorCliOptions);
             
             if (asciidoctorCliOptions.isVersion()) {
                 System.out.println("Asciidoctor " + asciidoctor.asciidoctorVersion() + " [http://asciidoctor.org]");
                 return;
             }
 
-            if (asciidoctorCliOptions.isVerbose()) {
-                JRubyRuntimeContext.get().evalScriptlet("$VERBOSE=true");
```
robertpanzer: I removed this again, as it was apparently handled in the same way below in `setVerboseLevel()` which I oversaw.

PR #699, opened 2018-10-21
Remove ContentPart and StructuredDocument API

robertpanzer:
This PR removes the ContentPart and StructuredDocument APIs from the 1.6 branch.
The AST API is a better replacement for this.


mojavelinux:
This looks correct.

If we find there's a need to add capabilities that match closely with the functionality that was in this API based on the official AST to help users transition, we can always add those APIs later. However, we have been saying for years not to use this API.

PR #700, opened 2018-10-22
Make Asciidoctor expose interfaces exclusively

robertpanzer:
Currently the return types of some methods of the Asciidoctor interface are classes that are tied to the JRuby implementation.
This PR refactors the code by extracting interfaces for JavaExtensionRegistry, RubyExtensionRegistry and JavaConverterRegistry.
As we are already binary incompatible with 1.5 we should use that as a chance to allow further changes in the future. (For example plumbing the Asciidoctor.js implementation behind the Asciidoctor interface)


mojavelinux:
An important step forward!

PR #706, opened 2018-10-28
Add wildfly integration test

robertpanzer:
This PR adds an integration test for Wildfly, which is special for AsciidoctorJ due to the jboss modules class loading.
As the Arquillian Wildfly seems to require an external installation this test currently can only be executed locally after installing Wildfly.
I was not aware that a Wildfly distribution can be downloaded from Maven central.

To run the test execute:
```
./gradlew clean integrationTest -Pjboss.home=<path to wildfly>
```

Thanks to @mgreau for providing the blueprint for this in the docker-asciidoctorj repo!

### DIFF
--- a/asciidoctorj-wildfly-integration-test/build.gradle
+++ b/asciidoctorj-wildfly-integration-test/build.gradle
@@ -54,37 +54,43 @@ test {
 
 }
 
-if (project.hasProperty('jboss.home')) {
 
-    task integrationTest(type: Test) {
-        useJUnit {
-            includeCategories 'org.asciidoctor.categories.Wildfly'
+test {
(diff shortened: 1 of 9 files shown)


mojavelinux:
It's great to see that this case will be continually tested. That should circumvent a lot of support requests.

Is it possible to have the test automatically skip if the path the Wildfly is not defined? Or is that how it already works?

robertpanzer:
I changed the PR to always run this integration test when the property `jboss.home` is defined.
At the same time we currently would not run this test on the CI.
I am a bit afraid of downloading the wildfly distro with every CI build as this will probably not be cached by TravisCI or Appveyor.

PR #711, opened 2018-11-03
Remove deprecated methods (render*) from the Asciidoctor API

Mogztter:


Mogztter:
@robertpanzer The task `:asciidoctorj-documentation:test` is failing on AppVeyor, do you know why ? Can we fix it or skip it ?

robertpanzer:
@Mogztter I opened #713 to fix this.
Thanks for the heads up :)

robertpanzer:
The asciidoctor-maven-plugin currently invokes the renderFile() method:
https://github.com/asciidoctor/asciidoctor-maven-plugin/blob/d82b333ce8edf93b0c928bd1f375a9b8a245d707/src/main/java/org/asciidoctor/maven/AsciidoctorMojo.java#L463

The ant plugin uses renderFile() as well: 
https://github.com/asciidoctor/asciidoctor-ant/blob/ee15e1e0cf5c9582a76c7ec89be586e4cdb3735c/src/main/java/org/asciidoctor/ant/AsciidoctorAntTask.java#L96

It would be great if we could change the plugins first and then remove these methods.

The Gradle plugin is already using convertFile().

robertpanzer:
I just saw that we actually never deprecated the `render*()` methods in AsciidoctorJ.
I am fine with removing them once the maven plugin is ready to handle it.
But it might come as a surprise and we cannot say: "but we told you" :-D

Mogztter:
@robertpanzer Alright. Actually the implementation was marked as `@Deprecated` but not the interface :wink:

robertpanzer:
Oops... That's really evil :-D
I created PR for the maven plugin to use `convertFile()` instead of `renderFile()`. Once this is merged and released I would feel safe to really remove these methods.

abelsromero:
Didn't see these comments.
The idea with the maven-plugin is to go directly from 1.5.7 to the 2.0.x. But we can still do a 1.5.8 release that already uses `convert` (no breaking change has been done).
That way the next 1.6.x-RC could also be tested with the maven-plugin.

Mogztter:
Rebased on master :+1:

abelsromero:
> Rebased on master 👍

Does this mean we agree on the maven-release plan?

Mogztter:
> Does this mean we agree on the maven-release plan?

@abelsromero Sure

> The idea with the maven-plugin is to go directly from 1.5.7 to the 2.0.x. But we can still do a 1.5.8 release that already uses convert (no breaking change has been done).

Sounds good.

> That way the next 1.6.x-RC (if any) could also be tested with the maven-plugin.

Since the `render` methods were not properly marked as `@deprecated` and they were present in the first release candidate, I think it's better to remove them in 2.0.0.
But you can already move to the `convert` methods in your next release (without upgrading AsciidoctorJ dependency).

@robertpanzer I think we should properly marked the `render` methods as deprecated in 1.6.0 with a comment saying "will be removed in 2.0.0 (next release)". What do you think ?

mojavelinux:
> > I think we should properly marked the render methods as deprecated in 1.6.0 with a comment saying "will be removed in 2.0.0 (next release)".
> >
> Makes absolute sense to me.

I second that! That's using the 1.6.0 / 2.0.0 releases exactly as intended.

robertpanzer:
I think it should be ok to close this PR, all render methods were already removed with #760.
@Mogztter: Is that ok for for you?

Mogztter:
Sure 👍

PR #712, opened 2018-11-03
Remove deprecated Factory.create methods with classLoader

Mogztter:


asciidoctorj-core/src/main/java/org/asciidoctor/Asciidoctor.java
```diff
@@ -810,32 +810,6 @@ public static Asciidoctor create(List<String> loadPaths) {
...
-         * used in environments where different threads may have different classloaders, like build tools sbt or ANT.
-         *
-         * @param classloader
-         * @return Asciidoctor instance which uses JRuby to wraps Asciidoctor
-         *         Ruby calls.
-         * @deprecated Please use {@link #create()} and set the TCCL before or {@link #create(List)} passing the paths
-         * that you would have used to create the ClassLoader.
-         */
-        public static Asciidoctor create(ClassLoader classloader) {
-            return JRubyAsciidoctor.create(classloader);
```
robertpanzer: I think these methods can also be removed.

robertpanzer:
I guess it's fine to remove these methods, I checked that the plugins, including the ant plugin no longer use them.
See #340 for the discussion about deprecating these methods.
It would be just great to also remove the implementations from JRubyAsciidoctor.

asciidoctorj-core/src/main/java/org/asciidoctor/Asciidoctor.java
Mogztter: Done

robertpanzer:
Thank! :)

asciidoctorj-core/src/main/java/org/asciidoctor/Asciidoctor.java
robertpanzer: @Mogztter I think we have to resurrect these methods.
I wasn't aware of this either, but it seems like these methods are needed to use AsciidoctorJ in an OSGI environment (see #721)
Until we have a replacement we should keep these methods.

PR #715, opened 2018-11-05
WIP: Move the extension API to the api module

Mogztter:
Just to keep the ball rolling :soccer: 

Some issues with the current implementation:

* ~~Groovy expects a static resolution of the processor, so I had to add `static JRubyProcessor processor = new JRubyProcessor()` in the tests~~
* Groovy does not support default methods in interfaces, so I hard to re-declare default methods in `JRubyProcessor` and `BaseProcessor`
* ~~I had to break the API in few places~~

The idea is to use this pull request as a starting point. Ideas ? Thoughts ?


asciidoctorj-core/src/test/groovy/org/asciidoctor/extension/TableCreatorTreeProcessor.groovy
```diff
@@ -10,41 +10,43 @@ import org.asciidoctor.ast.Table
 @CompileStatic
 class TableCreatorTreeProcessor extends Treeprocessor {
 
+    static JRubyProcessor processor = new JRubyProcessor()
```
robertpanzer: It is problematic to have this static, if we register the same `TableCreatorTreeProcessor` with two different Asciidoctor instances.
Every Asciidoctor instance has its own Ruby instance.
Therefore if we have it static here, the nodes will be created only in one Ruby instance, and somehow transition from one Ruby instance into the other.
This might work most of the time, but not always.
There are a few subtle differences when using objects that come from 2 Rubys, for example Symbols are no longer the same.

robertpanzer: I would prefer if the `create...()` methods would continue to just work without having to manually create a JRubyProcessor or JSProcessor.
The ExtensionRegistry could inject a node factory at registration time like it does with the config.
That way it would be completely transparent in which environment the processor is running, and maybe it would even work to reuse the same processor in both environments.

Mogztter: I completely agree but I didn't know how to work around this issue.

> The ExtensionRegistry could inject a node factory at registration time like it does with the config.

You mean in the `ExtensionGroupImpl` ? I think we can use the proxy to define the processor but I hope that Groovy won't complain.

Mogztter: Actually it's unnecessary :wink: 
I was confused by Groovy but as long as we don't use `default` methods, it's working!

asciidoctorj-core/src/test/groovy/org/asciidoctor/extension/WhenAJavaExtensionUsesCounters.groovy
```diff
@@ -169,10 +170,10 @@ testmacro::countera[]
             document
         }
 
-        Section processSection(Section section) {
+        static Section processSection(Section section, TestTreeProcessor that) {
             List<StructuralNode> newNodes = section.blocks.collect {
                 if (it instanceof Block && it.getRoles().contains('count')) {
-                    createBlock(section, PARAGRAPH, "Counter ${it.content} has value ${it.document.getAndIncrementCounter(it.source)}")
+                    that.createBlock(section, PARAGRAPH, "Counter ${it.content} has value ${it.document.getAndIncrementCounter(it.source)}")
```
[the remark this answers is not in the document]
Mogztter: Good catch!
It was needed when I had issues with Groovy and default methods but now it should be reverted.

asciidoctorj-api/src/main/java/org/asciidoctor/extension/BaseProcessor.java
```diff
@@ -0,0 +1,216 @@
...
+        this(new HashMap<>());
+    }
+
+    public BaseProcessor(Processor delegate) {
+        this(delegate, new HashMap<>());
+    }
+
+    public BaseProcessor(Map<String, Object> config) {
+        // FIXME: Use a strategy to resolve the processor
+        this(ServiceLoader.load(Processor.class).iterator().next(), config);
```
Mogztter: I think we should introduce a "service" to resolve the implementation so the logic can be easily tested.
It will also ensure that the logic is the same everywhere.

Mogztter:
Thanks for your review @robertpanzer 
I'm currently in vacation but when I'm back I will do a follow-up :wink:

Mogztter:
Thanks @robertpanzer 
I'm back!

> Can you tell where you introduced changes to the API?
Clearly it breaks binary compatibility due to the renaming of Processor to BaseProcessor, but are there also changes that would require changes other than recompiling?

Reading the code again I don't think it breaks the API anymore.

robertpanzer:
Looks great! 👍 
I'll try to create an RC2 tomorrow.

mojavelinux:
Nice (team)work!

asciidoctorj-api/src/main/java/org/asciidoctor/extension/BaseProcessor.java
Mogztter: @robertpanzer We still need to use a proper resolution mechanism here 🏷 
As long as someone does not provide another implementation we should be fine 😓

Mogztter:
@robertpanzer Awesome, next step is to move the `Asciidoctor` interface 😅

robertpanzer:
>  Awesome, next step is to move the Asciidoctor interface 😅

We can go for that :)
Only if we want to remove the factory methods that take a gem path at the same time I think we should wait for 2.0.0

Mogztter:
Yes you're right.
I thought about it and maybe this API can stay. I mean it may be useful to load from a given path in other implementations.

But at the same time it may be too generic and the users might get confused by what the factory actually does with this path.

I still feel like "advanced initialization" strongly depends on the context (and thus on the mplementation).

PR #717, opened 2018-11-06
Replace deprecated constants and methods in the extension package

Mogztter:
- @deprecated BlockProcessor context constants (replaced by Contexts)
- should we remove Processor constants ?
- Use diamond where possible `Map<String, String> config = new HashMap<>()`
- Reformat code (tabs -> spaces)

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/extension/Contexts.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/extension/Contexts.java
@@ -49,6 +49,23 @@ import java.lang.annotation.Target;
 public @interface Contexts {
 
     /**
+     * This value is used as the config option key when defining the block type
+     * a Processor should process.
+     * Its value must be a list of String constants:
+     *
+     * <p>Example to make a BlockProcessor work on listings and examples named foo:
+     * <pre>
+     * <verbatim>
+     * Map&lt;String, Object&gt; config = new HashMap&lt;&gt;();
+     * config.put(Contexts.KEY, Arrays.asList(Contexts.EXAMPLE, Contexts.LISTING));
+     * BlockProcessor blockProcessor = new BlockProcessor("foo", config);
+     * asciidoctor.javaExtensionRegistry().block(blockProcessor);
+     * </verbatim>
+     * </pre>
+     * </p>
+     */
+    String KEY = "contexts";
+    /**
      * Predefined constant for making a Processor work on open blocks.
      * <pre>
      * [foo]
@@ -58,7 +75,7 @@ public @interface Contexts {
      * --
      * </pre>
      */
-    public static final String OPEN = ":open";
+    String OPEN = ":open";
 
     /**
      * Predefined constant for making a Processor work on example blocks.
@@ -69,7 +86,7 @@ public @interface Contexts {
      * ====
      * </pre>
      */
-    public static final String EXAMPLE = ":example";
+    String EXAMPLE = ":example";
 
     /**
      * Predefined constant for making a Processor work on sidebar blocks.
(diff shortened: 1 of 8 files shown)


asciidoctorj-core/src/main/java/org/asciidoctor/extension/Contexts.java
```diff
@@ -1,10 +1,6 @@
 package org.asciidoctor.extension;
 
-import java.lang.annotation.Documented;
-import java.lang.annotation.ElementType;
-import java.lang.annotation.Retention;
-import java.lang.annotation.RetentionPolicy;
-import java.lang.annotation.Target;
+import java.lang.annotation.*;
```
robertpanzer: Nitpick: I generally never use star imports. I'm fine with it, but want to make sure this didn't slip in accidentally, because the default of IntelliJ seems to use them starting with 4 imported elements.

Mogztter: Indeed it's the default behavior of IntelliJ but I can revert this change :wink:

robertpanzer: I'm fine with it. :) I just wanted to be certain that we want this.
If you think it makes sense to have star imports I'll merge.

(Of course I have to admit that there are parts of the code where the formatting is not glorious 😄)

Mogztter: I can find it anymore but Mark Reinhold tweeted something like "you should use this language feature" about star imports but at the same time I think it's better to be explicit.
I don't have a strong opinion so explicit imports is fine (and it also consistent with the codebase).

Mogztter:
@robertpanzer  should we remove `Processor` constants as they were already marked as @deprecated (in 2016 !)

I think we should write a rule to define when `@deprecated` features will be removed (or add an explicit notice on the annotation, "will be removed in 2.0.0").
For instance in Scala:

> A deprecated element of the Scala language or a definition in the Scala standard library will be preserved at least for the current major version.
This means that an element deprecated in some 2.12.x release will be preserved in all 2.12.x releases, but may be removed in 2.13. (A deprecated element might be kept longer to ease migration. Developers should not rely on this.)

https://www.scala-lang.org/api/current/scala/deprecated.html

We could also add a `@since` so we know that this feature has been `deprecated` since version X (and since we are version Y it's pretty safe to remove it).

robertpanzer:
I would vote for removing it with 2.0.0.
There should be some users that are already on 1.6.0 for a longer time and we shouldn't break them like that.

Scala is a special case, it is just normal that it is incompatible with dot-releases, and therefore you see libraries available for Scala 2.11 and 2.12. So much for semantic versioning 😂

Mogztter:
> Scala is a special case, it is just normal that it is incompatible with dot-releases, and therefore you see libraries available for Scala 2.11 and 2.12. So much for semantic versioning

My point was to define a rule regarding `deprecated` elementa (we certainly don't want to use the same versioning strategy as Scala).
I think it's safe to use a rule where `deprecated` elements in 2.0.0 will be removed in 3.0.0. We may also noted that a deprecated element might be kept longer to ease migration but that developers should not rely on this. That way AsciidoctorJ users know that they need to remove deprecated code before upgrading to a major version.

robertpanzer:
Sorry if the comment about Scala was too harsh :)

Yes, that makes sense.
Removing deprecated elements in the next major release should be fine, so that we can remove these with 2.0.0.

mojavelinux:
> My point was to define a rule regarding deprecated elements

These are the kinds of rules we're trying to establish right now to make it easier to manage the projects and so users know what to expect. I recommend starting by adding them to a version policy section in the README. We can then go through the repositories and collect these up to display somewhere on asciidoctor.org.

PR #720, opened 2018-11-07
Use diamond, remove unnecessary public keyword on interfaces

Mogztter:
A little code cleanup in the extension package (to reduce the code changes in #715).

If you want to do a single commit to use diamond operator everywhere I can do it. I can also remove the unnecessary `public` keyword from interfaces at once ?


robertpanzer:
Thanks!
We can also iterate and update the code in steps.
At the time I added the public modifier to the interface methods because there's also the private modifier now.

PR #721, opened 2018-11-07
OSGi snippets update

twasyl:
Since version `1.5.7` of asciidoctorj, the given snippets for using it in an OSGi didn't work anymore as the JRuby dependency has changed to version 9.x. This make the use of the RubyInstanceConfig useless and not working.

Moreover, the load path passed to the `JavaEmbedUtils#initialize` should be prefix with `uri:classloader`. I didn't find another solution for make it work. The solution has been integrated in [SlideshowFX](https://github.com/twasyl/SlideshowFX/blob/dae814bfbc6bccbf09033adc8223d97ab41d18d3/slideshowfx-asciidoctor/src/main/java/com/twasyl/slideshowfx/markup/asciidoctor/AsciidoctorMarkup.java#L28) using asciidoctorj 1.5.8.1.

Also update the artifact version in the documentation.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -20,7 +20,7 @@ ifndef::awestruct[:uri-docs: http://asciidoctor.org/docs]
 :uri-repo: https://github.com/asciidoctor/asciidoctorj
 :uri-issues: {uri-repo}/issues
 :uri-discuss: http://discuss.asciidoctor.org
-:artifact-version: 1.5.7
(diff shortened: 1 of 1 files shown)


robertpanzer:
Thanks for the update!
Also that helped to refresh my mind why the create method with a ClassLoader is required.

twasyl:
You're welcome!

PR #722, opened 2018-11-09
Bump JRuby to 9.2.2.0

robertpanzer:
This PR bumps JRuby to the latest version for the 1.5.x branch.
We might not need it anymore, but if we come into the situation that we do another release we should have the latest version of JRuby.


robertpanzer:
Close this for now.
Anybody running AsciidoctorJ 1.5.x with Java 8 should be able to use JRuby 9.2.x.0.
I don't want to create a release for this anymore because.

PR #723, opened 2018-11-09
Bump JRuby to 9.2.2.0

robertpanzer:
This PR bumps JRuby to the latest version on master.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -44,7 +44,7 @@ ext {
   guavaVersion = '18.0'
   hamcrestVersion = '1.3'
   jcommanderVersion = '1.35'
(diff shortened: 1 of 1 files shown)


mojavelinux:
I just read a post that 9.2.3.0 was released to fix a bug with blocks. We
should probably use that version instead.

mojavelinux:
Or I just got lucky. But since it looked like it fixed a rather major problem, I wanted to point it out asap.

robertpanzer:
Yeah, definitely. Thanks for the heads up! I thought 9.2.2.0 was already a hotfix for a bug in 9.2.1.0. looks like the JRuby folks can do Hot-Hot Fixes :)

PR #724, opened 2018-11-09
Revert #712

robertpanzer:
This PR reverts #712 again.
Apparently (see #721) the create methods that take a Classloader are required to use AsciidoctorJ in OSGi environments.
Until we can offer alternatives we should therefore keep these methods.


mojavelinux:
Perhaps mention this in the API doc so we don't forget?

robertpanzer:
That's a good point!

PR #732, opened 2018-11-23
Move Asciidoctor to the API

Mogztter:
* Remove deprecated `render*` methods
* Remove `rubyExtensionRegistry` from the API (bounded to the JRuby implementation)
* Only keep `Asciidoctor.Factory.create()` method to instantiate the processor
  * Other methods are tightly coupled with the context/implementation but it's also a limitation of the current implementation with `ServiceLoader`. In fact `ServiceLoader` will call the default constructor (that's why I've added a public no-arg constructor on `JRubyAsciidoctor`).

I've kept `requireLibrary` and `requireLibraries` methods because I think they can be useful in other implementations.


robertpanzer:
This was already merged as part of #760

PR #733, opened 2018-11-24
Resolves #487, Use the fluent API in test

### LINKED ISSUE #487: Chain commands in JavaExtensionRegistry
It would be useful if the methods in the JavaExtensionRegistry supported method chaining. This would allow extensions to be registered in a single call and avoid having to assign the registry to a local variable.

``` java
asciidoctor.javaExtensionRegistry()
    .block(TabContainerBlock.class);
    .block(TabBlock.class);
```

The RubyExtensionRegistry already has this feature, which even further justifies why we should do it.

Mogztter:


robertpanzer:
Thanks!

PR #735, opened 2018-11-25
Run ci build with Java 11

robertpanzer:
I successfully built AsciidoctorJ locally with OpenJDK 11.
Let's cross fingers that it works on the ci systems as well.
There's probably not much value anymore running tests with Java9 which is already eol.


mojavelinux:
Nice work. And agree, it should be 8 and 11 only.

PR #740, opened 2018-12-01
update saxon dep to last 9.9.0-2

innovimax:


innovimax:
Ok it's used in ascii doctor core
Upgrade it then

robertpanzer:
Thanks! 
It's just test dependency, but still good to update them.

PR #746, opened 2018-12-27
make asciidoctorj compatible with java 11 module system

aanno:
I played around with asciidoctorj a bit. It's a great project but at present it is very painful to use it from the java 11 module system. This is because both `asciidoctorj.jar` and `asciidoctorj-api.jar` both have (java class) code in the `org.asciidoctor` package (and its sub packages).

The new java 11 (or java 9) module system does not like this kind of 'split jars'. The agreed-upon long-term solution to split jars is to separate the package namespaces. This is what I did. In detail:

* I moved all the java classes in `asciidoctorj-api.jar` to the `org.asciidoctor.api` package (and sub packages)
* I fixed the ruby java interaction in `asciidoctorclass.rb`
* I fixed all tests

The pull request is - for most parts - lengthy but boring. I would be glad it you could consider integration into branch master.

Kind regards,

aanno


aanno:
I guess a unsuccessful travis-ci check with 'We couldn't find the repository
asciidoctor/asciidoctorj' does not qualify as blocker...

robertpanzer:
Hi,

thanks for this awesome contribution, I didn't find time yet myself to try this out as a module.
I wonder though if JRuby already works as a module, it would be great if you could share your experience.

We are currently shortly before doing the next major release, 1.6.0, so I wouldn't want to bring in such a major change.
Regarding the moving of the classes we rather want to move everything from `org.asciidoctor.ast` to the api module, the implementation parts should go into `org.asciidoctor.ast.impl` or `org.asciidoctor.ast.internal`.
For most classes this should already be the case.

Also, eventually `org.asciidoctor.Asciidoctor` should go into the api module and all references to an implementation should be resolved via a ServiceLoader.

aanno:
1.6.0 is out now. For moving `org.asciidoctor.Asciidoctor` to API there is PR #732. So the thing that needs to be done is moving the `ast` stuff. Right?

robertpanzer:
The work on #732 has continued in https://github.com/asciidoctor/asciidoctorj/pull/760.
There we will discuss how we want to change the API to facilitate the move to the JPMS.
It would be great if you would join the discussion there.

robertpanzer:
Thanks for your contribution!
But the current master should run from the module path, there should not be any overlapping modules anymore.
Therefore I'm closing this one, given that we achieved JPMS support with #760 after a longer discussion with the community.

PR #756, opened 2019-01-13
Fix upstream build by upgrading jruby-maven-plugin to 1.1.6

robertpanzer:


robertpanzer:
Strange that it failed again.
It passed locally for me with Java 11 if I haven't done an error.

mojavelinux:
Can you explain why you expect this change to fix the build? I'm curious.

robertpanzer:
I created this issue (
https://github.com/torquebox/jruby-maven-plugins/issues/92) and saw that after that 1.1.6 was released. 
Therefore I thought it might have fixed this problem. 

And it fixed sth as some warnings are no longer in the Pom, but some messages from the JVM still sneaked in on Travis. 
Locally it works now for me, but apparently not on Travis.

robertpanzer:
Maybe I have to patch the Pom between the Maven and the Gradle build.

PR #757, opened 2019-01-13
Fixes #754. Initialize BlockProcessor.name correctly

### LINKED ISSUE #754: AsciidoctorJ fails when registering a block processor instance while passing the name as a parameter
To reproduce:
Register a Block processor passing the name as an argument, i.e. by calling this method:

```java
javaExtensionRegistry.block("yell", new MyBlockProcessor());
```

Rendering a document fails because Asciidoctor calls the `name` accessor of the extension when classes are registered.
This error affects AsciidoctorJ 1.6.0 only as previously extension instances were registered to Asciidoctor as instances.
This changed with 1.6.0 as now instances are wrapped in a Ruby class per registration.

robertpanzer:
This PR fixes #754.

Before merging this I will also have to fix the same problem for macros, as they like have it as well.


robertpanzer:
If there are no objections, I'd like to release a 1.6.1 with this fix.

PR #760, opened 2019-01-26
Move Asciidoctor to the API module ++

robertpanzer:
This PR builds on @Mogztter's PR #732.
I rebased it on top of current master.

Additionally I moved some other files to packages to avoid package conflicts between api and core.

There are still some open classes that I would like to discuss how to handle them:

- `org.asciidoctor.RubyExtensionRegistry`: Should we also move it to the API? Or into another package?
- `org.asciidoctor.Abstract|Asciidoc|GlobDirectoryWalker`: I don't really see these classes in the API. Should we move it to another package? Should we completely get rid of `org.asciidoctor.DirectoryWalker` in the API? In the end this is just a `Producer<File>`


Mogztter:
> org.asciidoctor.RubyExtensionRegistry: Should we also move it to the API? Or into another package?

In my opinion no because as the name suggests then the implementation must use Ruby. I think that every implementation should only provide a way to register Java extensions.
In a JRuby implementation, it can make sense to register Ruby extensions but not necessarily on other implementations. For instance in AsciidoctorG, I will not be able to implement this interface because I don't have a Ruby runtime but instead it could make sense to write extensions as JavaScript... So that's why I think it should stay in the implementation.

> org.asciidoctor.Abstract|Asciidoc|GlobDirectoryWalker: I don't really see these classes in the API. Should we move it to another package? Should we completely get rid of org.asciidoctor.DirectoryWalker in the API? In the end this is just a Producer<File>

Indeed I don't really see these classes in the API neither.

robertpanzer:
> In my opinion no because as the name suggests then the implementation must use Ruby. I think that every implementation should only provide a way to register Java extensions.

Makes sense. 
So we're not going with a way where we would have methods that would throw a UnsupportedOperationException for certain operations, rather asciidoctorj-api will only provide the minimal set of operations.

That also means that I must move the interface to another package so that we can end up with disjoint packages between the modules.

As we also need an additional factory for AsciidoctorJ instances – that take additional options like a gem path – should we use the package `org.asciidoctor.asciidoctorj` then?
And then probably also `org.asciidoctor.asciidoctorg` to stay coherent?

robertpanzer:
I updated the asciidoctorj-core module by moving all classes below `org.asciidoctor.asciidoctorj`. 
That way we have a clearer separation between the api and implementation modules without having to look at `impl` or `internal` parts in the package name.

robertpanzer:
That means the `org.asciidoctor.RubyExtensionRegistry` moves to `org.asciidoctor.asciidoctorj.RubyExtensionRegistry` now.
This brings at least a binary incompatibility with the Maven plugin.

As the next step I want to add a factory for AsciidoctorJ instances that takes the JRuby specific parameters.

mojavelinux:
It's just my first impression, but `org.asciidoctor.asciidoctorj` looks a bit strange to me. That might just be because it's new. But would `org.asciidoctor.ruby` make more sense? It's not so much about the implementation but what is available underneath.

robertpanzer:
Another change we might want to thing about is the move of `org.asciidoctor.extension.spi.ExtensionRegistry` and `org.asciioctor.converter.spi.ConverterRegistry` to the core module.
Moving these interfaces requires the user to also rename their services files, not just changing their Java code which would be immediately complained about by the compiler.
This could result in a lot of questions why an extension is suddenly not executed anymore.

robertpanzer:
If we keep them in the implementation module, the interface should probably also change to get an AsciidoctorJ instance, instead of the base Asciidoctor interface.

I.e.
```java
package org.asciidoctor.ruby.extension.spi;

public interface ExtensionRegistry {
    void register(AsciidoctorJ asciidoctor); // receives an instance of AsciidoctorJ so that it can immediately register Ruby extensions without having to unwrap.
}
```

robertpanzer:
Ruby extensions also leak through in the ExtensionGroup API, we also need to separate this too then.

```java
public interface ExtensionGroup {
...
  ExtensionGroup requireRubyLibrary(String requiredLibrary);
  ExtensionGroup loadRubyClass(InputStream rubyClassStream);

  ExtensionGroup rubyPreprocessor(String preprocessor);
  ExtensionGroup rubyPostprocessor(String postprocessor);
  ExtensionGroup rubyDocinfoProcessor(String docinfoProcessor);
...
}
```

robertpanzer:
Moved all content of asciidoctorj-core to org.asciidoctor.ruby as proposed.

Additionally I created an interface `AsciidoctorJ` and a corresponding factory `AsciidoctorJ.Factory`.
For me this is a necessary requirement that we can still support everything that we did before (i.e. define gem and load path etc).

robertpanzer:
With the current state I got AsciidoctorJ working from the module path on Java 11.
I called asciidoctorj like this:
```
java -p lib --add-modules=jdk.unsupported,java.scripting -m asciidoctorj/org.asciidoctor.ruby.cli.AsciidoctorInvoker -r asciidoctor-diagram -b pdf test.adoc
```

Additionally I had to remove the `module-info.class` from the jruby-complete.jar (see https://github.com/jruby/jruby/issues/5507)
Using the dependency org.jruby:jruby did not work. Java complained about overlapping packages.

mojavelinux:
> I wonder though if there would also be a chance for a TruffleRuby impl.

If that case arose, then we'd use a package name like truffleruby. What we'd be doing then is using the engine name as identified by RUBY_ENGINE. In that case, I wouldn't be opposed to naming the package jruby instead of ruby.

robertpanzer:
I am still undecided about the package name.
The fact that AsciidoctorJ uses JRuby under the hood should be just an implementation detail that we try to hide as good as possible.
We support ruby extensions on top of Java, but I could imagine to add support for other languages as well that don't map naturally or elegantly to a Java API (e.g. JavaScript, Frege, ...)

robertpanzer:
Just to list the changes how one would interact with AsciidoctorJ:
Somebody who uses the original Asciidoctor interface would have to do the following to register a Ruby extension:
```java
Asciidoctor asciidoctor = org.asciidoctor.Asciidoctor.Factory.create()<
asciidoctor.unwrap(org.asciidoctor.ruby.AsciidoctorJ.class) // <1>
   .rubyExtensionRegistry().register...
```
<1> Current state is to have asciidoctorj-core in the package org.asciidoctor.ruby.

Somebody who wants to create the AsciidoctorJ instance with a certain gem path or load path would have to use the new AsciidoctorJ Factory:
```java
org.asciidoctor.ruby.AsciidoctorJ asciidoctorj = 
    org.asciidoctor.ruby.AsciidoctorJ.Factory.create(gempath);
asciidoctorj.rubyExtensionRegistry().register...
```

abelsromero:
> Somebody who uses the original Asciidoctor interface would have to do the following to register a Ruby extension:

Would it make sense to have a Extension interface with different implementations depending on the implementation?
Each implementation would require a factory that would be located in its respective package.

robertpanzer:
Ok, I am open for anything :-D

So you're also proposing
```java
org.asciidoctor.ruby.AsciidoctorJRuby asciidoctor = org.asciidoctor.ruby.AsciidoctorJRuby.Factory.create(gemPath);
asciidoctor.rubyExtensionRegistry()...
```

Personally I don't have any preference, the only thing that I definitely want is to not remove any features that we need, e.g. passing in a gem path or classloader when creating an Asciidoctor instance.

@abelsromero What was your proposal exactly?
The Asciidoctor would return a common ExtensionRegistry which would then in turn be unwrapped to a JavaExtensionRegistry, RubyExtensionRegistry, JavascriptExtensionRegistry etc?
I.e. like this?
```
var asciidoctor = Asciidoctor.create();
ExtensionRegistry extensionRegistry = asciidoctor.extensionRegistry();
extensionRegistry.unwrap(RubyExtensionRegistry.class).rubyBlock(...);
extensionRegistry.unwrap(JavaExtensionRegistry.class).treeprocessor(...);
```

abelsromero:
> The Asciidoctor would return a common ExtensionRegistry which would then in turn be unwrapped to a JavaExtensionRegistry, RubyExtensionRegistry, JavascriptExtensionRegistry etc?

Trying to minimize usage of unwrap and making sure implementation details are hidden. From an integrator point of view the current approach requires that (1) extension location (searching Ruby file in path, instantiating Java class, etc.) and (2) registration to be aware of the implementation. Both steps could happen in different points of an application.

"As an integrator" I am aware of the source of the extension (step 1) because what extensions I use is a design desition. But I don't see the point on having to be aware of the implementation to register it (step 2). "As an integrator" I'd would like to be able to register an extension to Asciidoctor* and let it take care of the details without being aware of implementations.

robertpanzer:
Thanks for your input, @abelsromero.

So before proceeding with further changes I'd like to start a discussion on how the API should look like.
Factors that have to considered are:
- Different implementations might require different factory methods.
- Currently different implementations of Asciidoctor provide a single interface across multiple implementations if we ignore the ExtensionRegistries.
- Registering an Extension should be independent of the implementation itself

So one proposal would be the following:

```java
package org.asciidoctor;
public interface Asciidoctor {
    public static class Factory {
        public static Asciidoctor create() {...}
    }

    String convert(...);

    ExtensionRegistry extensionRegistry();
    
    public interface ExtensionRegistry {
        public <T> T unwrap(Class<T> clz);
    }
}

package org.asciidoctor.jruby;
public interface AsciidoctorJRuby { // it does not extend Asciidoctor!
    public static class Factory {
        public Asciidoctor create(String gemPath) {
            Asciidoctor asciidoctor = ... // create JRubyAsciidoctor instance
            return asciidoctor;
        }
    }
}

public class ExtensionRegistryAsciidoctorJRuby implements ExtensionRegistry {
    public <T> T unwrap(Class<T> clz) {
        if (RubyExtensionRegistry.class.isAssignableFrom(clz)) {
           // return my RubyExtensionRegistry
        } else if ...
    }
}
```

Please excuse any errors in the Java code, I haven't done much Java in the last months.

Another thing that we should overthink if we want to decouple the implementation from the extension type, then the RubyExtensionRegistry should probably go into asciidoctorj-api, as a Truffle implementation should use the same interface.
The JavaExtensionRegistry would become an equal citizen next to the RubyExtensionRegistry that is also optional. A Truffle implementation might initially not be able to support Java extensions, but it would support Ruby extensions.

Looking forward to feedback.

Mogztter:
Overall, I like it but it feels a bit weird that the `ExtensionRegistry` only provides an `unwrap` method.

> A Truffle implementation might initially not be able to support Java extensions, but it would support Ruby extensions.

That's also the case with the GraalVM JavaScript implementation where I can support JavaScript extensions but not Java extensions (at least for now).

So I guess it's the right choice as it might be too much constraint to make the JavaExtensionRegistry API mandatory.

robertpanzer:
Nice! I see where you want to go.

I just wonder if it wouldn't be a problem allowing to register any implementation of AsciidoctorExtension?
If somebody sees this API, would she or he try to implement AsciidoctorExtension her/hisself without using BlockProcessors, Treeprocessor etc?
Also the macros and BlockProcessors allow registration with an explicit name, how can we restrict the API so that only these can be registered with a name.

And if we want to support Lambdas as extensions – which I would really like to have because it brings Java closer to the Groovy extensions – how would the compiler be able to infer the type?
(For this we also still need to define functional interfaces)

robertpanzer:
I just tried implementing my own proposal, but it still feels awkward, and I have no idea yet how to implement it without always requiring `unwrap()` or having the superset in the API and throwing exceptions. (`unwrap()` would also throw exceptions so it seems to fail on both sides here)

abelsromero:
Tbh, I am leaning towards the `unwrap` solution after this exercise. Comparing JavaExtentionRegistry against RubyExtentionRegistry, I can't see how to find a common denominator that works, and even more, can work for future implementations like graal or js. Only thing, would be imposing the JavaRegistry interface with all extension interfaces, but that's useless in most cases as @Mogztter already pointed out.

Unwrap allows having specific interfaces adapted to the specific needs of every implementation. You could say that once you unwrap is more "type safe".

On a side note, I have recently dealt with some APIs that offer several implementations for a feature and to do that, they go to too many hops and abstractions to the point the code and runtime are way too complex. I wouldn't like AsciidoctorJ to end like that, I'd rather have some "awkwardness" and implementation leaking from time to time.

Mogztter:
Maybe we don't have enough hindsight. I mean it's basically a plug-in interface where we don't know exactly the use cases. Well we know one use case but we don't really know the requirements for the other implementations. 

While Abel proposal looks good I think it's important to remember that we can always add but we can "never" remove. So I think we should either remove the `ExtensionRegistry` from the API or provide a really basic `ExtensionRegistry`.

> I just tried implementing my own proposal, but it still feels awkward, and I have no idea yet how to implement it without always requiring `unwrap()` or having the superset in the API and throwing exceptions. (`unwrap()` would also throw exceptions so it seems to fail on both sides here)

In practice you can only `unwrap` for a known implementation and you will likely only have one implementation in your classpath right ? So `unwrap` should not throw exceptions. But again if it feels awkward maybe we should just leave it out until we have 3 or more implementations. Then we can start designing an API that makes sense based on real use cases and requirements.

robertpanzer:
Thinking about this, proposing `unwrap` to the user as the primary way to register extension doesn't feel right to me.
Until now we use `unwrap()` only internally to link things together and the user could only use it to do some advance stuff.
But proposing this publicly to the user as the only way to register extensions doesn't seem right.
And registering extensions is imo literally the second thing a user will do after rendering a document.
The developer experience with this is in my opinion is also rather bad, because the IDE will not guide you anymore.

Actually, for now, I would prefer to continue offering the current methods `rubyExtensionRegistry()` and `javaExtensionRegistry()` directly by the `AsciidoctorJRuby` interface.
Starting from the `Asciidoctor` interface we can surely add `unwrap()` to allow registering extensions, but imo it should stay as simple as it is today if you already have the right thing.

robertpanzer:
I moved the implementation to org.asciidoctor.jruby now.

I left the methods `Asciidoctor.javaExtensionRegistry()` and `Asciidoctor.rubyExtensionRegistry()` now where they are.
I think we have to lead this discussion outside of this PR to go forward as it seems to me that we have not found a common decision yet.

Mogztter:
>  I think we have to lead this discussion outside of this PR to go forward as it seems to me that we have not found a common decision yet.

I agree

asciidoctorj-api/src/main/java/org/asciidoctor/log/Severity.java
```diff
@@ -1,10 +1,20 @@
...
-    FATAL,
-    UNKNOWN
+    DEBUG(0),
+    INFO(1),
+    WARN(2),
+    ERROR(3),
+    FATAL(4),
+    UNKNOWN(5);
+
+    private int rubyId;
```
robertpanzer: Oops, this must go away.
I thought I removed the rubyIds, somehow they sneaked in again.

asciidoctorj-core/src/main/java/org/asciidoctor/jruby/cli/AsciidoctorInvoker.java
```diff
@@ -38,7 +36,7 @@ public void invoke(String... parameters) {
             JRubyAsciidoctor asciidoctor = buildAsciidoctorJInstance(asciidoctorCliOptions);
             
             if (asciidoctorCliOptions.isVersion()) {
-                System.out.println("AsciidoctorJ " + asciidoctor.asciidoctorVersion() + " [http://asciidoctor.org]");
+                System.out.println("AsciidoctorJRuby " + asciidoctor.asciidoctorVersion() + " [http://asciidoctor.org]");
```
robertpanzer: This comes from the find-replace, but opens an interesting question: Should we actually rename the command line tool as well to `asciidoctorjruby`?

Mogztter: Indeed it's a good question but I think we should stick with `asciidoctorj` at least for now.

Mogztter:
The build is failing on Travis but can we review the code ? (ie. is this pull request ready ?)

robertpanzer:
Yes, this can be reviewed. 
The upstream build on Java 11 always fails due to the error in the jruby-maven-plugin.
And it would now fail as well due to the changes in the Converter registry that I wasn’t able to adapt to yet.

Mogztter:
:rocket: :rocket: :rocket:

PR #765, opened 2019-02-02
OSGI update

twasyl:
- Update OSGI snippet in both README and JavaDoc because the behaviour has changed in 1.6.0
- Update artifact-version to 1.6.1

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -940,9 +940,10 @@ In an OSGi context it will not work because JRuby needs some paths to find the A
 import static org.asciidoctor.Asciidoctor.Factory.create;
 import org.asciidoctor.Asciidoctor;
 
-Asciidoctor asciidoctor = create(Arrays.asList("uri:classloader:/gems/asciidoctor-1.5.8/lib")); <1>
+Asciidoctor asciidoctor = create(Arrays.asList("uri:classloader:/gems/asciidoctor-1.5.8/lib")); <1><2>
 ----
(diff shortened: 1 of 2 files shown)


README.adoc
```diff
@@ -940,13 +940,9 @@ In an OSGi context it will not work because JRuby needs some paths to find the A
 import static org.asciidoctor.Asciidoctor.Factory.create;
 import org.asciidoctor.Asciidoctor;
 
-JavaEmbedUtils.initialize(Arrays.asList("uri:classloader:/gems/asciidoctor-1.5.8/lib")); <1><2>
-
-Asciidoctor asciidoctor = create(this.getClass().getClassLoader()); <3>
+Asciidoctor asciidoctor = create(Arrays.asList("uri:classloader:/gems/asciidoctor-1.5.8/lib")); <1>
```
robertpanzer: Should this be `asciidoctor-{artifact-version}` instead of `asciidoctor-1.5.8`?

twasyl: Well, the gem within the final Jar is still named `asciidoctor-1.5.8`. I don’t know if it’s a typo but this is working in [SlideshowFX](https://github.com/twasyl/SlideshowFX/blob/master/slideshowfx-asciidoctor/src/main/java/com/twasyl/slideshowfx/markup/asciidoctor/AsciidoctorMarkup.java)

robertpanzer: Ah right :) Sorry for this.
Yes, here the different versioning shows up.
Maybe others might be confused too, like me and a comment highlighting this would be great.
Just saying that this has to use the version of the packaged version of Asciidoctor, not AsciidoctorJ.

twasyl: No problem at all :)
I just pushed a clarification within the README for the version number. Feel free to comment the modification :)

PR #769, opened 2019-02-08
add asciidoctorj-diagram to list of artifacts

jeffcjohnson:


mojavelinux:
Thanks for the suggestion. I regret to say, however, that this is opposite of the direction we are trying to head. This strengthens the coupling between AsciidoctorJ and AsciidoctorJ Diagram, which we're trying to break. Instead, the artifact listing should go in the repository from which it's produced, such as: https://github.com/asciidoctor/asciidoctorj-diagram

Having said that, I do see value in having a landing page that lists the versions of all the AsciidoctorJ-related jars. Though perhaps it's something we can compute at the time we build out the (new) docs. It's a good requirement.

jeffcjohnson:
Hi Dan, thanks for information.  As an newcomer to these projects it wasn't clear to me that this artifact wasn't generated here.  I just assumed it was because the other two were listed here.

mojavelinux:
I wouldn't expect you to know that. You've arrived at a transition point. Although the references are still lingering, they are on the way out the door (to their own repositories).

PR #775, opened 2019-02-19
Fix upstream

robertpanzer:
This PR adapts AsciidoctorJ to the latest changes in Asciidoctor upstream.
Besides changes to the converter registration, also the structure of DefinitionLists in the AST has changed.

This PR must not be merged prior to Asciidoctor 2.0.0 as the changes are incompatible with the current version of Asciidoctor, 1.5.8.

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/converter/internal/ConverterProxy.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/converter/internal/ConverterProxy.java
@@ -45,6 +45,7 @@ public class ConverterProxy<T> extends RubyObject {
                 rubyRuntime.getObject(),
                 new ConverterProxy.Allocator(converterClass));
         includeModule(clazz, "Asciidoctor", "Converter");
+        includeModule(clazz, "Asciidoctor", "Converter", "BackendTraits");
 
         clazz.defineAnnotatedMethod(ConverterProxy.class, "initialize");
         clazz.defineAnnotatedMethod(ConverterProxy.class, "convert");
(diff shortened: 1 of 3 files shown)


mojavelinux:
Excellent.

> also the structure of DefinitionLists in the AST has changed.

This is unexpected. I was not aware that the AST changed for dlists. Can you explain what you think changed in the underlying AST? I can look into it to see what happened.

mojavelinux:
What I can tell you is that this no longer produces a dlist.

```
::
```

Perhaps in the test you have something like that and are creating a dlist node but the parser isn't giving you any entries for it?

robertpanzer:
The upstream build failed for this document:
https://github.com/asciidoctor/asciidoctorj/blob/cf34fb80f66e1afee0b7935657ad447ec6c4c3ed/asciidoctorj-core/src/test/groovy/org/asciidoctor/WhenADocumentContainsADefinitionList.groovy#L108-L111

It looks like previously a Description list item consisted of an array containing all terms and always a second element which was null or a ListItem.
Now it seems like the array has only one element if the definition is missing.
The structure of the first element, an array containing all terms as ListItems, has not changed.

mojavelinux:
Ah, that makes sense. But we're still missing a test for that assertion in
core. I'll add one and document what the official behavior will be.

mojavelinux:
I fixed this problem upstream by restoring the previous behavior. Here's the rule:

> In a description list (dlist), each item is a tuple that consists of a 2-item Array of ListItem terms and a ListItem description (i.e., [[term, term, ...], desc]. If a description is not set, then the second entry in the tuple is nil.

Nice catch!

robertpanzer:
Awesome, thanks!
I checked that the upstream build works again without the changes to NodeConverter.
I removed the corresponding commit form this PR.

robertpanzer:
Closing as this PR is also part of #783

PR #777, opened 2019-02-25
fix jruby-maven-plugin and build of gem

mkristian:


mkristian:
@robertpanzer I released the maven plugin and adjusted the poms here. let me know if you need more ;)

robertpanzer:
Thank you @mkristian !
This is awesome! 🎉

PR #783, opened 2019-03-23
Asciidoctor 2.0.0

robertpanzer:
This PR upgrades Asciidoctor core to 2.0.0.
I also upgraded a few of the other gems after looking at the dependencies of Asciidoctor and what are the most current compatible versions.

For a release a new version of asciidoctor-diagram and asciidoctorj-diagram is required.

Once we have that I would like to release an AsciidoctorJ 1.7.0.
(I don't want to release a 2.0.0 yet given that there is still so much disagreement about how the API should look like in the future)

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/extension/processorproxies/AbstractProcessorProxy.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/extension/processorproxies/AbstractProcessorProxy.java
@@ -217,7 +217,7 @@ public class AbstractProcessorProxy<T extends Processor> extends RubyObject {
                 positionalAttrs.add(positionalAttribute);
             }
             rubyClass.callMethod(rubyRuntime.getCurrentContext(), "option", new IRubyObject[]{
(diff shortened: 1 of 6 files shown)


asciidoctorj-core/src/main/java/org/asciidoctor/jruby/extension/processorproxies/AbstractProcessorProxy.java
```diff
@@ -217,7 +217,7 @@ private static void handlePositionalAttributesAnnotation(Class<? extends Process
                 positionalAttrs.add(positionalAttribute);
             }
             rubyClass.callMethod(rubyRuntime.getCurrentContext(), "option", new IRubyObject[]{
-                    rubyRuntime.newSymbol("pos_attrs"),
+                    rubyRuntime.newSymbol("positional_attrs"),
```
mojavelinux: I'm going to go ahead a remap this to avoid breaking existing extensions. But since AsciidoctorJ bundles a specific version of Asciidoctor core, you can just fix it directly here.

mojavelinux:
:tada:

PR #786, opened 2019-04-02
Fix logger and upgrade to Asciidoctor 2.0.5

robertpanzer:
This PR implements the methods for the logger stub that were missing from its Ruby counterpart but which are called by Asciidoctor.
That way logging of invalid refs should work now without having to enable the $VERBOSE variable.
This change also appeared in https://github.com/asciidoctor/asciidoctor-maven-plugin/pull/389

Additionally this PR upgrades Asciidoctor to 2.0.5 plus it should hopefully publish snapshot builds of master.


mojavelinux:
> logging of invalid refs should work now without having to enable the $VERBOSE variable.

:tada:

PR #794, opened 2019-04-12
Logs handling API - add javadoc

jmini:
I was creating my own logger, and I would have appreciated to find some Javadoc on the getters.
The text I wrote is based on what can be found in the README.adoc at the root of the repository.


abelsromero:
Not sure "rending" is the term we want to use 😆 The API moved from "rendering" to "converting", I think it would be better.

jmini:
I have changed "render" into "convert" in the javadoc and in the readme (only the LogHandler section)

robertpanzer:
Thanks! 🎉

PR #795, opened 2019-04-12
Fixes #791. Print version of AsciidoctorJ when calling asciidoctorj -…

### LINKED ISSUE #791: CLI reports asciidoctor version instead of Asciidoctorj
After installing 1.6.0, 1.6.1 and 1.6.2 versions of Asciidoctorj with sdkman I noticed they all report v 1.5.8 like this when using `-V` argument:
```
AsciidoctorJ 1.5.8 [http://asciidoctor.org]
Runtime Environment: jruby x.x.x.x (x.x.x)
```
I checked the code and this is the gem version.

The CLI should probably report both AsciidoctorJ and asciidoctor gem versions to avoid confusing users.

robertpanzer:
…-version

With this PR the output of `asciidoctorj --version` looks like this:
```
AsciidoctorJRuby 2.0.0-SNAPSHOT (Asciidoctor 2.0.6) [https://asciidoctor.org]
Runtime Environment: jruby 9.2.7.0 (2.5.3)
```

I have no clue though if the modifications to the gradle build to create a properties file with the version of AsciidoctorJ is ok, I felt stupid adding this.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -13,6 +13,15 @@ For a detailed view of what has changed, refer to the {uri-repo}/commits/master[
 
 == Unreleased
 
+Enhancements::
+
+  * Extended version info printed by `asciidoctorj --version` to show versions of Asciidoctor and AsciidoctorJ separately (@abelsromero) (#791)
+
+Improvements::
+
+
+Bug Fixes::
+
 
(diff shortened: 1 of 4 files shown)


mojavelinux:
I like it, except I don't like "AsciidoctorJRuby". That's not the name of the project, so it's misleading. It should just be "AsciidoctorJ".

PR #798, opened 2019-04-15
Change AsciidoctorJRuby to AsciidoctorJ in version output

robertpanzer:
As proposed [here](https://github.com/asciidoctor/asciidoctorj/pull/795#issuecomment-483078104) `asciidoctorj --version` should show itself as AsciidoctorJ instead of AsciidoctorJRuby.

Apparently I got confused myself after the discussions about package names.


mojavelinux:
:+1:

PR #801, opened 2019-04-17
Add logging support for converters

robertpanzer:
This PR is a proposal for adding logging support for converters.
If a converter implements the interface `org.asciidoctor.log.Logging`, then Asciidoctor will set a `org.asciidoctor.log.LogHandler` via `Logging.setLogHandler(logHandler)` after the converter was created.

The AbstractConverter already implements this interface and provides the method `log(LogRecord)` to log.


ysb33r:
LGTM

robertpanzer:
Awesome, thanks! It corresponds mostly to what you proposed on gitter.

robertpanzer:
I'll try to do a new RC with this.

PR #804, opened 2019-04-22
Clarify classpath in README

### LINKED ISSUE #802: Missing package org.asciidoctor in jar?
Follow this link:
http://dl.bintray.com/asciidoctor/maven/org/asciidoctor/asciidoctorj/2.0.0-RC.1/asciidoctorj-2.0.0-RC.1.jar
then open the download in archive manager.
Folder org/asciidoctor contains only jruby
but when I download the source I can see
org.asciidoctor.Asciidoctor.java in folder asciidoctor-api which I expect to see that folder.
Confused...

robertpanzer:
Resolves #802.

Clarify in the README that it is not sufficient to exclusively add asciidoctorj.jar to the class path to use it.

Clarify that versions of AsciidoctorJ and Asciidoctor no longer match.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -20,13 +20,19 @@ ifndef::awestruct[:uri-docs: http://asciidoctor.org/docs]
 :uri-repo: https://github.com/asciidoctor/asciidoctorj
 :uri-issues: {uri-repo}/issues
 :uri-discuss: http://discuss.asciidoctor.org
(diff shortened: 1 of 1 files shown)


rockyallen:
Perhaps you should also remove the 'jar' link in the downloads section, since it is no use by itself. Or add all the required jars.

PR #813, opened 2019-05-03
Configure build scan plugin

aalmiray:
This small PR configures the `build-scan` Gradle plugin (https://guides.gradle.org/creating-build-scans/) which yields further insight into the build, allowing the team to make informed decisions on where and when the build may be tweaked to get faster, more performant builds.

Here's an example scan https://scans.gradle.com/s/rfbcjuu3gqvkk

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,12 +20,12 @@ before_install:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update          ; fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install graphviz; fi
 
-install: ./gradlew -S -Pskip.signing assemble
+install: ./gradlew -S -Pskip.signing assemble --scan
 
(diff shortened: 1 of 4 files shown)


robertpanzer:
Oh wow, is it really necessary to agree to the Gradle Terms of Service to use the build scans?
I find this document quite overwhelming and I don't feel I can judge the consequences when running it on a CI.
Can we possibly disable it by default on the CI systems?

aalmiray:
Yes, we can disable for CI by default if you want 😄

aalmiray:
Closing due to inactivity. Feel free to reach out to Gradle's Dev Advocacy should you like to continue.

robertpanzer:
Sorry about this.
I was actually just waiting for an update to the PR that does not enable it by default.
Running a build scan seems to assume that the user accepts a license agreement that I cannot judge.
Therefore I'd like to avoid running it by default on the CI.

PR #819, opened 2019-05-20
remove useless assignment and change the if condition

xuejiangtao:
hello i find a Code Quality Issue detected by SonarQube that Remove this useless assignment to local variable `equalsIndex`.
variable   `equalsIndex` will be assigned by `equalsIndex = attribute.indexOf(ATTRIBUTE_SEPARATOR)` , so the `int equalsIndex = -1` will be never used. so I change the condition to the assignment to fix the Code Quality Issue.
[https://rules.sonarsource.com/java/RSPEC-1854](url)


robertpanzer:
Thanks!

PR #823, opened 2019-05-29
InlineMacroProcessor must return a PhraseNode instead of a String

### LINKED ISSUE #821: 2.0.0: InlineMacroProcessor returning a String
Inspired by the Ruby Test suite, the java implementation for the “say” macro (usage `say:yo[]`) could be an `InlineMacroProcessor` like this:

```java
public class SayMacroProcessor extends InlineMacroProcessor {

    public SayMacroProcessor(String macroName, Map<String, Object> config) {
        super(macroName, config);
    }

    @Override
    public Object process(ContentNode parent, String target, Map<String, Object> attributes) {
        return target;
    }
}
```

This was valid with earlier version of Asciidoctor/AsciidoctorJ.

Due to the change introduced in `2.0.0.rc.3` returning a String is no longer the expected way to do this. 
A INFO log entry is raised:
```
INFO: expected substitution value for custom inline macro to be of type Inline; got String: say:yo[]
```

See the corresponding code in [substitutors.rb lines 333-343](https://github.com/asciidoctor/asciidoctor/blob/66ae48db19fc8f7d17df78a67e9b1208014f9de8/lib/asciidoctor/substitutors.rb#L333-L343).

---
All example in this repository of macro extending `InlineMacroProcessor` are calling `createPhraseNode(..)`

https://github.com/asciidoctor/asciidoctorj/blob/45857eccb6ae90f5e41b9f41c73cf47dbc87f670/asciidoctorj-core/src/test/java/org/asciidoctor/extension/ManpageMacro.java#L19-L26

But I did not find the way to return an `Inline` that contains just some text as `String`.

I have tried:

```java
return createPhraseNode(parent, "text", target, attributes, options);
```

But this creates an Error:

```
org.asciidoctor.jruby.internal.AsciidoctorCoreException: org.jruby.exceptions.NoMethodError: (NoMethodError) undefined method `convert_inline_text' for #<Asciidoctor::Converter::Html5Converter:0x10b4e7f8>
Did you mean?  convert_inline_button
               convert_inline_quoted
               convert_inline_menu
               convert_inline_image
               convert_inline_break
               convert_inline_kbd
```

Can you tell me what I should use?

I think the [Migrating from AsciidoctorJ 1.6.x to 2.0.x](https://github.com/asciidoctor/asciidoctorj/blob/master/docs/extension-migration-guide.adoc#migrating-from-asciidoctorj-16x-to-20x) section could be extended. I am ready to help when I know how to solve this.

jmini:
Fixes #821

### DIFF
--- a/asciidoctorj-core/src/test/groovy/org/asciidoctor/extension/AnnotatedLongInlineMacroProcessor.groovy
+++ b/asciidoctorj-core/src/test/groovy/org/asciidoctor/extension/AnnotatedLongInlineMacroProcessor.groovy
@@ -19,6 +19,6 @@ class AnnotatedLongInlineMacroProcessor extends InlineMacroProcessor {
         Map<String, Object> options = new HashMap<String, Object>()
         options['type'] = ':link'
         options['target'] = "${target}.html"
-        createPhraseNode(parent, 'anchor', target, attributes, options).convert()
+        createPhraseNode(parent, 'anchor', target, attributes, options)
     }
 }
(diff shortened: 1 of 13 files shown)


mojavelinux:
Technically the Ruby implementation accepts either return types, but the String return type is considered deprecated. Thus, I support this change.

jmini:
This PR is only about updating documentation & tests.

The API stays the same (expecting an `Object` as return type allow to return a `String` as previously or a `PhraseNode`).

robertpanzer:
Thanks for this update! 🎉

PR #825, opened 2019-06-02
Add sourcemap, catalog_assets and parse to OptionBuilder

### LINKED ISSUE #808: OptionsBuilder: support for sourcemap
From https://asciidoctor.org/docs/user-manual/#ruby-api-options

> `:sourcemap`
> Keeps track of the file and line number for each parsed block. (Useful for tooling applications where the association between the converted output and the source file is important). 
> default value: `false`
> Allowed values: boolean

Unless it is set to `true`, `StructuralNode#getSourceLocation()` returns `null`.

---
There is no support in `OptionsBuilder`.

Is it just missing (in this case I can create a PR to add it), or is there a reason why it is not present?

---
Workaround:
```java
OptionsBuilder.options()
                .option("sourcemap", true)
```

jmini:
Fixes #808

This PR do not contains the version bump to `2.1.0`

### DIFF
--- a/asciidoctorj-api/src/main/java/org/asciidoctor/Options.java
+++ b/asciidoctorj-api/src/main/java/org/asciidoctor/Options.java
@@ -18,7 +18,9 @@ public class Options {
     public static final String TO_DIR = "to_dir";
     public static final String MKDIRS = "mkdirs";
     public static final String SAFE = "safe";
+    public static final String SOURCEMAP = "sourcemap";
     public static final String ERUBY = "eruby";
+    public static final String CATALOG_ASSETS = "catalog_assets";
     public static final String COMPACT = "compact";
     public static final String DESTINATION_DIR = "destination_dir";
     public static final String SOURCE_DIR = "source_dir";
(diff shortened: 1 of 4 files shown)


asciidoctorj-api/src/main/java/org/asciidoctor/Options.java
```diff
@@ -124,10 +127,30 @@ public void setSafe(SafeMode safeMode) {
         this.options.put(SAFE, safeMode.getLevel());
     }
 
+    /**
+     * Keeps track of the file and line number for each parsed block. (Useful for tooling applications where the association between the converted output and the source file is important).
+     * 
+     * @param sourcemap
+     *            value.
+     */
+    public void setSourcemap(boolean sourcemap) {
```
robertpanzer: Would you mind updating this test to use this setter?
https://github.com/asciidoctor/asciidoctorj/blob/1a33db5703859b5c8ce9aa840270350f65f8e263/asciidoctorj-core/src/test/java/org/asciidoctor/WhenAsciiDocIsRenderedToDocument.java#L263-L268

robertpanzer:
Could you also please add an entry to the change log? :)

jmini:
I implemented the changes.

robertpanzer:
Thank you! 
Sorry for the delay. I am not as fast in reviewing as I’d like to be.

jmini:
Thank you for merging this. No worry about the delay. 
The pace of reviews in the Asciidoctor(J) project is great.

PR #830, opened 2019-06-23
marks templateDir in OptionsBuilder as deprecated

abelsromero:
Aligning with asciidoctor description of `:template_dir` (https://asciidoctor.org/docs/user-manual/#ruby-api-options) this PR marks `templareDir` method as deprecated pointing users to use `templateDirs` instead.

Additionally, fixes a couple of javadoc references in `Options`.


robertpanzer:
Thanks! 🍻

PR #834, opened 2019-07-22
Include logger name for java util log handler

### LINKED ISSUE #833: Include logger name for java util log handler
**Current behavior**
I am using [logback](https://logback.qos.ch/) in my application and [a bridge handler](https://www.slf4j.org/api/org/slf4j/bridge/SLF4JBridgeHandler.html) so that JUL (java.util.logging) messages are logged.

As is common in a java application, my logging pattern includes the logger name.  For log messages from asciidoc, I get `unknown.jul.logger` as the logger name for log messages from asciidoc.

**Preferred behavior**
After a look at the asciidoctorj code, I only one logger named `asciidoctor`.  
It would be nice if I saw `asciidoctor` in my logs rather than `unknown.jul.logger`.

**Action**
I've tested a fix locally for this will be happy to submit a PR.

lread:
When using the log4j java util logger bridge handler, logs now show
`asciidoctor` instead of `unknown.jul.logger` for logger name.

fixes #833

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/log/internal/JULLogHandler.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/log/internal/JULLogHandler.java
@@ -9,7 +9,8 @@ import java.util.logging.Logger;
 
 public class JULLogHandler implements LogHandler {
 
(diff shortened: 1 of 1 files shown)


robertpanzer:
Thank you!

PR #837, opened 2019-07-25
Updated changelog

robertpanzer:


mojavelinux:
Going on a tangent here...

Speaking from my own experience, I highly recommend making a CHANGELOG entry a required element of any PR (just like tests). I've been doing this in core and several other projects, and it has substantially cut down on the effort to release. In fact, I can't recommend the practice highly enough.

robertpanzer:
Yes, very good idea!
I'll also check how to add a template for a PR, so that somebody who creates a PR immediately sees that an update to the changelog is required.

PR #844, opened 2019-08-15
fixes #842 - allow LogHandlers to break conversion throwing an exception

### LINKED ISSUE #842: LogHandler errors not protected in `AbstractConverter`
While messing with a LogHandler I noticed that exceptions in `JRubyAsciidoctor` are protected with a try-catch.

```java
    @Override
    public void log(LogRecord logRecord) {
        for (LogHandler logHandler : logHandlers) {
            try {
                logHandler.log(logRecord);
            } catch (Exception e) {
                LOGGER.log(Level.WARNING, "Unexpected exception while logging Asciidoctor log entry", e);
            }
        }
    }
```

However, this is not the case for `AbstractConverter`, which just has 

```java
    @Override
    public void log(LogRecord logRecord) {
        if (logHandler != null) {
            logHandler.log(logRecord);
        }
    }
```

Is this on purpose?

abelsromero:


abelsromero:
Honest question, because I can't decide...Should add this as fix or improvement in Changelog?

robertpanzer:
As it seems to be inconsistent between different situations I’d personally treat this as  a bug. 
If all calls were consistently catching exceptions it could’ve been an improvement, but like this I think it’s a bug.

abelsromero:
Alongside with the update I fixed a couple of errors in the CHANGELOG.

robertpanzer:
Thanks! 🎉

PR #848, opened 2019-09-23
Update section about running on WildFly

anthonyvdotbe:
* update the JRuby dependency (this is required to make it work with WildFly 17.0.1, and matches the JRuby version that the current v2.1.0 depends upon)
* rename "WildFly AS" to just "WildFly"


robertpanzer:
Thank you for your PR!
Could you please also update the version of JCommander in the README?
We are using 1.72 right now.

anthonyvdotbe:
You're welcome. I updated the version of JCommander as well.

robertpanzer:
Thanks!

PR #849, opened 2019-09-24
adding close to Asciidoctor to avoid to leak when creating an asciidoctor instance

rmannibucau:
Currently Asciidoctor instance can't be released.
With JRuby it means at least 2 pools are leaking each time an instance is created.
It leads to a lot of issues in several environments.
This PR adds a close() method to Asciidoctor and ensures JRuby instance is properly released when called.

Saved ~1mn on 1mn16s of execution with exec-maven-plugin for instance and avoids leakage in EE context.

### DIFF
--- a/asciidoctorj-api/src/main/java/org/asciidoctor/Asciidoctor.java
+++ b/asciidoctorj-api/src/main/java/org/asciidoctor/Asciidoctor.java
@@ -24,7 +24,7 @@ import java.util.stream.Collectors;
 /**
  * @author lordofthejars
  */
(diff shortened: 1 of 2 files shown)


headius:
FWIW, a JRuby instance that becomes GCable will eventually shut itself down, but we strongly recommend tearing it down once the instance is no longer needed.

PR seems 👍 to me.

robertpanzer:
Thanks!
Just have to remember that the next release has to be 2.2.0 instead of 2.1.1.

PR #854, opened 2019-10-02
Documentation: link to tarball; add Quick Win section

il-pazzo:
This is a documentation-only update.

Added tarball links to the Distribution section.

Added a Quick Win section so users could get started straight away with doing a conversion, whether that's the end goal, or to get an idea of how the results would look after integrating the API.


robertpanzer:
Thanks! 🎉

PR #856, opened 2019-10-19
chore: upgrade to Gradle 5.6.3

Fiouz:
Upgrade build to Gradle 5.6.3


robertpanzer:
OSX build doesn't seem to work in general right now on TravisCI.
But it worked on my Mac, so I guess this is ok.

Thank you!

PR #858, opened 2019-10-30
Upgrade asciidoctorj-pdf to 1.5.0-beta.7

robertpanzer:


mojavelinux:
Can you also update rouge?

robertpanzer:
Sure! Is the latest 3.12.0 ok?

mojavelinux:
Yep. We're not coupled to its API, so this is about giving the use the latest and greatest. When it comes to the syntax highlighter, we always want to push it forward because it brings the latest language updates.

robertpanzer:
:+1: I'll update the PR.
The upstream build seems to be failing right now. I'll fix that and update the PR

robertpanzer:
Hmm, looks like coderay now places the stylesheet at the bottom of the page.
I wouldn't expect this looking at the code: https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/syntax_highlighter/coderay.rb#L35

robertpanzer:
Oh wait, this is my implementation, not the Ruby one :D

robertpanzer:
It's the other way around, Asciidoctor moved it into the head.
I might ignore the test for now until latest release and upstream are in sync again.

mojavelinux:
That's correct, it moved the head so the document validates. You could loosen the assertion to make sure it's in the document, but not whether it is in the head or body.

Issue #861, opened 2019-11-01
Gradle bintrayUpload not able to upload distribution


robertpanzer:
I was trying to make a release of AsciidoctorJ 2.2.0.
However the upload of asciidoctorj-distribution fails because it tries to upload jars from the libs/ directory created by this project, even though it only creates a zip and a tar in the distributions directory with the distribution plugin:
```
> Task :asciidoctorj-distribution:bintrayUpload FAILED
Caching disabled for task ':asciidoctorj-distribution:bintrayUpload' because:
  Build cache is disabled
Task ':asciidoctorj-distribution:bintrayUpload' is not up-to-date because:
  Task has not declared any outputs despite executing actions.
Gradle Bintray Plugin version: 1.8.4
:asciidoctorj-distribution:bintrayUpload: file /Users/robertpanzer/dev/asciidoctorj/asciidoctorj-distribution/build/libs/asciidoctorj-2.2.0-bin.jar.asc could not be found.
maven plugin was not applied, no pom will be uploaded.
:asciidoctorj-distribution:bintrayUpload: file /Users/robertpanzer/dev/asciidoctorj/asciidoctorj-distribution/build/libs/asciidoctorj-2.2.0-bin.jar.asc could not be found.
:asciidoctorj-distribution:bintrayUpload (Thread[Daemon worker Thread 10,5,main]) completed. Took 0.003 secs.
:bintrayPublish (Thread[Daemon worker Thread 10,5,main]) started.
```

@Fiouz Do you maybe have an idea what could be going on here?

Fiouz:
I am not familiar with the [Bintray plugin](https://github.com/bintray/gradle-bintray-plugin), but could you please confirm that the following error message also appear:

```
Execution failed for task ':asciidoctorj-distribution:bintrayUpload'.
> No such property: toSignArtifact for class: org.gradle.plugins.signing.Signature
```

If it does, the failure may not be directly related to the non-existent `.jar` file but rather due to an incompatibility between the gradle-bintray-plugin and Gradle 5.x:

- **gradle-bintray-plugin** [makes use](https://github.com/bintray/gradle-bintray-plugin/blob/1.8.4/src/main/groovy/com/jfrog/bintray/gradle/tasks/BintrayUploadTask.groovy#L514) of the `Signature.getToSignArtifact()` method
[...]

PR #862, opened 2019-11-02
fix: hack for Gradle 5 + signed artifacts + gradle-bintray-plugin 1.8.4

Fiouz:
WARNING: since the Groovy Gradle API is modified, this breaks build
isolation when sharing a common Gradle daemon instance.

This hack works because gradle-bintray-plugin is using Groovy dynamic
compilation (i.e. compiled in Groovy without @CompileStatic) hence it is
affected by Groovy runtime meta-programming.

Tested with: Gradle 5.6.3 / gradle-bintray-plugin 1.8.4

asciidoctor/asciidoctorj#861
bintray/gradle-bintray-plugin#300

### DIFF
--- a/asciidoctorj-distribution/build.gradle
+++ b/asciidoctorj-distribution/build.gradle
@@ -32,6 +32,24 @@ if ( !project.hasProperty('skip.signing') ) {
   signPom.enabled = false
 }
 
+/*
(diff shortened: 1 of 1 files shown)


robertpanzer:
Thanks a log! 🎉

PR #863, opened 2019-11-05
Use headless jdk on all platforms

nicerloop:


mojavelinux:
I'd go ahead and set it for all JVMs. There's no reason AsciidoctorJ should ever need to use AWT or Swing.

robertpanzer:
I don't know why, but the build fails, both on the CI systems and locally too with this error:
```
> Task :asciidoctorj-distribution:startScripts FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':asciidoctorj-distribution:startScripts'.
> Could not write to file '/Users/rpanzer/dev/asciidoctorj/asciidoctorj-distribution/build/scripts/asciidoctorj'.
```

asciidoctorj-distribution/src/main/resources/customUnixStartScript.txt
```diff
@@ -80,6 +80,10 @@ if [ -n "\$JAVA_HOME" ] ; then
 Please set the JAVA_HOME variable in your environment to match the
 location of your Java installation."
     fi
+# For Darwin, find JAVA_HOME for 1.8
+elif $darwin; then
+    JAVA_HOME=`/usr/libexec/java_home -v 1.8`
```
robertpanzer: Why is it necessary to run with Java 1.8?
It should also run with Java 11, if it does not then this would definitely be a bug.

robertpanzer:
I don't know why, but the build fails, both on the CI systems and locally too with this error:
```
> Task :asciidoctorj-distribution:startScripts FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':asciidoctorj-distribution:startScripts'.
> Could not write to file '/Users/rpanzer/dev/asciidoctorj/asciidoctorj-distribution/build/scripts/asciidoctorj'.
```
Can you please check?

asciidoctorj-distribution/src/main/resources/customUnixStartScript.txt
nicerloop: WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.headius.backport9.modules.Modules (file:/usr/local/Cellar/asciidoctorj/2.1.0/libexec/lib/jruby-complete-9.2.7.0.jar) to field java.io.FileDescriptor.fd
WARNING: Please consider reporting this to the maintainers of com.headius.backport9.modules.Modules
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release

nicerloop: These warnings are written to STDOUT, and asciidoctor-vscode detects an error; for now asciidoctorj runs without warning on java 8, not java 11

nicerloop: It seems it is a problem with jruby and java 9+ https://github.com/jruby/jruby/issues/5864

robertpanzer: In case somebody is offended by this I would leave it up to the user then to set the JAVA_HOME instead of making such a decision for them and additionally break  extensions that were compiled with Java 11.

robertpanzer: In particular only doing this on OSX seems very strange to me. 
I guess the same arguments should hold on Windows and Linux, no?

nicerloop: I don't know of a simple way to find and choose a java version on linux and windows, whereas macOX has this feature

robertpanzer: In my opinion it is very uncommon for a software to change the JAVA_HOME that a user has chosen explicitly. 
Unless I get feedback from other users saying that this is fine I don’t really want to merge this. 
Regarding the hint towards vscode: Does it really run the shell script on every run? I can hardly imagine this and I always assumed it was using asciidoctorjs. 

I am fine with adding the vm option for awt headless mode, even though it would be awesome to add a list of possible implications of this, or if there are none just state that.

nicerloop: Just checked, injecting JAVA_HOME inside asciidoctor-vscode settings works. I leave JAVA_HOME alone, and leave it to the user as previously done.

nicerloop: In fact, I just tried to set JAVA_HOME to 1.8 only if not set, instead of defaulting to java from PATH which will rapidly by 9+ and exhibit the reflective access warnings from jruby

nicerloop: regarding asciidoctor-vscode, it defaults to using asciidoctorjs which has some limitations regarding including external diagram files in particular, thus I tried to used asciidoctorj for redering, evenf if it is slower

nicerloop: regarding awt headless, the adverse implication of _not_ being headless on macOX is that a full GUI application is launched, appears in the dock even if it has no window, and grabs the keyboard focus, which is quite annoying for a command-line tool like asciidoctorj

mojavelinux: > In my opinion it is very uncommon for a software to change the JAVA_HOME that a user has chosen explicitly. Unless I get feedback from other users saying that this is fine I don’t really want to merge this.

Agreed about JAVA_HOME.

robertpanzer: > regarding awt headless, the adverse implication of not being headless on macOX is that a full GUI application is launched, appears in the dock even if it has no window, and grabs the keyboard focus, which is quite annoying for a command-line tool like asciidoctorj

Agreed, I just want to be sure that it doesn't break the asciidoctor-diagram extension.
I tested manually that the Ditaa and Plantuml sequence and class diagrams still work.

robertpanzer:
Looks good to me.
I also think it is the right way to add the option to the DEFAULT_JVM_OPTS, that way a user can still override this property when required like this:
```
JAVA_OPTS="-Djava.awt.headless=false" asciidoctorj
```
or 
```
ASCIIDOCTORJ_OPTS="-Djava.awt.headless=false" asciidoctorj
```

PR #864, opened 2019-11-10
restore STDIN input

nicerloop:
JCommander has been updated and now passes '-' as an input File when encountered

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/cli/AsciidoctorInvoker.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/cli/AsciidoctorInvoker.java
@@ -78,7 +78,7 @@ public class AsciidoctorInvoker {
                 timings.callMethod(JRubyRuntimeContext.get(asciidoctor).getCurrentContext(), "print_report");
             }
 
-            if (!"".equals(output.trim())) {
+            if (output != null && !"".equals(output.trim())) {
                 System.out.println(output);
             }
         }
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/cli/AsciidoctorInvoker.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/cli/AsciidoctorInvoker.java
@@ -150,14 +150,10 @@ public class AsciidoctorInvoker {
     }
 
     private String renderInput(Asciidoctor asciidoctor, Options options, List<File> inputFiles) {
-        
 
-        // jcommander bug makes this code not working.
-        /*
-        if("-".equals(inputFile)) {
+        if(inputFiles.size() == 1 && "-".equals(inputFiles.get(0).getName())) {
             return asciidoctor.convert(readInputFromStdIn(), options);
         }
-        */
 
         StringBuilder output = new StringBuilder();
 
@@ -186,8 +182,8 @@ public class AsciidoctorInvoker {
     }
 
     private String readInputFromStdIn() {
-        Scanner in = new Scanner(System.in);
-        String content = in.nextLine();
+        Scanner in = new Scanner(System.in).useDelimiter("\\A");
+        String content = in.next();
         in.close();
 
         return content;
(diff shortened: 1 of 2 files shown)


robertpanzer:
Can you please describe what this PR is solving?
I had a look at the linked issue, but I am not fully clear about what this change achieves.

asciidoctorj-core/src/main/java/org/asciidoctor/jruby/cli/AsciidoctorInvoker.java
```diff
@@ -186,8 +182,8 @@ private String renderInput(Asciidoctor asciidoctor, Options options, List<File>
     }
 
     private String readInputFromStdIn() {
-        Scanner in = new Scanner(System.in);
-        String content = in.nextLine();
+        Scanner in = new Scanner(System.in).useDelimiter("\\A");
+        String content = in.next();
```
robertpanzer: I see, this code was not correct before and only read the first line of the input from stdin (if it would have been called, which it wasn't yet)

Can you please call IOUtils.readFull() here instead of copying its implementation here?

(I am actually not too happy about using Scanner for this and would like to replace that now that I have seen it. Scanner doesn't seem to be the right thing for reading a file fully into memory)

robertpanzer:
Now I understand what this is about.
Thanks for this fix!

Can you please open a ticket, because even I wasn't aware that the AsciidoctorJ CLI currently does not support reading from stdin.
Also could you please extend the description of the input files parameter at `AsciidoctorCLIOptions.parameters` stating that a single value of `-` reads the input from stdin?

nicerloop:
Document use of '-' to read STDIN.
Use IOUtils instead of Scanner - not really more legible in my opinion, due to exceptions in close().
Open and link issue https://github.com/asciidoctor/asciidoctorj/issues/865

robertpanzer:
While testing I figured these 2 differences from the asciidoctor CLI, which I think we should align with:
1. When converting to an html document from stdin, asciidoctor creates a full document including headerFooter, whereas asciidoctorj now only creates the body of the document, ie without the headerFooter option:
```
cat ~/dev/asciidoctorj/test.adoc|asciidoctor -|head
<!DOCTYPE html>
<html lang="en">
<head>
<meta charset="UTF-8">
...
```
vs
```
cat ~/dev/asciidoctorj/test.adoc|./asciidoctorj -|head
<div id="preamble">
<div class="sectionbody">
<div class="paragraph">
<p>Test</p>
</div>
</div>
</div>
...
```
2. When converting to pdf without outputfile asciidoctor writes the PDF as text to stdout, while asciidoctorj fails with an exception:
```
cat ~/dev/asciidoctorj/test.adoc| asciidoctor-pdf - 2&>/dev/null |head
%PDF-1.4
%����
1 0 obj
<< /Title (Test)
/Creator (Asciidoctor PDF 1.5.0.beta.6, based on Prawn 2.2.2)
/Producer (Asciidoctor PDF 1.5.0.beta.6, based on Prawn 2.2.2)
...
```
vs
```
cat ~/dev/asciidoctorj/test.adoc| ./asciidoctorj -b pdf - |head
Nov 13, 2019 4:34:27 PM org.asciidoctor.jruby.internal.JRubyAsciidoctor convert
SEVERE: org.jruby.RubyTypeError
Exception in thread "main" org.asciidoctor.jruby.internal.AsciidoctorCoreException: org.jruby.exceptions.TypeError: (TypeError) cannot convert instance of class org.jruby.gen.RubyObject50 to class java.lang.String
	at org.asciidoctor.jruby.internal.JRubyAsciidoctor.convert(JRubyAsciidoctor.java:318)
	at org.asciidoctor.jruby.internal.JRubyAsciidoctor.convert(JRubyAsciidoctor.java:334)
	at org.asciidoctor.jruby.internal.JRubyAsciidoctor.convert(JRubyAsciidoctor.java:329)
	at org.asciidoctor.jruby.cli.AsciidoctorInvoker.renderInput(AsciidoctorInvoker.java:156)
	at org.asciidoctor.jruby.cli.AsciidoctorInvoker.invoke(AsciidoctorInvoker.java:73)
	at org.asciidoctor.jruby.cli.AsciidoctorInvoker.main(AsciidoctorInvoker.java:222)
Caused by: org.jruby.exceptions.TypeError: (TypeError) cannot convert instance of class org.jruby.gen.RubyObject50 to class java.lang.String
```

I guess if no output is set and input is read from stdin, we need to set the stream stdout as the output stream.

PR #866, opened 2019-11-11
Run CI build using Github Actions

robertpanzer:
This PR adds CI builds using Github Actions.
Right now it runs a build on master on push to master and every day at 9AM UTC.
PRs are also built.

In the beginning we can keep this in parallel to the build on TravisCI until we are confident that Github works as we need it.

Open points:
- Github Actions also supports building on Windows, I plan to add this later
- I have to look up how to implement the trigger that we currently use for the asciidoctor upstream build. Until we have that we have the scheduled daily build though which should also capture changes, even though it is of course not easy to track down the breaking changes if there were multiple commits to Asciidoctor within a day.

### DIFF
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -1,0 +1,46 @@
+name: Build Master
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+  schedule:
+    - cron: '0 0 * * *'
+
+jobs:
+  build:
+    name: Build
+    strategy:
+      max-parallel: 2
+      matrix:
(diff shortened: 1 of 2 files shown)


robertpanzer:
This is probably a way trigger an upstream build from the asciidoctor/asciidoctor repo: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/events-that-trigger-workflows#external-events-repository_dispatch

mojavelinux:
Super cool! You're such a pioneer!

robertpanzer:
Thanks! :)
I’ll be traveling next week, I hope I’ll find some time then to add the upstream triggers

PR #875, opened 2019-11-16
Fix reading input from stdin and writing to stdout

### LINKED ISSUE #865: CLI does not support reading input from STDIN
The corresponding code in commented out in AsciidoctorInvoker.java due to JCommander inability to handle '-' as a parameter. JCommander limitation lifted by newer version already included in master.

robertpanzer:
## Kind of change

[X] Bug fix
[ ] New non-breaking feature
[ ] New breaking feature
[ ] Documentation update
[ ] Build improvement

### What is the goal of this pull request?

This PR builds on the awesome changes from @nicerloop in #864 
On top of that I added a few fixes where that PR deviated from the behaviour in asciidoctor:

`echo "Hello World"|asciidoctorj -` will create a full document including header and footer, just like asciidoctor does.

`echo "Hello World"|asciidoctorj - -b pdf` writes the pdf document to stdout.

### How does it achieve that?

In addition to the changes in #864 this PR by default sets the option `:standalone` to `true`, that's also what the Asciidoctor CLI does [here](https://github.com/asciidoctor/asciidoctor/blob/4155916105b3276e5971349b1730b9d095d29db3/lib/asciidoctor/cli/options.rb#L15).
If the cli argument `-s/--no-header-footer` is passed, `:standalone` is set to `false`, also similar to what the Asciidoctor CLI does [here](https://github.com/asciidoctor/asciidoctor/blob/4155916105b3276e5971349b1730b9d095d29db3/lib/asciidoctor/cli/options.rb#L73-L75)

Then, instead of printing out the output, it sets the outfile option to the stdout stream, so that also PDFs are correctly converted to stdout instead of printing the result of the PDF converter, which is the PDF converter object itself.

Yes, using the Asciidoctor command line parsing directly, but there are some caveats that I have to think about, like additional options for setting sth like a classpath, which doesn't exist in Asciidoctor/Ruby.

## Issue

Fixes #865

### DIFF
--- a/asciidoctorj-api/src/main/java/org/asciidoctor/Options.java
+++ b/asciidoctorj-api/src/main/java/org/asciidoctor/Options.java
@@ -19,6 +19,7 @@ public class Options {
     public static final String MKDIRS = "mkdirs";
     public static final String SAFE = "safe";
     public static final String SOURCEMAP = "sourcemap";
+    public static final String STANDALONE = "standalone";
     public static final String ERUBY = "eruby";
     public static final String CATALOG_ASSETS = "catalog_assets";
     public static final String COMPACT = "compact";
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/cli/AsciidoctorCliOptions.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/cli/AsciidoctorCliOptions.java
@@ -248,6 +248,10 @@ public class AsciidoctorCliOptions {
         return this.version;
     }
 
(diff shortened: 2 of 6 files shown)


robertpanzer:
@nicerloop Would be great to get some feedback from you on this, if this still matches your requirements.

nicerloop:
I was starting to look at the points you rightfully made in https://github.com/asciidoctor/asciidoctorj/pull/864#issuecomment-553456594, and you beat me to it. Tested on my side, and it completely fulfils my needs, and is much more in line with the asciidoctor CLI. +1 from me obviously :)

PR #876, opened 2019-11-18
Add upstream build trigger

robertpanzer:
## Kind of change

[ ] Bug fix
[ ] New non-breaking feature
[ ] New breaking feature
[ ] Documentation update
[x] Build improvement

## Description

This PR adds a build trigger for the upstream build, so that asciidoctor/asciidoctor can trigger upstream builds on every commit, similar to what we have today with travis.

How does it achieve that?

Registers a workflow that is triggered on `repository_dispatch`.
An upstream build can be triggered like this now:
```
curl -XPOST https://api.github.com/repos/asciidoctor/asciidoctorj/dispatches \
  -H 'Accept: application/vnd.github.everest-preview+json' \
  -H 'authorization: token <token>' \
  -H 'content-type: application/json' \
  --data '{
  "event_type":"test_upstream", 
  "client_payload": {
    "message":"Triggered by commit xyz", 
    "branch": "master"
  }
}'
```
The token must have the `repo` scope.

Staying on Travis or moving to some other CI system.

No

### DIFF
--- a/.github/workflows/upstream-build.yaml
+++ b/.github/workflows/upstream-build.yaml
@@ -1,0 +1,19 @@
+name: Build Upstream
+on: repository_dispatch
+
+jobs:
+  build-upstream:
+    name: ${{ github.event.client_payload.message }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+        with:
+          fetch-depth: 1
+          ref: ${{ github.event.client_payload.branch }} 
+      - uses: actions/setup-java@v1
+        with:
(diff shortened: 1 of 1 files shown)


robertpanzer:
An example run that I triggered manually is available here: https://github.com/robertpanzer/asciidoctorj/commit/f5902debdb85e874da8e45c68365737fda4892e6/checks?check_suite_id=316662793

I'll try to add this call here: https://github.com/asciidoctor/asciidoctor/blob/master/tasks/dependents.rake

PR #898, opened 2020-01-19
start implement feature to use @Name annotation as meta-annotation

### LINKED ISSUE #897: Support @Name annotation as meta annotation
hi guys,

it would be nice if we could support `@Name` annotations also as meta annotations, so i could create my own annotations. I create a little extension support framework, for my different extensions. In this framework i implement a feature to register required libraries via annotations. 

So i have an annotation called `AsciidoctorRequiredLibrary` and with that i could create in my extensions an annotation like the following.

```java
@Documented
@Target(TYPE)
@Retention(RUNTIME)
@AsciidoctorjRequiredLibrary("asciidoctor-diagram")
@Name("decisiontable")
public @interface DecisionTableBlockMacroExtensionRequireAsciidoctorDiagram {
}
```

But if we support `@Name` as an meta annotation, which recursive level would we support?

uniqueck:

## Kind of change

- [ ] Bug fix
- [ ] New breaking feature
- [ ] Documentation update
- [ ] Build improvement

## Description

See #897

How does it achieve that?
Improve method getName in JavaExtensionRegistryImpl, to support recursive search for an @Name annotation.

maybe, we can discuss it.

I only provide currently a new test in the class BlockMacroRegistrationTest, i think we have to add it to the other RegistrationTests also.

## Issue
Fixes #897 

## Release notes

Name annotation can used as a meta-annotation

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/internal/JavaExtensionRegistryImpl.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/internal/JavaExtensionRegistryImpl.java
@@ -22,6 +22,11 @@ import org.jruby.Ruby;
 import org.jruby.RubyClass;
 import org.jruby.RubyModule;
 
+import java.lang.annotation.Annotation;
+import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
 public class JavaExtensionRegistryImpl implements JavaExtensionRegistry {
 
     private final JRubyAsciidoctor asciidoctor;
@@ -192,8 +197,9 @@ public class JavaExtensionRegistryImpl implements JavaExtensionRegistry {
 
     @Override
     public JavaExtensionRegistry block(Class<? extends BlockProcessor> blockProcessor) {
-        String name = getName(blockProcessor);
-        block(name, blockProcessor);
+        getNames(blockProcessor).forEach(name -> {
+            block(name, blockProcessor);
+        });
         return this;
     }
 
@@ -222,9 +228,10 @@ public class JavaExtensionRegistryImpl implements JavaExtensionRegistry {
 
     @Override
     public JavaExtensionRegistry blockMacro(Class<? extends BlockMacroProcessor> blockMacroProcessor) {
-        String name = getName(blockMacroProcessor);
-        RubyClass rubyClass = BlockMacroProcessorProxy.register(asciidoctor, blockMacroProcessor);
-        getAsciidoctorModule().callMethod("block_macro", rubyClass, rubyRuntime.newString(name));
+        getNames(blockMacroProcessor).forEach(name -> {
+            RubyClass rubyClass = BlockMacroProcessorProxy.register(asciidoctor, blockMacroProcessor);
+            getAsciidoctorModule().callMethod("block_macro", rubyClass, rubyRuntime.newString(name));
+        });
         return this;
     }
 
@@ -290,9 +297,10 @@ public class JavaExtensionRegistryImpl implements JavaExtensionRegistry {
 
     @Override
     public JavaExtensionRegistry inlineMacro(Class<? extends InlineMacroProcessor> inlineMacroProcessor) {
-        String name = getName(inlineMacroProcessor);
-        RubyClass rubyClass = InlineMacroProcessorProxy.register(asciidoctor, inlineMacroProcessor);
-        getAsciidoctorModule().callMethod("inline_macro", rubyClass, rubyRuntime.newString(name));
+        getNames(inlineMacroProcessor).forEach(name -> {
+            RubyClass rubyClass = InlineMacroProcessorProxy.register(asciidoctor, inlineMacroProcessor);
+            getAsciidoctorModule().callMethod("inline_macro", rubyClass, rubyRuntime.newString(name));
+        });
         return this;
     }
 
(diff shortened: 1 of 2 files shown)


robertpanzer:
Hi @uniqueck, thanks for your PR.
I like that idea.
I am just unsure about the best behavior for handling multiple annotations.
Just picking the first one might lead to surprises, I am not aware that there are any guarantees on the order of annotations returned by `Class.getAnnotations()`.
I wonder if it would make more sense to throw an exception in case of multiple names, or register them under all names.

asciidoctorj-core/src/main/java/org/asciidoctor/jruby/internal/JavaExtensionRegistryImpl.java
```diff
@@ -318,12 +324,27 @@ public JavaExtensionRegistry inlineMacro(String inlineMacroProcessor) {
...
+    private String getName(List<Annotation> annotations) {
+        String name = null;
+        Optional<Annotation> nameAnnotation = annotations.stream().filter(a -> a.annotationType().isAssignableFrom(Name.class)).findFirst();
+
+        if (nameAnnotation.isPresent()) {
+            name = ((Name)nameAnnotation.get()).value();
+        } else {
+            for (Annotation anno : annotations) {
+                List<Annotation> tempAnnotationList = new ArrayList<>(Arrays.asList(anno.annotationType().getAnnotations()));
+                tempAnnotationList.removeAll(annotations);
```
robertpanzer: This is to avoid following direct cycles, right? (Class C has an annotation A1 and A2, where A2 also has an annotation A1).
Are there situations where other longer cycles could happen?
What about doing a breadth-first-search that keeps a context of all yet visited annotations?
I think that would be clearer.

uniqueck:
Ok thanks for your thoughts. I came back with better solution. 
Is it possible to register the same extension with different names? If not we should throw an exception. 

Regards Constantin

robertpanzer:
It is generally possible unless I am missing sth right now. 
Technically there's nothing speaking against that.

asciidoctorj-core/src/main/java/org/asciidoctor/jruby/internal/JavaExtensionRegistryImpl.java
uniqueck: Hi @robertpanzer,

Some time has passed, but now I found time again to continue here. ;)
I introduce a list to mark all visited annotations, so i think this approach is a better way, than the first one.

uniqueck:
> Hi @uniqueck, thanks for your PR.
> I like that idea.
> I am just unsure about the best behavior for handling multiple annotations.
> Just picking the first one might lead to surprises, I am not aware that there are any guarantees on the order of annotations returned by `Class.getAnnotations()`.
> I wonder if it would make more sense to throw an exception in case of multiple names, or register them under all names.

Ok for that i have to investigate for another solution. We collect all Name annotations and then we register this extensions with all names.

uniqueck:
@robertpanzer can you review it again, please?

asciidoctorj-core/src/main/java/org/asciidoctor/jruby/internal/JavaExtensionRegistryImpl.java
```diff
@@ -318,12 +328,31 @@ public JavaExtensionRegistry inlineMacro(String inlineMacroProcessor) {
...
-        if (nameAnnotation == null) {
+    private List<Name> getNameAnnotations(List<Annotation> annotations, List<Annotation> visitedAnnotations) {
+        List<Name> allNameAnnotations = new ArrayList<>();
+        for (Annotation anno : annotations) {
+            if (anno.annotationType().isAssignableFrom(Name.class)) {
+                allNameAnnotations.add((Name) anno);
+            } else {
+                visitedAnnotations.add(anno);
+                List<Annotation> annotationsFromAnnotation = getAnnotationsFromAnnotationWithoutAlreadyVisitedAnnotations(anno, visitedAnnotations);
+                allNameAnnotations.addAll(getNameAnnotations(annotationsFromAnnotation, visitedAnnotations));
```
robertpanzer, line 339: Do we need the detection of duplicates?
If I understand it correctly it would only result in a relatively small optimisation in cases like this:
```java

@Target(TYPE)
@Retention(RUNTIME)
@Name("ABC")
public @interface MetaRoot {
    String NAME = "annotated";
}

@Target(TYPE)
@Retention(RUNTIME)
@MetaRoot
public @interface MetaOne {
    String NAME = "annotated";
}

@Target(TYPE)
@Retention(RUNTIME)
@MetaRoot
public @interface MetaTwo {
    String NAME = "annotated";
}

@MetaOne
@MetaTwo
class MyBlockProcessor extends AbstractBlockProcessor {...}
```

I think it is fine to fully iterate through the tree and collect the names into a set instead of a list.

uniqueck:
Hi @robertpanzer,

it's done. I used a set and remove unnecessary code.

robertpanzer:
Awesome! Could you please add an entry to the CHANGELOG.adoc?

robertpanzer:
I can update the change log as well.
In the next step the other annotations like `@Location`, `@Format`, `@Contexts`, etc should be supported as meta annotations too.

uniqueck:
Thanks for the changelog, i forgot that in my daily business. I start implement an feature for the other annotations. I think we can also support an annotation for required libraries. So you can combine a Name annotation and a annotation like RequiredLibrary. With the ability of meta annotation you can create an annotation like that. RequireLibraryAsciidoctorDiagram. What do you think about that?

PR #917, opened 2020-04-26
Fixes #916. Add factory methods to create lists

### LINKED ISSUE #916: How to create a list
I am trying to create BlockMacroProcessor (using Java) that will return block that contains a list of items. However, I am not sure how to create a list. There is a createListItem method that takes List object as parameter, but there is not a method to create list itself.
Is it possible to create list from scratch?

Note: I can see that in Ruby it is possible to instantiate List object and then add list items to it.
Note #2: I apologize for creating an issue, but wasn't sure where to ask for help.

robertpanzer:
## Kind of change

- [X] Bug fix
- [ ] New breaking feature
- [ ] Documentation update
- [ ] Build improvement

## Description

As @glisicivan identified in #916 that the Extension API is currently missing methods to create Lists.
This PR adds the missing methods.
 
How does it achieve that?

The PR adds the following method to `Processor` and a few overrides if no attributes or options are used:

```java
org.asciidoctor.ast.List createList(StructuralNode parent, 
        String context, 
        Map<String, Object> attributes, 
        Map<Object, Object> options);
```

To create an ordered list `"olist"` must be passed as the context parameter.
For unordered lists `"ulist"` must be passed.

That means to create an ordered list simply call this from within an extension:
```java
var list = createList(parent, "list");
```

An alternative might have been to define methods like `createUnorderedList()`, `createOrderedList()` etc.
But passing the context as a String closer matches the API for `createBlock()` etc and is closer to the original Ruby API.

@mojavelinux Wdyt?

## Issue

Fixes #916

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -22,11 +22,8 @@ Improvement::
 
 Bug Fixes::
 
-  * Add missing factory methods to create Lists (@glisicivan) (#916)
-
-Build::
-
-  * Upgrade jruby-gradle-plugin to 2.0.0-alpha.7 and load Gems directly from rubygems.org instead of using torquebox Maven proxy.
+  * Add missing factory methods to create Lists. (@glisicivan) (#916)
+    The API `Processor.createList()` is experimental and may change with any release until declared to be stable.
 
 Documentation::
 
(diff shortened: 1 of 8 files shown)


robertpanzer:
I declared the createList API as experimental.
That way I can create a new release with the fix and wait for feedback without having to make major releases for bug fixes.

(I am not afraid of major releases, but I wouldn't want to go through several rounds of major releases because of missing feedback.)

Issue #953, opened 2020-09-15
Don't check for a specific version of highlight.js in the tests


mojavelinux:
The version of highlight.js in core changes gets updated from time to time. The tests should be able to handle this change without having to update the test each time. Leave the highlight.js version as a wildcard in the test.

mojavelinux:
This test is still too fragile. I think it should be checking just that highlight.js is somewhere in the HTML. It's the the responsibility of AsciidoctorJ to put it there, so it shouldn't care where it is or what the exact text is. Every time we change something in core, this test seems to fail. But there's no reason it should.

robertpanzer:
The intention of this test was to prove that it's possible to write a syntax highlighter in Java that produces an identical result to the counterpart in Ruby. 

Just checking if the word highlightjs is part of the result doesn't really meet that expectation imo. 

Maybe we could decouple the test entirely from the Ruby version and just check against a fixed, expected result instead.

mojavelinux:
Ah, I understand. Sounds like a good plan. :+1:

PR #954, opened 2020-09-18
Fix upstream build

robertpanzer:

## Kind of change

- [ ] Bug fix
- [ ] New breaking feature
- [ ] Documentation update
- [x] Build improvement

## Description

This PR fixes the upstream build which does not succeed anymore since the upstream version of highlights has changed.

How does it achieve that?
It's an intermediate step of updating the version number used by the test highlighter.

Yes, as suggested in (#953) the version number should not be checked for exact equality.
Unfortunately I don't have the time right now, but still want to unblock the upstream built.

## Issue

Fixes #Issue 

## Release notes

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -13,6 +13,10 @@ For a detailed view of what has changed, refer to the {uri-repo}/commits/master[
 
 == Unreleased
 
+Build::
+
+* Fix upstream build to adapt for changes to highlightjs syntax highlighter. (#954)
+
 == 2.4.1 (2020-09-10)
 
 Bug Fixes::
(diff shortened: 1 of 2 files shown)


abelsromero:
> Unfortunately I don't have the time right now, but still want to unblock the upstream built.

I have time now, but I don't fully understand the problem. Is just that normal tests work, but upstream fails because a change in asciidoctor main/dev branch?

robertpanzer:
Yes, exactly.
Right now the test uses 2 syntax highlighters implemented in Groovy and checks that either of the results when using it matches exactly what is produced by Asciidoctor Ruby.
Now this breaks when Asciidoctor upgrades the version of highlightjs as the version number gets into the resulting HTML.

One of the 2 highlighters should match the current version of Asciidoctor Ruby, the other should match the current upstream.
(Besides the version number there sometimes are other changes, for example where the script tag is placed)

The goal would be to make the check more flexible so that the test does not break when Asciidoctor Ruby upgrades the version Highlightjs.

PR #968, opened 2020-10-17
AST now provides access to document catalog footnotes and refs

### LINKED ISSUE #958: AST access to footnotes and refs
I am experimenting with [writing a converter](https://github.com/asciidoctor/asciidoctorj/blob/master/docs/integrator-guide.adoc#publish-everywhere-adapt-asciidoctor-to-your-own-target-format) with AsciidoctorJ (thanks for providing the capability!).

I am using [htm5.rb](https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/converter/html5.rb) as a reference. 

There are currently 2 items that the Ruby code is using that I don't see mirrored in AsciidoctorJ AST classes.

1. footnotes - [used to render the footnotes at the bottom of a document](https://github.com/asciidoctor/asciidoctor/blob/cabde9c25fe040221ad71bb6aaf8754d2e5ec099/lib/asciidoctor/converter/html5.rb#L327).
2. refs - [used, in a way my limited Ruby grokking skills does not fully understand yet, in inline anchors](https://github.com/asciidoctor/asciidoctor/blob/cabde9c25fe040221ad71bb6aaf8754d2e5ec099/lib/asciidoctor/converter/html5.rb#L1115).

Should these be exposed to the AsciiDoctorJ AST? 
Or maybe they are available in some way I have missed?
Are there other similar values that should be exposed?

If exposing these values makes sense, I am happy to do the work. 
To me they look like they should be accessible from [Document](https://github.com/asciidoctor/asciidoctorj/blob/master/asciidoctorj-api/src/main/java/org/asciidoctor/ast/Document.java).
Lemme know!

lread:
## Kind of change

- [ ] Bug fix
- [ ] New breaking feature
- [ ] Documentation update
- [ ] Build improvement

## Description

Adds AST access to document footnotes and refs so that a converter may refer to them.

Access is provided under `Document` via:

- `getCatalog()`
     - `getFootnotes()`
     - `getRefs()`

## Issue

Fixes #958

## Release notes

CHANGELOG has been updated accordingly.

### DIFF
--- a/asciidoctorj-api/src/main/java/org/asciidoctor/ast/Catalog.java
+++ b/asciidoctorj-api/src/main/java/org/asciidoctor/ast/Catalog.java
@@ -1,0 +1,18 @@
+package org.asciidoctor.ast;
+
+import java.util.List;
+
+/**
+ * The asciidoctor catalog contains data collected by asciidoctor that is useful to a converter.
+ * The catalog is considered an internal structure by asciidoctor, and subject to change.
+ */
+public interface Catalog {
+
+    /**
+     * Note that asciidoctor does not populate footnotes until after the
+     * Document's content has been converted.
+     *
+     * @return footnotes occurring in document.
+     */
+    List<Footnote> getFootnotes();
+}
--- a/asciidoctorj-api/src/main/java/org/asciidoctor/ast/Catalog.java
+++ b/asciidoctorj-api/src/main/java/org/asciidoctor/ast/Catalog.java
@@ -2,15 +2,11 @@ package org.asciidoctor.ast;
 
 import java.util.List;
 
-/**
- * The asciidoctor catalog contains data collected by asciidoctor that is useful to a converter.
- * The catalog is considered an internal structure by asciidoctor, and subject to change.
- */
+
 public interface Catalog {
 
     /**
-     * Note that asciidoctor does not populate footnotes until after the
-     * Document's content has been converted.
+     * Note that footnotes are only available after `Document.getContent()` has been called.
      *
      * @return footnotes occurring in document.
      */
--- a/asciidoctorj-api/src/main/java/org/asciidoctor/ast/Catalog.java
+++ b/asciidoctorj-api/src/main/java/org/asciidoctor/ast/Catalog.java
@@ -1,6 +1,7 @@
 package org.asciidoctor.ast;
 
 import java.util.List;
+import java.util.Map;
 
 
 public interface Catalog {
@@ -8,7 +9,24 @@ public interface Catalog {
     /**
      * Note that footnotes are only available after `Document.getContent()` has been called.
      *
+     * A converter uses cataloged footnotes to render them, presumably, at the bottom of a document.
+     *
      * @return footnotes occurring in document.
      */
     List<Footnote> getFootnotes();
+
+    /**
+     * Refs is a map of asciidoctor ids to asciidoctor document elements.
+     *
+     * For example, by default, each section is automatically assigned an id.
+     * In this case the id would map to a {@link Section} element.
+     *
+     * Ids can also be explicitly assigned by document authors to any document element.
+     * See https://asciidoctor.org/docs/user-manual/#id
+     *
+     * A converter might use cataloged refs to lookup ids to support rendering inline anchors.
+     *
+     * @return a map of ids to elements that asciidoctor has collected from the document.
+     */
+    Map<String, Object> getRefs();
 }
--- a/asciidoctorj-api/src/main/java/org/asciidoctor/ast/Document.java
+++ b/asciidoctorj-api/src/main/java/org/asciidoctor/ast/Document.java
@@ -75,4 +75,7 @@ public interface Document extends StructuralNode {
      * @param state The state in which to put the sourcemap (true = on, false = off).
      */
     void setSourcemap(boolean state);
+
+
+    Catalog getCatalog();
 }
--- a/asciidoctorj-api/src/main/java/org/asciidoctor/ast/Document.java
+++ b/asciidoctorj-api/src/main/java/org/asciidoctor/ast/Document.java
@@ -77,5 +77,12 @@ public interface Document extends StructuralNode {
     void setSourcemap(boolean state);
 
 
+    /**
+     * The catalog contains data collected by asciidoctor that is useful to a converter.
+     *
+     * Note that the catalog is not part of the asciidoctor public API and is subject to change.
+     *
+     * @return catalog
+     */
     Catalog getCatalog();
 }
--- a/asciidoctorj-api/src/main/java/org/asciidoctor/ast/Footnote.java
+++ b/asciidoctorj-api/src/main/java/org/asciidoctor/ast/Footnote.java
@@ -1,0 +1,10 @@
+package org.asciidoctor.ast;
+
+public interface Footnote {
+
+   Long getIndex();
+
+   String getId();
+
+   String getText();
+}
--- a/asciidoctorj-api/src/main/java/org/asciidoctor/ast/Footnote.java
+++ b/asciidoctorj-api/src/main/java/org/asciidoctor/ast/Footnote.java
@@ -2,9 +2,24 @@ package org.asciidoctor.ast;
 
 public interface Footnote {
 
+   /**
+    * The index is the number asciidoctor has assigned to the footnote.
+    * Footnotes start at 1 and are numbered consecutively throughout the document.
+    *
+    * @return footnote number
+    */
    Long getIndex();
 
+   /**
+    * Each footnote can optionally have an id.
+    * Ids are used when a document author wants to reference a single footnote more than once.
+    *
+    * @return footnote id or null
+    */
    String getId();
 
+   /**
+    * @return the text the document author has specified for the footnote
+    */
    String getText();
 }
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/ast/impl/FootnoteImpl.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/ast/impl/FootnoteImpl.java
@@ -16,7 +16,7 @@ public class FootnoteImpl implements Footnote {
     private String text;
 
     private static Object aref(RubyStruct s, String key)  {
-       return JavaEmbedUtils.rubyToJava(s.aref(JavaEmbedUtils.javaToRuby(s.getRuntime(), key)));
+       return JavaEmbedUtils.rubyToJava(s.aref(s.getRuntime().newString(key)));
     }
 
     public static Footnote getInstance(Long index, String id, String text)  {
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/ast/impl/FootnoteImpl.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/ast/impl/FootnoteImpl.java
@@ -1,0 +1,45 @@
+package org.asciidoctor.jruby.ast.impl;
+
+import org.jruby.RubyStruct;
+
+import org.asciidoctor.ast.Footnote;
+import org.jruby.javasupport.JavaEmbedUtils;
+
+public class FootnoteImpl implements Footnote {
+
+    private static final String INDEX_KEY_NAME = "index";
+    private static final String ID_KEY_NAME = "id";
+    private static final String TEXT_KEY_NAME = "text";
+
+    private Long index;
+    private String id;
+    private String text;
+
+    private static Object aref(RubyStruct s, String key)  {
+       return JavaEmbedUtils.rubyToJava(s.aref(JavaEmbedUtils.javaToRuby(s.getRuntime(), key)));
+    }
+
+    public static Footnote getInstance(Long index, String id, String text)  {
+        FootnoteImpl footnote = new FootnoteImpl();
+        footnote.index = index;
+        footnote.id = id;
+        footnote.text = text;
+        return footnote;
+    }
+
+    public static Footnote getInstance(RubyStruct rubyFootnote) {
+        return FootnoteImpl.getInstance(
+                (Long) aref(rubyFootnote, INDEX_KEY_NAME),
+                (String) aref(rubyFootnote, ID_KEY_NAME),
+                (String) aref(rubyFootnote, TEXT_KEY_NAME));
+    }
+
+    @Override
+    public Long getIndex() { return index; }
+
+    @Override
+    public String getId() { return id; }
+
+    @Override
+    public String getText() { return text; }
+}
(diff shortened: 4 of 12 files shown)


asciidoctorj-api/src/main/java/org/asciidoctor/ast/Catalog.java
```diff
@@ -0,0 +1,18 @@
+package org.asciidoctor.ast;
+
+import java.util.List;
+
+/**
+ * The asciidoctor catalog contains data collected by asciidoctor that is useful to a converter.
+ * The catalog is considered an internal structure by asciidoctor, and subject to change.
```
abelsromero, line 7: Not sure we should state the second line. Maybe causes consufions in API users.

asciidoctorj-core/src/main/java/org/asciidoctor/jruby/ast/impl/FootnoteImpl.java
```diff
@@ -0,0 +1,45 @@
+package org.asciidoctor.jruby.ast.impl;
+
+import org.jruby.RubyStruct;
+
+import org.asciidoctor.ast.Footnote;
+import org.jruby.javasupport.JavaEmbedUtils;
+
+public class FootnoteImpl implements Footnote {
```
abelsromero, line 8: Given the interface only allows getters, we could make this immutable with "final" attributes and a public "all args" constructor. That's assuming we don't want to expose setters, which I honestly don't know if we should or should not.

abelsromero: Looking into more detail I see this deviates from other implementations because we are not keeping the reference with the rubyObject, nor we are using `RubyObjectWrapper` as base class to hande conversion.
The implementation works, but won't be able to add setters in the future and debugging may be a bit harder.

abelsromero:
Added couple of comments, but @robertpanzer has the final saying.

lread:
> Added couple of comments, but @robertpanzer has the final saying.

Thanks so much for your thoughtful review @abelsromero!

asciidoctorj-api/src/main/java/org/asciidoctor/ast/Catalog.java
lread: Maybe you are right here. I got this intel from the issue discussion and found it valuable. Happy to adjust as needed.

asciidoctorj-api/src/main/java/org/asciidoctor/ast/Document.java
```diff
@@ -75,4 +75,7 @@
      * @param state The state in which to put the sourcemap (true = on, false = off).
      */
     void setSourcemap(boolean state);
+
+
+    Catalog getCatalog();
```
robertpanzer, line 80: It would be great to have a Javadoc here stating that this will only return sth useful when called from within a Converter (and/or after Document.getContent() was called).

asciidoctorj-api/src/main/java/org/asciidoctor/ast/Footnote.java
```diff
@@ -0,0 +1,10 @@
+package org.asciidoctor.ast;
+
+public interface Footnote {
+
+   Long getIndex();
```
robertpanzer, line 5: We might want to mention here that the index starts at 1 as it might be surprising.

asciidoctorj-api/src/main/java/org/asciidoctor/ast/Document.java
lread: Good idea, but this caveat only apply to footnotes.  
The catalog will soon also have `getRefs()` which will be available before `Document.getContent()` is called.
(I think I am right about the above, will verify/adjust javadocs as I add refs).

asciidoctorj-api/src/main/java/org/asciidoctor/ast/Footnote.java
lread: Will do, tx!

lread:
Thanks for comments @robertpanzer, if the general approach agrees with you and @abelsromero, I will continue with adding refs on your go-ahead (and addressing any review feedback of course).

asciidoctorj-core/src/main/java/org/asciidoctor/jruby/ast/impl/FootnoteImpl.java
```diff
@@ -0,0 +1,45 @@
...
+    private static final String INDEX_KEY_NAME = "index";
+    private static final String ID_KEY_NAME = "id";
+    private static final String TEXT_KEY_NAME = "text";
+
+    private Long index;
+    private String id;
+    private String text;
+
+    private static Object aref(RubyStruct s, String key)  {
+       return JavaEmbedUtils.rubyToJava(s.aref(JavaEmbedUtils.javaToRuby(s.getRuntime(), key)));
```
robertpanzer, line 19: As we know that we need a RubyString would it make sense to directly create it instead of calling `JavaEmbedUtils.javaToRuby(s.getRuntime(), key)`?

robertpanzer:
Yes, please go ahead.
This is great! :+1:

asciidoctorj-core/src/main/java/org/asciidoctor/jruby/ast/impl/FootnoteImpl.java
lread: Oh! Maybe! I think you are suggesting the constants could be RubyStrings?  And the aref signature would be `aref(RubyStruct s, RubyString key)`.  If so, ya, I like it!

asciidoctorj-api/src/main/java/org/asciidoctor/ast/Catalog.java
lread: You know what? I think you are right. If asciidoctor docs/code stated this somewhere I could repeat it, but I've not yet found it stating this.  I'll turf the 2nd line.

abelsromero: Or as @robertpanzer suggested, maybe it makes more sense in the getter.
It just seems to me the class description is not the place for internal details, but what do I know :man_shrugging:

asciidoctorj-core/src/main/java/org/asciidoctor/jruby/ast/impl/FootnoteImpl.java
robertpanzer: I could stay `private static Object aref(RubyStruct s, String key)`, but instead of calling `JavaEmbedUtils.javaToRuby(s.getRuntime(), key)`to get the key, we could directly create the RubyString from the correct RubyRuntime with `s.getRuntime().newString(key)`.
Because I guess we should be careful to make sure that all objects belong to the same Ruby instance.

lread: Oh right... coolio! Will do!

asciidoctorj-api/src/main/java/org/asciidoctor/ast/Catalog.java
robertpanzer: That's indeed a difficult question and I am probably wrong most of the time😂

What I find good is to mention that this is subject to change if it's also not directly part of the public API of asciidoctor.

lread: Oh I kind of missed that getter point.  Ya, makes sense, I think.  
I'll move it but also get rid of the 2nd line.

asciidoctorj-core/src/test/groovy/org/asciidoctor/converter/WhenFootnotesAreUsed.groovy
```diff
@@ -0,0 +1,116 @@
...
+ * Note that it is a current limitation of asciidoctor that footnotes are not available until
+ * after they have been converted for the document.
+ */
+@RunWith(ArquillianSputnik)
+class WhenFootnotesAreUsed extends Specification {
+    static final String CONVERTER_BACKEND = 'footnote'
+
+    @ArquillianResource
+    private Asciidoctor asciidoctor
+    private static List<Footnote> footnotesBeforeConvert
```
abelsromero, line 32: To make test-methods fully indepent from eachother we could convert these into class attributes of the Converter and register a different instance with something like:

```java
asciidoctor.javaConverterRegistry().register(new Converter(...), CONVERTER_BACKEND)
```

asciidoctorj-api/src/main/java/org/asciidoctor/ast/Catalog.java
lread: Ok @robertpanzer, I'll use that wording then.

asciidoctorj-core/src/test/groovy/org/asciidoctor/converter/WhenFootnotesAreUsed.groovy
lread: I don't see a `register` that takes an instance, only one that takes a class.

I suppose I could `unregisterAll` as part of a test tear down.

abelsromero: Omg is true! I assumes because extensions have this options.
Then just ignore the comment, no need to overcomplicate things and unregistering won't fix concurrency issues.

We may need to address this in another PR.

lread: No problemo, thanks for the continued comments and reviews, I really appreciate them.

lread:
Ok, @abelsromero & @robertpanzer, I've added access for doc catalog refs.

I think my changes to RubyHashMapDecorator make good sense, but I'm sure you'll let me know if you prefer an alternate approach. 

Otherwise, looking forward to any feedback you may have.

robertpanzer:
This looks great! Sorry for not responding the last days.
I'll try to test it tomorrow, in particular the changes around the RubyHashMapDecorator.

lread:
Cool, thanks for the update/status!

asciidoctorj-core/src/main/java/org/asciidoctor/jruby/internal/RubyHashMapDecorator.java
```diff
@@ -22,9 +22,35 @@
...
+    public RubyHashMapDecorator(RubyHash rubyHash, Class rubyKeyType) {
         this.rubyRuntime = rubyHash.getRuntime();
         this.rubyHash = rubyHash;
+        if (rubyKeyType != RubySymbol.class && rubyKeyType != String.class) {
+            throw new UnsupportedOperationException("key type must be either RubySymbol or String");
+        }
+        this.rubyKeyType = rubyKeyType;
+    }
+
+    private Object coerceKey(Object key) {
```
robertpanzer, line 49: Is there anything speaking against making `key` a `String`?

asciidoctorj-api/src/main/java/org/asciidoctor/ast/Catalog.java
```diff
@@ -0,0 +1,32 @@
...
+     * In this case the id would map to a {@link Section} element.
+     *
+     * Ids can also be explicitly assigned by document authors to any document element.
+     * See https://asciidoctor.org/docs/user-manual/#id
+     *
+     * A converter might use cataloged refs to lookup ids to support rendering inline anchors.
+     *
+     * @return a map of ids to elements that asciidoctor has collected from the document.
+     */
+    Map<String, Object> getRefs();
```
robertpanzer, line 31: Could this be a `Map<String, ContentNode>`?
Or is it possible that this contains sth else than an AST node?

robertpanzer:
Overall this looks great.
I just have 2 minor remarks.

asciidoctorj-core/src/main/java/org/asciidoctor/jruby/internal/RubyHashMapDecorator.java
lread: Right you are! I think the method signature could and should be `private Object coerceKey(String key)`

asciidoctorj-api/src/main/java/org/asciidoctor/ast/Catalog.java
lread: That is a good question that I don't know the answer to.  

In my experiments and testing I did not see a case where a ref id points to anything but a `ContentNode`, but I do see allowances for it possibly being something else in the [html5 Ruby converter](https://github.com/asciidoctor/asciidoctor/blob/4f155298308b646f3327619f6b0dfdbe52124366/lib/asciidoctor/converter/html5.rb#L1115), so I cautiously left it as an `Object`.

lread:
Afterthought, do any supporting docs need updating?

asciidoctorj-core/src/main/java/org/asciidoctor/jruby/internal/RubyHashMapDecorator.java
lread: Oh... I remember now, I left it as `Object` to avoid casting from all the callers.  

Lemme know what you prefer.  I have no strong opinion.

asciidoctorj-api/src/main/java/org/asciidoctor/ast/Catalog.java
robertpanzer: I see, then it is probably safer to keep it as Object for now.

asciidoctorj-core/src/main/java/org/asciidoctor/jruby/internal/RubyHashMapDecorator.java
robertpanzer: I found it easier to reason about this method when I saw that the key is a String. (Would have been nice if Java had a variant type, then the signature would have been even clearer, but that doesn't exist afaik)
But I guess this is fine for now a the method is private.

robertpanzer:
It should probably be also mentioned in the integrator guide.
If you like we can do this later and I can merge now.
Or do you want to add it to this PR?

asciidoctorj-core/src/main/java/org/asciidoctor/jruby/internal/RubyHashMapDecorator.java
lread: Hey, it is easy for me to switch it to String. I'll go ahead and do so.

asciidoctorj-api/src/main/java/org/asciidoctor/ast/Catalog.java
lread: I think so too.

lread:
I'd better do the docs with the code, else I might never do the docs!

lread:
Ok @robertpanzer, took a stab at a small mention of the catalog in the integrator's guide, lemme know if it is sufficient.

asciidoctorj-api/src/main/java/org/asciidoctor/ast/Catalog.java
robertpanzer: Checked the code in Asciidoctor core again. I agree, if Asciidoctor core is not sure whether it is an AbstractNode or not, we shouldn't be either.

robertpanzer:
Thank you putting so much effort into this! This is great.
Thank you!

PR #979, opened 2020-12-02
Add test case with inlineMacro with substitutions

jmini:
## Kind of change

- [ ] Bug fix
- [ ] New breaking feature
- [ ] Documentation update
- [ ] Build improvement
- [x] Additional test case

## Description

If you want that to apply the substitutions in an Inline macro, you need to use following attribute:

* `key`: `"subs"`
* `value`: `":normal"`

See this [StackOverflow question](https://stackoverflow.com/questions/65017746/how-to-have-the-returned-inline-being-formatted-in-an-asciidoctorj-inlinemacropr) and the related discussion with @mojavelinux on [twitter](https://twitter.com/mojavelinux/status/1331935753593491456).

This PR adds a test case to ensure this approach is working (as guaranty there will be no regression in the future).

This correspond to this reference test in the ruby asciidoctor project:
https://github.com/asciidoctor/asciidoctor/blob/master/test/extensions_test.rb#L1415-L1431

On the long term, some convenience could be added to the attribute builder to set substitutions.

### DIFF
--- a/asciidoctorj-core/src/test/java/org/asciidoctor/extension/SayMacro.java
+++ b/asciidoctorj-core/src/test/java/org/asciidoctor/extension/SayMacro.java
@@ -1,0 +1,27 @@
+package org.asciidoctor.extension;
+
+import org.asciidoctor.ast.ContentNode;
+import org.asciidoctor.ast.StructuralNode;
+
+import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


robertpanzer:
Thank you! 👍

PR #984, opened 2021-01-10
Added support of toclevels attribute

l0wlik34G6:
## Kind of change

- [ ] Bug fix
- [ ] New breaking feature
- [ ] Documentation update
- [ ] Build improvement

## Description

To enable support for toclevels attribute, which defines how many levels should be displayed in the toc

How does it achieve that?
By adding the constant and the setter for this functionality

I guess not

No

## Release notes

Added support of `toclevels` attribute


robertpanzer:
Thank you! Sorry for taking so long to review.

PR #989, opened 2021-02-07
Release 2.4.3

robertpanzer:
This PR hopefully completely fixes the publishing to Maven Central.
With these changes I have published AsciidoctorJ 2.4.3 to a staging repository, that means it is not really published yet, but can be tested by everyone first.

To use this version add the following repository to your Maven config (every publish gets its own repo):
```xml
    <repositories>
        <repository>
            <id>staging</id>
            <url>https://oss.sonatype.org/content/repositories/orgasciidoctor-1240</url>
            <releases>
                <enabled>true</enabled>
            </releases>
        </repository>
    </repositories>
```

Before merging it would be great if somebody else could also check this. 

One point where I need help is in asciidoctorj-distribution/build.gradle.
In that file the build has to add the result of the distZip task to the publication configured in the asciidoctorj module.
Right now I don't know if the configuration order is random and it just works accidentally. However I added at least some check so that the build should fail if the zip cannot be added to the publication.
That said, I wonder though if we maybe should switch to publishing distributions to Github releases

To stage a release simple run 
```
./gradlew publishToSonatype -i
```
After that you should find an open staging repository at https://oss.sonatype.org/#stagingRepositories.

The repository can be closed after the push with 
```
./gradlew closeRepository -i
```
Finally, if everything is fine it should be possible to release this to Maven central with 
```
./gradlew releaseRepository -i
```

The following properties should be set, for example in ~/.gradle/gradle.properties:
```
# Settings required for signing the artifacts
signing.keyId=***
signing.password=***
signing.secretKeyRingFile=***

# Settings required for publishing to sdkman
sdkman_consumer_key=***
sdkman_consumer_token=***

# Settings required for publishing to sonatype
sonatypeUsername=***
sonatypePassword=***
```

I'll update the README with a release description once everybody is fine with this.

### DIFF
--- a/.github/workflows/upstream-build.yaml
+++ b/.github/workflows/upstream-build.yaml
@@ -9,7 +9,7 @@ jobs:
       - uses: actions/checkout@v1
         with:
           fetch-depth: 1
-          ref: ${{ github.event.client_payload.branch }} 
+          ref: ${{ github.event.client_payload.branch }}
       - uses: actions/setup-java@v1
         with:
           java-version: '1.8'
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -13,6 +13,8 @@ For a detailed view of what has changed, refer to the {url-repo}/commits/master[
 
 == Unreleased
 
+== 2.4.3 (2021-02-12)
+
 Improvement::
 
 * Upgrade to asciidoctorj-pdf 1.5.4 (#986)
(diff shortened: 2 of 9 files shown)


abelsromero:
> ```
>    <url>https://oss.sonatype.org/content/repositories/orgasciidoctor-1240</url>
> ```

Is the number something subject to change?

robertpanzer:
Yes, for every new publish the number will be incremented. 
So if I would do some minor change and publish again you would get a new repo. 
And if this repo has some bug I can simply drop it entirely again in the UI

robertpanzer:
If everybody is happy with this release I would publish it by the end of the week.

PR #992, opened 2021-02-13
Add release description to README.

robertpanzer:
Removed documentation that is available on docs.asciidoctor.org from the README.

## Kind of change

- [ ] Bug fix
- [ ] New breaking feature
- [X] Documentation update
- [ ] Build improvement

## Description

Adding the steps to build and publish a new release to Maven central.

Additionaly I removed content from the README that is also available on the Antora site.
Imo this is too detailed for a README anyway, and duplication rather results in diverging information.

I also updated the docs for how to build in the Antora module, I wonder though if this rather only belongs in the README.

@abelsromero @mojavelinux Would be great if you could skim over this.

How does it achieve that?

## Issue

Fixes #Issue 

## Release notes

### DIFF
--- a/docs/modules/project/pages/continuous-integration.adoc
+++ b/docs/modules/project/pages/continuous-integration.adoc
@@ -3,51 +3,68 @@
 Continuous integration for the AsciidoctorJ project is performed by GitHub Actions.
 You can find recent build results, including the build status of pull requests, on the https://github.com/asciidoctor/asciidoctorj/actions[asciidoctor/asciidoctorj] page.
 
-== Publish the artifacts
+== Release and publish the artifacts
 
-Artifacts are published to Maven Central and jCenter by way of Bintray's _Distribution as a Service_ platform.
+To build and publish a release the following properties have to be set, for example in the file ~/.gradle/gradle.properties:
 
-Before publishing, you need to configure your gpg signing and Bintray credentials.
-Create the file [path]_$HOME/.gradle/gradle.properties_ and populate the following properties.
-
+.~/.gradle/gradle.properties
+[source,properties]
 ----
-signing.keyId=
-signing.password=
+signing.keyId=...              # <1>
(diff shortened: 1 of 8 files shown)


docs/modules/project/pages/continuous-integration.adoc
```diff
@@ -3,51 +3,68 @@
...
+sonatypePassword=...
+----
+<1> Settings for signing the artifacts
+<2> Credentials for publishing a new release to sdkman
+<3> Credentials for publishing artifacts to oss.sonatype.org
 
- $ ./gradlew asciidoctorj-api:artifactoryPublish asciidoctorj:artifactoryPublish asciidoctorj-distribution:artifactoryPublish
+The following steps are necessary to build a new release:
 
+1. Update the version in gradle.properties to a release version, i.e. from `2.4.4-SNAPSHOT` to `2.4.4`.
```
abelsromero, line 29: Should we rely on auto-numering? :roll_eyes:

robertpanzer: Do you mean to take the concrete version from some computed variable?

mojavelinux:
> Additionaly I removed content from the README that is also available on the Antora site.
Imo this is too detailed for a README anyway, and duplication rather results in diverging information.

Yes, that is the direction we want to head. We haven't yet completed a style and structure guide for projects, but when we do it would cover this topic. And we can always adjust if necessary when the time comes. So I vote to proceed!

docs/modules/project/pages/continuous-integration.adoc
robertpanzer: Ah, now I get what you mean!
I can change it to auto-numbering, sure. 
I personally find the source more readable if the numbers are in, but for consistency it might be better to use auto-numbering if we also apply it elsewhere.

Issue #973, opened 2020-11-10
Add SpringBoot integration test


abelsromero:
Opening this issue to discuss this in more details.

Integration problems with SpringBoot is something that has come up from time to time due to how it packages jars, so it makes sense to have a it-test to ensure, that it works, and also to have an example rady to use.

The problem here is that the issue appears when running a SpringBoot app from a fatjar. This means the test should have to build a jar, and run it as a separated process in order to test it.

This is not a big problem, but we need to see how to integrate in the gradle build. I am thinking we could have a new module `asciidoctorj-springboot-integration-test` containing a submodule `springboot-app` with the app and a junit test.
Then the test task from `asciidoctorj-springboot-integration-test` could deppend on `springboot-app`. Would that work?

Issue #820, opened 2019-05-26
problem with uber jar


JurrianFahner:
When adding the following to the pom.xml:

```xml
        <dependency>
            <groupId>org.asciidoctor</groupId>
            <artifactId>asciidoctorj</artifactId>
            <version>2.0.0</version>
        </dependency>
        <dependency>
            <groupId>org.asciidoctor</groupId>
            <artifactId>asciidoctorj-api</artifactId>
            <version>2.0.0</version>
        </dependency>
```

It will run and functionally well until the application is package as an uberjar. Which is commonly the case with spring-boot when building with `mvn package`. 

If I run the jar I get the following stacktrace:
```
Caused by: org.jruby.exceptions.LoadError: (LoadError) no such file to load -- asciidoctor
	at org.jruby.RubyKernel.require(org/jruby/RubyKernel.java:987) ~[jruby-core-9.2.7.0.jar!/:na]
	at RUBY.require(uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:54) ~[na:na]
	at RUBY.<main>(<script>:12) ~[na:na]
```

To circumvent this problem a work around is extracting the jar and run it directly.

abelsromero:
SpringBoot uses a custom packaging system that does not work well with jRuby (https://docs.spring.io/spring-boot/docs/current/gradle-plugin/reference/html/#packaging-executable-configuring-unpacking).

If you are using maven, the plugin also has the same `requiresUnpackage` option https://docs.spring.io/spring-boot/docs/current/maven-plugin/repackage-mojo.html.

Disclaimer: there's an open issue similar to these one, so it could be something else is going on https://github.com/asciidoctor/asciidoctorj/issues/666

htmfilho:
I'm facing the same issue in a SpringBoot application - htmfilho/digger#14 . I'm trying to repackage it with the following SpringBoot Maven Plugin configuration:

    <plugin>
        <groupId>org.springframework.boot</groupId>
        <artifactId>spring-boot-maven-plugin</artifactId>
        <executions>
            <execution>
                <id>repackage</id>
                <goals>
                    <goal>repackage</goal>
                </goals>
                <configuration>
                    <requiresUnpack>
                        <dependency>
                            <groupId>org.jruby</groupId>
                            <artifactId>jruby-core</artifactId>
                        </dependency>
                    </requiresUnpack>
                </configuration>
            </execution>
        </executions>
[...]

PR #994, opened 2021-02-13
Add SpringBoot integration test to showcase cp configuration

abelsromero:
## Kind of change

- [ ] Bug fix
- [ ] New breaking feature
- [x] Documentation update
- [x] Build improvement

## Description

Fixes #976 adding a test to validate and showcase SpringBoot required configuration for a Gradle project.

How does it achieve that?

Add a new module `asciidoctorj-springboot-integration-test` that contains a spring-boot rest service `springboot-app` to convert Asciidoc sources and a test that:
* Starts and intances of a SpringBoot service in the background that uses Asciidoctorj
* Asserts it's working
* To build the SpringBoot project using the snapshot asciidoctorj version `springboot-app` is added as another module to the overall build and the `assemble` task is set as dependency for the  `asciidoctorj-springboot-integration-test` `test`.

Studied a version where `springboot-app` was a fully independent project and was build from the test using a `Process`, however:
* I could see no way to inject the snapshot version of asciidoctorj
* There was a problem in my machine in which even setting the PATH in the child process (see `ProcessHelper` class) the build was executed with the default's OS javac. I suspect this still can happen when starting the service but should not be an issue in CI :crossed_fingers: .

* The test covers Gradle configuration, Maven is obviously different but uses a XML parameter with the same name.
* Not sure if it's worth it adding a guide to the docs. I am open to suggestions.
On one side SpringBoot issues is a regular question and would be nice to add something users can find easily...on the other, I'd rather not add manual tasks or duplicate docs from other sources. Also having a note in the docs is no guarantee users will check docs since this the error message and cause of it is quite missleading.

## Issue

Fixes #Issue 
#973 #820

## Release notes


robertpanzer:
This is awesome, thank you!
I guess it would be worth adding a page to the docs explaining the major roadblocks similar to what we have with wildfly: https://docs.asciidoctor.org/asciidoctorj/latest/guides/run-in-wildfly/

PR #1010, opened 2021-04-14
Expose RevisionInfo from Document interface

### LINKED ISSUE #1008: Expose RevisionInfo through Document
The revision information (revdate, revnumber, revremark [docs](https://docs.asciidoctor.org/asciidoc/latest/document/revision-information/)) is only obtainable via attributes or using `readDocumentHeader` methods with provides a `RevisionInfo` class inside `DocumentHeader`.

As discussed in #970 readDocumentHeader does not provide any advantage over `load` and in fact it calls load under-the-hood, so it makes sense to deprecate it.
However, in order to do so `readDocumentHeader` we need to provide an equivalent to `RevisionInfo` class. In this case the Ruby implementation does not provide much guidance (imho). There's olny a method to obtain `reavdate` https://github.com/asciidoctor/asciidoctor/blob/e913d4fa0cde353c2659621e9680d96db0d855c4/lib/asciidoctor/document.rb#L757, so it's up to the user to read the other properties from attributes.

So, I suggest to add `RevisionInfo get revisionInfo()` to Document and expose the three fields as a single entity in a more Java-like way. With that, we can deprecate the readHeader methods.

abelsromero:

## Kind of change

- [ ] Bug fix
- [ ] New breaking feature
- [ ] Documentation update
- [ ] Build improvement

## Description

Expose RevisionInfo from Document to be accessible from extensions.
Note this is different from how the info is exposed in Asciidoctor Ruby, but aligned with Asciidoctor.js.

How does it achieve that?

Adds `RevisionInfo getRevisionInfo()` to `Document` interface.
Uses same implementation found in `DocumentHeaderImpl`.

:point_right: Only thing to discuss is we want to deprecate the Header methods in this same PR or do it separatedly. :point_left: 
Other than that, the PR is harmless.

No.

## Issue

Fixes #Issue 

Fixes #1008 

## Release notes

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -19,6 +19,7 @@ Improvement::
 * Upgrade to jruby 9.2.17.0 (#1004)
 * Upgrade to asciidoctorj-diagram 2.1.2 (#1004)
 * Add getAuthors method to Document (#1007)
+* Add getRevisionInfo method to Document. Make `DocumentHeader` class and `readDocumentHeader` methods as deprecated (#1008)
 
 Build Improvement::
 
--- a/asciidoctorj-api/src/main/java/org/asciidoctor/Asciidoctor.java
+++ b/asciidoctorj-api/src/main/java/org/asciidoctor/Asciidoctor.java
@@ -13,12 +13,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.Reader;
 import java.io.Writer;
-import java.util.ArrayList;
-import java.util.Collection;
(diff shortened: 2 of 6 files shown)


robertpanzer:
In my opinion we can also deprecate DocumentHeader as part of this PR, as this PR also completes the replacement if I understand correctly.

abelsromero:
Updated deprecated references with alternatives.

robertpanzer:
Awesome, thank you!

PR #1014, opened 2021-04-18
Try to fix running tests in IntelliJ

robertpanzer:

## Kind of change

- [x] Bug fix
- [ ] New breaking feature
- [ ] Documentation update
- [ ] Build improvement

## Description

It seems like it is not possible anymore to run or debug a single test in IntelliJ anymore since very recently.
This PR tries to fix this.

How does it achieve that?

My last status was that IntelliJ complained about the annotation org.osgi.annotation.Export not being available.
Changing the dependency type from compileOnly to implementation seems to solve this situation.
(Don't ask me why, I don't understand why compileOnly is not good here either)

@abelsromero As you were also facing this issue, could you please test this?

## Issue

Fixes #Issue 

## Release notes

### DIFF
--- a/asciidoctorj-api/build.gradle
+++ b/asciidoctorj-api/build.gradle
@@ -1,7 +1,7 @@
 apply plugin: 'biz.aQute.bnd.builder'
 
 dependencies {
-  compileOnly "org.osgi:osgi.annotation:$osgiVersion"
+  implementation "org.osgi:osgi.annotation:$osgiVersion"
 }
 
 
--- a/asciidoctorj-api/build.gradle
+++ b/asciidoctorj-api/build.gradle
@@ -1,7 +1,8 @@
 apply plugin: 'biz.aQute.bnd.builder'
 
 dependencies {
-  implementation "org.osgi:osgi.annotation:$osgiVersion"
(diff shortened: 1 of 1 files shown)


robertpanzer:
Changing the scope of the dependency makes Gradle add a dependency with runtime scope to the pom.
I am not sure right now if this is acceptable or not, right now we don't have a dependency from asciidoctorj-api to org.osgi:osgi.annotation.

robertpanzer:
It seems like adding a testRuntimeOnly dependency on osgi is sufficient to run tests in IntelliJ.
Plus it does not have the "downside" of adding a dependency to the pom.

robertpanzer:
Merging for now since we don't change the dependencies with it.
If we still feel that we should expose an optional dependency on osgi.annotation we can still add it later.

abelsromero:
Did nothing to me :shrug: more the opposite

image

robertpanzer:
Hmmm.. is it complaining about everything in red?
The last missing piece for me to run a test was the osgi annotations.

abelsromero:
Interesting enough dependencies are ok in the Gradle pannel, but in the code complains it cannot find junit. But in any case I can stll do a "Run all project tests" and debug with that. It's really odd.

Issue #977, opened 2020-11-28
Simplify Options and Attributes creation and usage


abelsromero:
This is not a new thing, but is something I think we haven't given much thought. And while going through the docs I see the docs are convoluted because of how the code works. Its basically "You can use X,Y,Z in this ways...but it's all for the same.".

`Asciidoctor.convert` accepts `Options`, `OptionsBuilder` and `Map<String,Object>`. Same applies for `convertDirectory`, `convertFile` and `convertFiles`. This creates a lot of methods for the same and is not a bad thing on its own.
What I think requires re-working is that those method's usage is not obvious to new users, specially when using IDE autompletion. Furthermore, Options has a public normal constructor, but OptionsBuilder has a static initializer plus the `get` and `asMap` methods that return Options and Map.
Also, we have `load` and `loadFile` which only accept `Map<String,Object>`·

To make things easier and more _Java-ish_ wtyt if we...
* Make `Ascidoctor` methods only accept 1 type. I am thinking of `Options` for the sake of abstraction.
* Make Options package protected to be used only internally by the builder.
* Add public static `builder` method to Options. This will return an instance of OptionsBuillder.
I am not sure if we should have a method that accepts map, one for free string-value for sure, accepting a map seems something out of convinience more than good design.
* Rename `get` in OptionsBuilder to `build` following common Java conventions.
* Make Options.map method package protected so it can only be used internally.
* Overall remove use of maps from public components.

This should be marked for a 3.0 release since it may break integrations.

PS: same applies for `Attributes`, `AttributesBuilder`...

robertpanzer:
Totally agree!

abelsromero:
Given the scope, maybe we should mark this as branking change and 3.x.x release.

robertpanzer:
Yes, this is definitely a breaking change.
But we can already do a few things right now to guide users to the new route and make them aware that things might change in the future:
- Deprecate the constructors of Options and Attributes.
- Add methods OptionsBuilder.build() and AttributesBuilder.build()
- Add methods OptionsBuilder.builder() and AttributesBuilder.builder().
- Deprecate static methods OptionsBuilder.options() and AttributesBuilder.attributes().
- Deprecate methods OptionsBuilder/AttributesBuilder.asMap() & .get().

Wdyt?

abelsromero:
> * Add methods OptionsBuilder.builder() and AttributesBuilder.builder().

[...]

PR #1015, opened 2021-04-18
Simplify Options and Attributes creation

abelsromero:
Make compatible changes to guide users for future changes

## Kind of change

- [ ] Bug fix
- [x] New breaking feature
- [ ] Documentation update
- [ ] Build improvement

## Description

Make the experience interacting with Asciidoctor API simpler and less confusing, while mantaining compatibility with current methods and types.

How does it achieve that?
Marks non-prefered methods as deprecated, so users can strt applying changes.

In terms of making the changes smooth no.

No

## Issue

Related to #977  but does not close it completely.

## Release notes

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -21,6 +21,7 @@ Improvement::
 * Add getRevisionInfo method to Document. Make `DocumentHeader` class and `readDocumentHeader` methods as deprecated (@abelsromero) (#1008)
 * Add getAuthors method to Document (@abelsromero) (#1007)
 * Upgrade to asciidoctor 2.0.13 (#1006)
+* Deprecated methods Asciidoctor, Options and Attributes API scheduled for future removal (@abelsromero)(#1015) 
 
 Build Improvement::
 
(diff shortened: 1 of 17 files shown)


abelsromero:
I must say, IMHO, that with this changes the interaction is much nicer to the point that I think most new users should be able to do a test without checking the docs.

abelsromero:
Don't merge yet, updating docs.

abelsromero:
This reduces the number of convert methods from 21 to 7 :smile:

abelsromero:
I saw for load we only accept Maps https://github.com/asciidoctor/asciidoctorj/blob/6523da86a6d92a8107c9182bdd3236c1b17b2c47/asciidoctorj-api/src/main/java/org/asciidoctor/Asciidoctor.java#L520-L536. I preffer to deal with that in a separate PR after this since it will require some real code and more tests.

robertpanzer:
Thank you for this, this is great!
I'd like to make a new release soon including this.
Imo these changes justify a new minor release, i.e. 2.5.0.
What do you think?

robertpanzer:
> I preffer to deal with that in a separate PR after this since it will require some real code and more tests.

...and add a PR that makes Asciidoctor.load() accept an instance of `Options`.

abelsromero:
> ...and add a PR that makes Asciidoctor.load() accept an instance of `Options`.

Exactly 😄 just let me add that later today and we should be good to go!

PR #1059, opened 2021-10-18
change default branch name for core in test upstream script

mojavelinux:
## Kind of change

- [x] Build improvement

## Description

The default branch name of Asciidoctor core has been renamed to main. This patch updates the test upstream script to use it as the fallback branch name.


robertpanzer:
Thanks 🙏

PR #1069, opened 2021-12-07
Fixes #1068. Clarify execution order of extensions

### LINKED ISSUE #1068: Description of conversion process
On this page: 
https://docs.asciidoctor.org/asciidoctorj/latest/extensions/conversion-process-overview/ (issue 2.5)
it says that 
"Block macro, inline macro, and block processors are called during processing in the order that they appear in the document."
I don't think this is true. When I test it (v2.5.2) by doing prinltns, the sequence is: block macro and block processors interleaved in the order that they appear in the document, then inlines in the order they appear in the document.
I only tried a single file-I don't know if it is affected by includes.

The same page says "The different extension types are called in different steps of the conversion process in the order shown in the table. Within each type:"
I have no idea what this means. Type? Table?

robertpanzer:

## Kind of change

- [ ] Bug fix
- [ ] New breaking feature
- [x] Documentation update
- [ ] Build improvement

## Description

This PR clarifies the documentation wrt the execution order of extensions.

### DIFF
--- a/docs/modules/extensions/pages/conversion-process-overview.adoc
+++ b/docs/modules/extensions/pages/conversion-process-overview.adoc
@@ -13,10 +13,10 @@ In reality, Asciidoctor has implementation details that divert from the 3 steps 
 
 The different extension types are called in different steps of the conversion process in the following order:
 
-. IncludeProcessors are called in an arbitrary and changeable order during processing and are called whenever an `include::` is found.
-. Preprocessors are called just before parsing.
-. BlockMacroProcessors and BlockProcessors are called during processing in the order that they appear in the document.
-. Treeprocessors are called right before processing.
-. InlineMacroProcessors are called during processing in the order that they appear in the document.
+. Preprocessors are called just before parsing the document.
+. IncludeProcessors are called whenever an `include::` directive is found while reading the Asciidoc source.
+. BlockMacroProcessors and BlockProcessors are called while parsing in the order that they appear in the source document.
+. Treeprocessors are called after the document has been completely parsed into the Document tree right before processing.
+. InlineMacroProcessors are called during output generation in the order that they appear in the document.
+. DocinfoProcessors are called before or after the InlineMacroProcessors based on their location, i.e., header or footer.
 . Postprocessors are called before output generation.
-. DocinfoProcessors are called in an arbitrary and changeable order during processing.
--- a/docs/modules/extensions/pages/conversion-process-overview.adoc
+++ b/docs/modules/extensions/pages/conversion-process-overview.adoc
@@ -15,8 +15,7 @@ The different extension types are called in different steps of the conversion pr
 
 . IncludeProcessors are called in an arbitrary and changeable order during processing and are called whenever an `include::` is found.
 . Preprocessors are called just before parsing.
-. BlockMacroProcessors are called during processing in the order that they appear in the document.
-. BlockProcessors are called during processing in the order that they appear in the document.
+. BlockMacroProcessors and BlockProcessors are called during processing in the order that they appear in the document.
 . Treeprocessors are called right before processing.
 . InlineMacroProcessors are called during processing in the order that they appear in the document.
 . Postprocessors are called before output generation.
--- a/docs/modules/extensions/pages/conversion-process-overview.adoc
+++ b/docs/modules/extensions/pages/conversion-process-overview.adoc
@@ -13,10 +13,11 @@ In reality, Asciidoctor has implementation details that divert from the 3 steps 
 
 The different extension types are called in different steps of the conversion process in the following order:
 
-. Preprocessors are called just before parsing the document.
-. IncludeProcessors are called whenever an `include::` directive is found while reading the Asciidoc source.
+. Preprocessors are called when the parser requires the AsciiDoc source to parse.
+. IncludeProcessors are called whenever an `include::` directive is found while reading the AsciiDoc source.
 . BlockMacroProcessors and BlockProcessors are called while parsing in the order that they appear in the source document.
-. Treeprocessors are called after the document has been completely parsed into the Document tree right before processing.
+. TreeProcessors are called after the document has been completely parsed into the Document tree right before processing.
 . InlineMacroProcessors are called during output generation in the order that they appear in the document.
-. DocinfoProcessors are called before or after the InlineMacroProcessors based on their location, i.e., header or footer.
-. Postprocessors are called before output generation.
+. DocinfoProcessors are called at the beginning of output generation if they add content to the header.
+  And they are called at the end of output generation if they add content to the footer.
+. Postprocessors are called after output generation before the content is written to the destination.
--- a/docs/modules/extensions/pages/conversion-process-overview.adoc
+++ b/docs/modules/extensions/pages/conversion-process-overview.adoc
@@ -6,7 +6,7 @@ As any language processing tool, the process can be roughly split into three ste
 . Parsing: the raw sources content is read and analyzed to generate the internal representation, the AST (_abstract syntax tree_).
 . Processing: the AST is processed. For example to detect possible errors, add automatically generated content (toc), etc.
 . Output generation: once the final AST is set, it's again processed to generate the desired output.
-For example, a sub-section of the AST representing a title with a paragraph will be converted into it's correspondent HTML or PDF output.
+For example, a sub-section of the AST representing a title with a paragraph will be converted into its correspondent HTML or PDF output.
 
 NOTE: Some liberty is taken to make the process easier to understand.
 In reality, Asciidoctor has implementation details that divert from the 3 steps above.
--- a/docs/modules/extensions/pages/conversion-process-overview.adoc
+++ b/docs/modules/extensions/pages/conversion-process-overview.adoc
@@ -11,12 +11,13 @@ For example, a sub-section of the AST representing a title with a paragraph will
 NOTE: Some liberty is taken to make the process easier to understand.
 In reality, Asciidoctor has implementation details that divert from the 3 steps above.
 
-The different extension types are called in different steps of the conversion process in the order shown in the table.
-Within each type:
+The different extension types are called in different steps of the conversion process in the following order:
 
-* Prepocessors is called just before parsing.
-* Treeprocessors is called right before processing.
-* Block macro, inline macro, and block processors are called during processing in the order that they appear in the document.
-* Postprocessors are called before output generation.
-* Include and DocinfoProcessors are called in an arbitrary and changeable order during processing.
-In the case of Include processors, it is called whenever an `include::` is found.
+. IncludeProcessors are called in an arbitrary and changeable order during processing and are called whenever an `include::` is found.
+. Preprocessors are called just before parsing.
+. BlockMacroProcessors are called during processing in the order that they appear in the document.
+. BlockProcessors are called during processing in the order that they appear in the document.
+. Treeprocessors are called right before processing.
+. InlineMacroProcessors are called during processing in the order that they appear in the document.
+. Postprocessors  are called before output generation.
+. DocinfoProcessors are called in an arbitrary and changeable order during processing.
--- a/docs/modules/extensions/pages/conversion-process-overview.adoc
+++ b/docs/modules/extensions/pages/conversion-process-overview.adoc
@@ -19,5 +19,5 @@ The different extension types are called in different steps of the conversion pr
 . BlockProcessors are called during processing in the order that they appear in the document.
 . Treeprocessors are called right before processing.
 . InlineMacroProcessors are called during processing in the order that they appear in the document.
-. Postprocessors  are called before output generation.
+. Postprocessors are called before output generation.
 . DocinfoProcessors are called in an arbitrary and changeable order during processing.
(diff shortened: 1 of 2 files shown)


docs/modules/extensions/pages/conversion-process-overview.adoc
```diff
@@ -11,12 +11,13 @@ For example, a sub-section of the AST representing a title with a paragraph will
...
-* Postprocessors are called before output generation.
-* Include and DocinfoProcessors are called in an arbitrary and changeable order during processing.
-In the case of Include processors, it is called whenever an `include::` is found.
+. IncludeProcessors are called in an arbitrary and changeable order during processing and are called whenever an `include::` is found.
+. Preprocessors are called just before parsing.
+. BlockMacroProcessors are called during processing in the order that they appear in the document.
+. BlockProcessors are called during processing in the order that they appear in the document.
+. Treeprocessors are called right before processing.
+. InlineMacroProcessors are called during processing in the order that they appear in the document.
+. Postprocessors  are called before output generation.
```
abelsromero, line 22: Silly double space typo, sorry :nerd_face:

robertpanzer: Hah :D If it's only that

abelsromero (approved):

abelsromero:
I saw a previous typo `it’s correspondent HTML or PDF output`, should not use 's` maybe we can take the oportunity to fix it.

docs/modules/extensions/pages/conversion-process-overview.adoc
```diff
@@ -6,17 +6,18 @@ As any language processing tool, the process can be roughly split into three ste
 . Parsing: the raw sources content is read and analyzed to generate the internal representation, the AST (_abstract syntax tree_).
 . Processing: the AST is processed. For example to detect possible errors, add automatically generated content (toc), etc.
 . Output generation: once the final AST is set, it's again processed to generate the desired output.
-For example, a sub-section of the AST representing a title with a paragraph will be converted into it's correspondent HTML or PDF output.
+For example, a sub-section of the AST representing a title with a paragraph will be converted into its correspondent HTML or PDF output.
```
robertpanzer, line 9: Did you mean this, @abelsromero ?

abelsromero: yes! thank a lot

docs/modules/extensions/pages/conversion-process-overview.adoc
rockyallen: Hope you dont mind me jumping in? The critical information for me is that BlockProcessors and BlockMacroProcessors are interleaved in document order, then InlineMacroProcessors are run. Your revision says nothing about the relative ordering of the three types.

mojavelinux: Agreed. It should be emphasized that inline macro processors are called in a different phase, the convert phase, though still in document order.

docs/modules/extensions/pages/conversion-process-overview.adoc
```diff
@@ -6,17 +6,18 @@ As any language processing tool, the process can be roughly split into three ste
...
-Within each type:
+The different extension types are called in different steps of the conversion process in the following order:
 
-* Prepocessors is called just before parsing.
-* Treeprocessors is called right before processing.
-* Block macro, inline macro, and block processors are called during processing in the order that they appear in the document.
-* Postprocessors are called before output generation.
-* Include and DocinfoProcessors are called in an arbitrary and changeable order during processing.
-In the case of Include processors, it is called whenever an `include::` is found.
+. IncludeProcessors are called in an arbitrary and changeable order during processing and are called whenever an `include::` is found.
```
mojavelinux, line 16: I don't think we should say that IncludeProcessors are called in an arbitrary and changeable order. They are called when an include directive is encountered. That's all it needs to say.

docs/modules/extensions/pages/conversion-process-overview.adoc
```diff
@@ -6,17 +6,17 @@ As any language processing tool, the process can be roughly split into three ste
...
+. Preprocessors are called just before parsing.
+. BlockMacroProcessors and BlockProcessors are called during processing in the order that they appear in the document.
```
robertpanzer, line 18: @rockyallen Does this make it clearer for you?

docs/modules/extensions/pages/conversion-process-overview.adoc
abelsromero: > Agreed. It should be emphasized that inline macro processors are called in a different phase, the convert phase, though still in document order.

That's right, back when I wrote that I did not have the insight I have now. I tried to make an analogy with standard compilers with parsing, processing & convert phases and that's not accurate. I assume the correct would be to talk about preprocessing, parsing and converting?

For context, and to see if we should redirect the explanation...when I wrote this I wanted to answer some user questions I've seen and I myself have:
* Do I have acces to the AST and can I manipulate it the this extension?
* Can I output formatted content (e.g HTML or text) from this extension?
* Why if inline macro processor is run during convert can I inject AST nodes? Like in example https://docs.asciidoctor.org/asciidoctorj/latest/extensions/inline-macro-processor/

But going for a "yes/no" seemed too complex, so I mapped the extensions to the phases explaining what info is available in each phase.

docs/modules/extensions/pages/conversion-process-overview.adoc
robertpanzer: I am just testing with this and it seems like IncludeProcessors are called in order when the corresponding Asciidoc is read.

docs/modules/extensions/pages/conversion-process-overview.adoc
```diff
@@ -6,17 +6,17 @@ As any language processing tool, the process can be roughly split into three ste
...
+The different extension types are called in different steps of the conversion process in the following order:
 
-* Prepocessors is called just before parsing.
-* Treeprocessors is called right before processing.
-* Block macro, inline macro, and block processors are called during processing in the order that they appear in the document.
-* Postprocessors are called before output generation.
-* Include and DocinfoProcessors are called in an arbitrary and changeable order during processing.
-In the case of Include processors, it is called whenever an `include::` is found.
+. IncludeProcessors are called in an arbitrary and changeable order during processing and are called whenever an `include::` is found.
+. Preprocessors are called just before parsing.
```
robertpanzer, line 17: This seems to be wrong, Preprocessors seem to be called before an IncludeProcessor.

docs/modules/extensions/pages/conversion-process-overview.adoc
```diff
@@ -6,17 +6,17 @@ As any language processing tool, the process can be roughly split into three ste
...
+. BlockMacroProcessors and BlockProcessors are called during processing in the order that they appear in the document.
+. Treeprocessors are called right before processing.
+. InlineMacroProcessors are called during processing in the order that they appear in the document.
+. Postprocessors are called before output generation.
```
robertpanzer, line 21: This should go last in the list, at least for the default HTML converter Docinfoprocessors are called before. For the header they are called before the inline macros, for the footer they are called after the inline macros.

robertpanzer:
@mojavelinux Merry Christmas to you and the entire Asciidoctor community!
I tested a bit with the execution order of extensions and also trying to figure out the correct order from the Asciidoctor sources.
Can you check if this is clear enough and correct to help users understand the execution order, but also not to artificially restrict implementation details. 
(Some might argue that anything mentioned here is an API contract that must never change in the future without a major release, and I'd like to avoid navigating us into such a situation.)

docs/modules/extensions/pages/conversion-process-overview.adoc
```diff
@@ -6,17 +6,17 @@ As any language processing tool, the process can be roughly split into three ste
...
-Within each type:
+The different extension types are called in different steps of the conversion process in the following order:
 
-* Prepocessors is called just before parsing.
-* Treeprocessors is called right before processing.
-* Block macro, inline macro, and block processors are called during processing in the order that they appear in the document.
-* Postprocessors are called before output generation.
-* Include and DocinfoProcessors are called in an arbitrary and changeable order during processing.
-In the case of Include processors, it is called whenever an `include::` is found.
+. Preprocessors are called just before parsing the document.
```
mojavelinux, line 16: The best way to phrase this is that preprocessors are called in front of parsing the lines of the document. The preprocessor visits the line first, then the parser touches it. But they work together.

docs/modules/extensions/pages/conversion-process-overview.adoc
```diff
@@ -6,17 +6,17 @@ As any language processing tool, the process can be roughly split into three ste
...
+. IncludeProcessors are called whenever an `include::` directive is found while reading the Asciidoc source.
```
mojavelinux, line 17: Please capitalize "Asciidoc" as "AsciiDoc"

docs/modules/extensions/pages/conversion-process-overview.adoc
```diff
@@ -6,17 +6,17 @@ As any language processing tool, the process can be roughly split into three ste
...
+. BlockMacroProcessors and BlockProcessors are called while parsing in the order that they appear in the source document.
+. Treeprocessors are called after the document has been completely parsed into the Document tree right before processing.
```
mojavelinux, line 19: TreeProcessors (note the capitalization) are called after the blocks of the documents have been completed parsed. The document isn't technically fully parsed at this stage.

docs/modules/extensions/pages/conversion-process-overview.adoc
```diff
@@ -6,17 +6,17 @@ As any language processing tool, the process can be roughly split into three ste
...
+. InlineMacroProcessors are called during output generation in the order that they appear in the document.
+. DocinfoProcessors are called before or after the InlineMacroProcessors based on their location, i.e., header or footer.
```
mojavelinux, line 21: DocinfoProcessors are called during output generation. They are either called at the start of the output generation (head and header) or at the end of the output generation (footer).

docs/modules/extensions/pages/conversion-process-overview.adoc
```diff
@@ -6,17 +6,17 @@ As any language processing tool, the process can be roughly split into three ste
...
+. Postprocessors are called before output generation.
```
mojavelinux, line 22: Postprocessors are called after the output has been generated, but before it is written to any output stream / file.

mojavelinux:
I left some comments for clarification. The order is right, but the explanations still need to be more precise.

docs/modules/extensions/pages/conversion-process-overview.adoc
robertpanzer: Looking at the code trying to understand the relation between Reader, Parser and Preprocessor I guess we might have missed one kind of preprocessors here:
Currently AsciidoctorJ's preprocessors don't return anything, so they have to do all their work on the Reader that was passed to it.
But it seems like it's also possible to return a new Reader that is then used by the Parser to get the content.

Fixing this would be a breaking change though. I'll create a ticket for this.

docs/modules/extensions/pages/conversion-process-overview.adoc
robertpanzer: I've used Treeprocessors here since the superclass is called `org.asciidoctor.extension.Treeprocessor`. 
I'll update it here.

docs/modules/extensions/pages/conversion-process-overview.adoc
robertpanzer: Thanks! Is the relationship to the InlineMacroProcessors correct? That's at least what I've observed in my test.

docs/modules/extensions/pages/conversion-process-overview.adoc
robertpanzer: Yeah sure. Where was my head?
Thank you!

robertpanzer:
Updated the document.

robertpanzer:
Merging this for now.
If there is further feedback we can still create a new PR.

docs/modules/extensions/pages/conversion-process-overview.adoc
mojavelinux: It's definitely a bit tricky. The reason an inline macro can add AST nodes is because parsing for inline content is happening during conversion. So it's really a "just in time" AST mutation. Going even deeper, an inline macro doesn't actually add a node to the AST. Rather, the return value is immediately converted and added into the conversion stream.

Keep in mind, this is likely going to change in the spec / Austen...which will move inline processing into the parsing phase (eventually).

docs/modules/extensions/pages/conversion-process-overview.adoc
mojavelinux: You are correct that the preprocessor is called before any lines are read, which means before any include processors. A preprocessor extension is its own phase. An include processor is more of a "pre line reader", which gets called just before the line is given to the parser.

docs/modules/extensions/pages/conversion-process-overview.adoc
mojavelinux: > Currently AsciidoctorJ's preprocessors don't return anything, so they have to do all their work on the Reader that was passed to it. But it seems like it's also possible to return a new Reader that is then used by the Parser to get the content.

That's correct.

docs/modules/extensions/pages/conversion-process-overview.adoc
mojavelinux: Yes, that looks right.

PR #1070, opened 2021-12-09
Fix example on Logs Handling API documentation

hcadavid:
The custom LogHandle given as an example in the Logs Handling API sections didn't work, as it used 'extends' rather than 'implements' to the LogHandler interface.

## Kind of change

- [ ] Bug fix
- [ ] New breaking feature
- [x] Documentation update
- [ ] Build improvement

## Description

   Fix an error on the code given in the Logs Handling API documentation section

How does it achieve that?

   Code snippet updated (extends replaced with 'implements')

   N/A 

   N/A
## Issue

Fixes #Issue 

## Release notes

### DIFF
--- a/docs/modules/ROOT/pages/logs-handling.adoc
+++ b/docs/modules/ROOT/pages/logs-handling.adoc
@@ -76,7 +76,7 @@ import org.asciidoctor.log.LogRecord;
 /**
  * Stores LogRecords in memory for later analysis.
  */
-public class MemoryLogHandler extends LogHandler {
+public class MemoryLogHandler implements LogHandler {
 
   private List<LogRecord> logRecords = new ArrayList<>();
 


robertpanzer:
Thank you!

Ideally we should also test this part of the documentation automatically, I'll try to set it up when I find the time.
Thanks for finding and fixing this!

abelsromero:
> Ideally we should also test this part of the documentation automatically, I'll try to set it up when I find the time.
> Thanks for finding and fixing this!

We should be able to use symlinks to reference sources outside of modules now, that was not possible when the docs where migrated https://gitlab.com/antora/antora/-/issues/865

mojavelinux:
Yes indeed. And the docs.asciidoctor.org build is always using the newest release of Antora.

PR #1071, opened 2021-12-28
Add documentation for logging from extensions

robertpanzer:

## Kind of change

- [ ] Bug fix
- [ ] New breaking feature
- [x] Documentation update
- [ ] Build improvement

## Description

Add documentation for how an extension can log messages that are also captured by build tools like the maven or gradle plugin.


docs/modules/ROOT/pages/asciidoctor-api-options.adoc
```diff
@@ -115,6 +115,19 @@ OptionsBuilder optionsBuilder =
             Options.builder().templateDirs(new File("templates_path"));
 ----
 
+[[sourcemap]]
+== sourcemap
+
+Keeps track of the file and line number for each parsed block. (Useful for tooling applications where the association between the converted output and the source file is important).
```
abelsromero, line 121: I think the parenthesis are a leftover?

mojavelinux:
I tested this and it works great. This is exactly what I had in mind for being able to log from an extension.

```
log(new org.asciidoctor.log.LogRecord(org.asciidoctor.log.Severity.WARN, "Hello, World!"));
```

mojavelinux:
One thing I thing we should add to the docs is how to configure JUL so that it matches the default log output of Asciidoctor Ruby. This is more of demonstration, but it gives the user a launching off point.

logging.properties
```
handlers = java.util.logging.ConsoleHandler
.level = WARNING
java.util.logging.ConsoleHandler.formatter = java.util.logging.SimpleFormatter
java.util.logging.SimpleFormatter.format = %3$
```

Calling asciidoctorj:

```
JAVA_OPTS="-Djava.util.logging.config.file=$PWD/logging.properties" asciidoctorj document.adoc
```

Let me know if you'd like a separate issue for that and I'll file it.

mojavelinux:
Keep in mind that if the default level is INFO instead of WARNING, users will see provisional messages about possible invalid xrefs...which could be valid xrefs that Asciidoctor just can't resolve.

docs/modules/ROOT/pages/asciidoctor-api-options.adoc
robertpanzer: I copied this sentence as is from the Asciidoctor.js module to make sure that we don't provide deviating definitions. :D
But yeah, I'll remove the parens.

robertpanzer:
Thanks for the feedback!
I added some notes on this to the description of the log handling API, since this is also the place that mentions that log messages are automatically forwarded to JUL.

robertpanzer:
Added a comment to the converter docs that they can also log by calling `log()`.
The current docs "just" show how to implement a string based converter that extends the abstract class StringConverter.
For arbitrary, binary formats there is also an approach based exclusively on interfaces.
We might want to document this in the future.

PR #1074, opened 2022-01-06
Upgrade to asciidoctor 2.0.17

robertpanzer:

## Kind of change

- [ ] Bug fix
- [ ] New breaking feature
- [ ] Documentation update
- [ ] Build improvement

## Description

This PR upgrades Asciidoctor to the latest and greatest version 2.0.17.
I also updated JRuby to 9.2.20.1.
I tried to upgrade it to 9.3, however the build failed with OOM errors in `org.jruby.util.ByteList#ByteList(int)`. I need to check if this is due to building on ARM64 or if it also reproduces on Intel.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -15,9 +15,11 @@ For a detailed view of what has changed, refer to the {url-repo}/commits/main[co
 
 Improvement::
 
+* Upgrade to asciidoctorj 2.0.17 (#1074)
 * Upgrade to asciidoctorj-pdf 1.6.2 (#1073)
-* Upgrade to rouge 3.27.0 (#1073)
 * Upgrade to asciidoctorj-diagram 2.2.1 (#1066) (@abelsromero) (#1065)
+* Upgrade to jruby 9.2.20.1 (#1074)
+* Upgrade to rouge 3.27.0 (#1073)
 * Upgrade CLI to use JCommander 1.81 (@abelsromero) (#1067)
 
 Build Improvement::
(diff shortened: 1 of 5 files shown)


robertpanzer:
Interestingly I don't seem to get this OOM on x86.
I'd like to stay conservative for now and stick with the older version.

I'll see over the weekend if I can somehow get hold of a heap dump and open a ticket once I have more information.

abelsromero:
> I'd like to stay conservative for now and stick with the older version.

I think that's a good call.
If it helps, I run the build several times with 2 latests (v9.3.1.0 & v9.3.2.0) on Linux + Ryzen and it run fine and same time as this branch (3 min 1~3 sec). OOM could be an OS thing maybe?

robertpanzer:
It looks like there is a problem with the MacOS specific implementation of fstat() which, for example, returns a file size of 976233465 instead of 111 bytes.
When reading such a file allocating the required memory fails of course. (It works a bit better when I run with -Xmx16000m lol)

I'll create a ticket for JRuby, but stick with the older version for now.

robertpanzer:
Created https://github.com/jruby/jruby/issues/6988 to follow up on the memory issue.

robertpanzer:
Updating jnr-posix to 3.1.15 seems to fix the OOM error.
I am still getting AccessDeniedExceptions.
It seems like JRuby support for M1 still has a few issues which seem to be fixed with 9.2.20.1.
So I assume that sticking with the older version is still the better option.

robertpanzer:
There are other issues with the permissions of files created by asciidoctorj-pdf.
These are also fixed by using jruby 9.3.2.0 (instead of jruby-complete) and jnr-posix 3.1.15 for the asciidoctorj-distribution tests.
So let's wait for that upgrade until there is a new version of JRuby.

PR #1081, opened 2022-01-24
Extend Preprocessors to return Readers

robertpanzer:
## Kind of change

- [ ] Bug fix
- [x] New breaking feature
- [ ] Documentation update
- [ ] Build improvement

## Description

The Asciidoctor API currently allows a preprocessor to return a reader.
The AsciidoctorJ API does not allow this, making certain preprocessors awkward or even impossible.
This PR wants to align the AsciidoctorJ API with the API of Asciidoctor.

How does it achieve that?

I added method to the Processor base class to create new readers.
The Preprocessor API is extended to return a Reader instead of void.
If a preprocessor returns null, or the original reader, the processor proxy returns nil to Asciidoctor.
Otherwise the new Reader is returned.

As a PoC I've implemented the FrontMatterPreprocessor (which does not create a new reader) and the HardBreaksPreprocessor (which does).

I was first considering to allow creation of new Readers directly, but of course that doesn't work in an implementation neutral way, also Reader exposes a complex API and internal state that is not easy to reimplement correctly from scratch for a simple extension.

**!!!It's a breaking change! Even recompiling will not work, so this would pull the trigger for a 3.0!!!**

## Issue

Fixes #Issue 

## Release notes


robertpanzer:
Updated the PR with an extension migration guide which documents the necessary changes for existing preprocessors, and updated the docs to let the sample Preprocessor use `Processor.newReader()` instead of `Reader.restoreLines()`.

PR #1083, opened 2022-02-13
Fixes #1082. Return empty list in DescriptionList.getBlocks() instead…

### LINKED ISSUE #1082: Failure getting `blocks` when using description lists
See https://asciidoctor.zulipchat.com/#narrow/stream/279644-users.2Fasciidoctorj/topic/Failure.20getting.20.60blocks.60.20when.20using.20description.20lists

Seems that `DescriptionListImpl.blocks` returns a `RubyArrayTwoObject`, instead of a `List<StructuralNode>`. 
This makes a bunch of processing steps more complicated and error-prone.
For example, this Treeprocessor:
```java
   public static class SolutionsProcedureTreeprocessor extends Treeprocessor {
        public SolutionsProcedureTreeprocessor() {}
        @Override
        public Document process(Document document) {
            processBlock((StructuralNode) document, 0);
            return document;
        }

        private void processBlock(StructuralNode block, int deep) {
            List<StructuralNode> blocks = block.getBlocks();
            for (int i = 0; i < blocks.size(); i++) {
                final StructuralNode currentBlock = blocks.get(i);
                if (currentBlock instanceof StructuralNode) {
                    System.out.println(" ".repeat(deep) + currentBlock.getClass());
                    processBlock(currentBlock, ++deep);
                }
            }
        }
    }
```
Fails when trying to invoke `block.getBlocks()` because the `block` is not a StructuralNode, but a `RubyArrayTwoObject`:
```
org.asciidoctor.jruby.internal.AsciidoctorCoreException: org.jruby.exceptions.NoMethodError: (NoMethodError) asciidoctor: FAILED: nav.adoc: Failed to load AsciiDoc document - undefined method `blocks' for #<Array:0xfb0dc3ba>
```
Proposal: Overwrite the `DescriptionListImpl.blocks` method so it returns a flattened list of nodes.

robertpanzer:
… of return a list of DescriptionListItems, which are not blocks

## Kind of change

- [x] Bug fix
- [ ] New breaking feature
- [ ] Documentation update
- [ ] Build improvement

## Description

#1082 reports that DescriptionList.getSections() returns a list of objects that are not StructuralNodes but DescriptionListEntries instead. (This only works due to type erasure).

How does it achieve that?

This PR tries to fix the problem by making DescriptionList.getSections() return a new empty list.
That way there will no longer be a ClassCastException when recursively traversing the AST only via getSections().

It could be possible to change the form of the AST by making DescriptionListEntry a StructuralNode that has the terms as properties. 
That might be difficult as the implementation of Asciidoctor Ruby currently treats the terms and its description as a list with 2 items: `[[ListItem Term1, ListItem Term2], ListItem Description]`.
This might be difficult to implement, leaves the question open how to handle description list entries without description and also this would be a breaking change.

Adding any nodes to the result of DescriptionList.getSections() is without any effect.

## Issue

Fixes #1082 

## Release notes

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/ast/impl/DescriptionListImpl.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/ast/impl/DescriptionListImpl.java
@@ -9,6 +9,8 @@ import org.jruby.runtime.builtin.IRubyObject;
 
 import java.util.List;
 
+import static java.util.Collections.emptyList;
+
 public class DescriptionListImpl extends StructuralNodeImpl implements DescriptionList {
 
     public DescriptionListImpl(IRubyObject delegate) {
@@ -39,7 +41,6 @@ public class DescriptionListImpl extends StructuralNodeImpl implements Descripti
 
     @Override
     public List<StructuralNode> getBlocks() {
-        RubyArray rubyBlocks = (RubyArray) getRubyProperty("blocks");
-        return new RubyBlockListDecorator<>(rubyBlocks);
+        return emptyList();
     }
 }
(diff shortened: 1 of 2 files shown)


abelsromero:
Been thinking on & off and I think it's best to keep it simple and be consistent with the fact these are not blocks and return empty.

PR #1114, opened 2022-09-05
Add --failure-level command line option to force non-zero exit code (fixes #1113)

### LINKED ISSUE #1113: CLI should exit with non-zero status when e.g. missing include files
The Ruby asciidoc CLI has option `--failure-level` which forces the CLI to exit with non-zero status code if the specified logging level (severity) is reached. See: https://github.com/asciidoctor/asciidoctor/pull/2674, https://github.com/asciidoctor/asciidoctor/issues/2003, https://github.com/asciidoctor/asciidoctor/issues/854

It would be nice to have the same feature in AsciidoctorJ CLI, too. (To e.g. catch missing included files in CI builds) 

I haven't touched AsciidoctorJ code before, but it looks like the relevant Ruby code is in https://github.com/asciidoctor/asciidoctor/blob/main/lib/asciidoctor/cli/invoker.rb , which AsciidoctorJ does not call.

Here's a quick attempt at doing what the Ruby CLI does: https://github.com/pasieronen/asciidoctorj/commit/f94d332eb7aaf468e4f1674bf650a5c3037fa095
Any thoughts on that?

pasieronen:
## Kind of change

- [ ] Bug fix
- [X] New non-breaking feature
- [ ] New breaking feature
- [ ] Documentation update
- [ ] Build improvement

## Description

The Ruby asciidoc CLI has option --failure-level which forces the CLI to exit with non-zero status code if the specified logging level (severity) is reached. See: https://github.com/asciidoctor/asciidoctor/pull/2674, https://github.com/asciidoctor/asciidoctor/issues/2003, https://github.com/asciidoctor/asciidoctor/issues/854

This PR implements the same functionality for AsciidoctorJ, pretty much the same way as the Ruby code (keeps track of highest seen log message severity, and checks that the end).

The default value is the same as in Ruby - only "fatal" severity causes non-zero exit code - which probably won't break many things (but someone running AsciidoctorJ as part of a CI build would probably want to fail on "error" severity as well, maybe even "warn"). 

## Issue

Fixes #1113

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/cli/SeverityConverter.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/cli/SeverityConverter.java
@@ -1,12 +1,17 @@
 package org.asciidoctor.jruby.cli;
 
-import com.beust.jcommander.IStringConverter;
-import org.asciidoctor.SafeMode;
+import com.beust.jcommander.converters.EnumConverter;
 import org.asciidoctor.log.Severity;
 
-public class SeverityConverter implements IStringConverter<Severity> {
+public class SeverityConverter extends EnumConverter<Severity> {
+
+    public SeverityConverter() {
+        super("--failure-level", Severity.class);
+    }
 
     @Override
-    public Severity convert(String argument) { return Severity.valueOf(argument.toUpperCase()); }
+    public Severity convert(String argument) {
+        return super.convert(argument.toUpperCase());
+    }
 
 }
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/cli/SeverityConverter.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/cli/SeverityConverter.java
@@ -1,0 +1,12 @@
+package org.asciidoctor.jruby.cli;
+
+import com.beust.jcommander.IStringConverter;
+import org.asciidoctor.SafeMode;
+import org.asciidoctor.log.Severity;
+
+public class SeverityConverter implements IStringConverter<Severity> {
+
+    @Override
+    public Severity convert(String argument) { return Severity.valueOf(argument.toUpperCase()); }
+
+}
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/internal/JRubyAsciidoctor.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/internal/JRubyAsciidoctor.java
@@ -22,7 +22,6 @@ import org.asciidoctor.jruby.log.internal.LogHandlerRegistryExecutor;
 import org.asciidoctor.jruby.syntaxhighlighter.internal.SyntaxHighlighterRegistryExecutor;
 import org.asciidoctor.log.LogHandler;
 import org.asciidoctor.log.LogRecord;
-import org.asciidoctor.log.Severity;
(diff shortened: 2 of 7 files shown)


asciidoctorj-core/src/main/java/org/asciidoctor/jruby/cli/SeverityConverter.java
```diff
@@ -0,0 +1,12 @@
+package org.asciidoctor.jruby.cli;
+
+import com.beust.jcommander.IStringConverter;
+import org.asciidoctor.SafeMode;
+import org.asciidoctor.log.Severity;
+
+public class SeverityConverter implements IStringConverter<Severity> {
+
+    @Override
+    public Severity convert(String argument) { return Severity.valueOf(argument.toUpperCase()); }
```
robertpanzer, line 10: When I pass an invalid option like `INFOX` I currently get a stacktrace back that does not show the currently supported options:
```
Exception in thread "main" java.lang.IllegalArgumentException: No enum constant org.asciidoctor.log.Severity.INFOX
	at java.base/java.lang.Enum.valueOf(Enum.java:273)
	at org.asciidoctor.log.Severity.valueOf(Severity.java:3)
	at org.asciidoctor.jruby.cli.SeverityConverter.convert(SeverityConverter.java:10)
...
```
What do you think about changing it so that you'll get just a message with the invalid value and all supported values like:
```
Could not convert severity 'INFOX'. Supported values are [DEBUG, INFO, WARN, ERROR, FATAL, UNKNOWN]
```

Nitpick: Could you please add newlines after the opening brace?

asciidoctorj-core/src/main/java/org/asciidoctor/jruby/internal/JRubyAsciidoctor.java
```diff
@@ -507,8 +510,15 @@ private RubyClass getExtensionGroupClass() {
 
     @Override
     public void log(LogRecord logRecord) {
+        if (this.maxSeverity.compareTo(logRecord.getSeverity()) < 0) {
+            this.maxSeverity = logRecord.getSeverity();
```
robertpanzer, line 514: JRubyAsciidoctor can be used in a multithreaded way when embedded into other environments like Spring Boot apps or sth similar.
In such scenarios it would be dangerous to rely on this property.
Could this be moved into a custom `org.asciidoctor.log.LogHandler` that is only added by the AsciidoctorInvoker?

robertpanzer (commented):
Thank you for your PR!
This looks great!
I added a few comments. What do you think about them?

asciidoctorj-core/src/main/java/org/asciidoctor/jruby/cli/SeverityConverter.java
pasieronen: I agree - and it turns out JCommander has `EnumConverter` class which gets us this nice error message for free.

asciidoctorj-core/src/main/java/org/asciidoctor/jruby/internal/JRubyAsciidoctor.java
pasieronen: Ok, moved this to MaxSeverityLogHandler under org.asciidoctor.jruby.cli package.

robertpanzer:
Thanks for the PR! 🎉 
I'll take care of porting this into the 2.5.x maintenance branch.

PR #1121, opened 2022-10-09
Expose Section's sectnum property

abelsromero:

## Kind of change

- [ ] Bug fix
- [ ] New breaking feature
- [ ] Documentation update
- [ ] Build improvement

## Description

Exposes [sectnum](https://github.com/asciidoctor/asciidoctor/blob/4bab183b9f3fad538f86071b2106f3b8185d3832/lib/asciidoctor/section.rb#L119) to users using the API for extensions, converters, etc.

This property is required to easily obtain the section numbering aligned with other Asciidoctor attributes. The alternative is reimplementing all the logic of level counting, secnumlevel, separators formating, etc.

How does it achieve that?

Adds 2 new methods:
* `String getSectnum()`
* `String getSectnum(String delimiter)`

to the Section interface, those simply call the Asciidoctor Ruby property (I assume property is the right name, it's a method).

While working on a converter I could not find it. `index` returns 0, and there's a lot of reimplementation to do to obtain a similar result.

Right now I am using internals `String sectnum = ((SectionImpl) node).getString("sectnum")`, which is not ideal.

1. The property in Asciidoctor is well documented as public, I am assuming this is stable and can be used, but could be wrong. @mojavelinux can you confirm? :bow: 
2. I am keeping the default behauviour when `sectnums` is not set, that is, the delimiters without numbers is returned (see test `should_return_invalid_sectnum_when_sectionNumbers_are_not_enabled`). Is odd and I am tempted to add a check and return empty string, but maybe this should be fixed upstream? //cc @mojavelinux 
3. I am also not adding support to pass the `append` parameter, to keep things simple, and also adding a suffix is something that API users can just do just outside of the method.

## Issue

n/a

## Release notes

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -22,6 +22,7 @@ Improvement::
 * Upgrade to JRuby 9.3.8.0 (#1116)
 * Upgrade to tilt 2.0.11 (#1109)
 * Upgrade to asciimath 2.0.4 (#1109)
+* Expose `sectnum` property in Section interface (#1121)
 
 Bug Fixes::
 
--- a/asciidoctorj-api/src/main/java/org/asciidoctor/ast/Section.java
+++ b/asciidoctorj-api/src/main/java/org/asciidoctor/ast/Section.java
@@ -74,4 +74,24 @@ public interface Section extends StructuralNode {
      */
     boolean isNumbered();
 
+
+    /**
+     * Get the section number for the current Section.
+     * <p>
+     * The section number is a dot-separated String that uniquely describes the position of this
+     * Section in the document. Each entry represents a level of nesting. The value of each entry is
+     * the 1-based outline number of the Section amongst its numbered sibling Sections.
+     */
+    String getSectnum();
+
+    /**
+     * Get the section number for the current Section.
+     * <p>
+     * The section number is a dot-separated String that uniquely describes the position of this
+     * Section in the document. Each entry represents a level of nesting. The value of each entry is
+     * the 1-based outline number of the Section amongst its numbered sibling Sections.
+     *
+     * @param delimiter the delimiter to separate the number for each level
+     */
+    String getSectnum(String delimiter);
 }
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/ast/impl/SectionImpl.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/ast/impl/SectionImpl.java
@@ -70,4 +70,14 @@ public class SectionImpl extends StructuralNodeImpl implements Section {
         return getBoolean("numbered");
     }
 
+    @Override
+    public String getSectnum() {
+        return getString("sectnum");
+    }
+
+    @Override
+    public String getSectnum(String delimiter) {
+        return getString("sectnum", new Object[]{delimiter});
+    }
+
 }
--- a/asciidoctorj-core/src/test/java/org/asciidoctor/jruby/ast/impl/SectionImplTest.java
+++ b/asciidoctorj-core/src/test/java/org/asciidoctor/jruby/ast/impl/SectionImplTest.java
@@ -1,0 +1,97 @@
+package org.asciidoctor.jruby.ast.impl;
+
+import org.asciidoctor.Asciidoctor;
+import org.asciidoctor.Attributes;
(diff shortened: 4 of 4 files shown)


robertpanzer:
Awesome, thank you!
It looks good to me.
I'll wait for feedback from @mojavelinux 

@abelsromero Can you shed some light on the `append` parameter? I am not sure if I understood what you meant.

abelsromero:
> @abelsromero Can you shed some light on the `append` parameter? I am not sure if I understood what you meant.

If you look at the documentation in the Ruby implementation
```
the String to append at the end of the section number
             or Boolean to indicate the delimiter should not be
             appended to the final level
```

so, it allows the behaviours:
* By default (null, not set, or true) it will appent the delimiter so we get something like "1.1." (see final dot.)
* When false, won't append delimiter -> we get "1.1"
* When set to a string, this is appended -> for example calling with "_", we get "1.1_"

I personally think (but it's subjective :smile:) these options do not add much in terms of features. Any user that requires this level of fine tunning can do so outside of the method with normal String manipulations. And if in the future this is requested, we can always add it based on a real use case.

abelsromero:
Checking the errors in test-upstream, I am not sure the reason. Locallly works fine and the CI logs show the version installed being `Version: 2.1.0.pre.alpha.0`, not sure where the later `SNAPSHOT` in same ci execution comes from.
```
inconsistent module metadata found. Descriptor: rubygems:asciidoctor:2.1.0.pre.alpha.0-SNAPSHOT Errors: bad version: expected='2.1.0-alpha.0' found='2.1.0.pre.alpha.0-SNAPSHOT'
```

robertpanzer:
I guess this behavior might be related to this:
https://twitter.com/mojavelinux/status/1575764592675811328?s=20&t=St8K5u1adGt9-pJcR9veBQ

mojavelinux:
Yep, I need to switch core back to using dot to separate the prerelease component in the version because, once again, Bundler changed its rules and our workaround no longer works.

abelsromero:
Sorry to be insistent, but:
* locally works fine (running the script or running line by line), at least in my Linux machine
* logs show the gem is installed locally and the version

So , if I understand correctly the gem plugin is doing some extra remote check that is not happenign in my machine, and rubygems returns a different version number? If so, maybe we can disable that check.

mojavelinux:
The fix has been applied upstream, so now it should return the version number as entered in version.rb.

mojavelinux:
It appears like there are other problems:

```
inconsistent module metadata found. Descriptor: rubygems:asciidoctor:2.0.18.pre-SNAPSHOT Errors: bad version: expected='2.0.18.pre' found='2.0.18.pre-SNAPSHOT'
```

I don't know what the source of this check is, but it is clearly choking on this problem I alluded to in gem versioning. RubyGems doesn't allow the use of a hyphen, but semver doesn't allow a use of the hyphen after the third dot. It might be necessary to put some hacks somewhere in this build to deal with this incompatibility. The quick thing to do would be to change the version in version.rb to `2.0.18` after cloning.

abelsromero:
> I don't know what the source of this check is

gem-maven-plugin is doing it. The '-SNAPSHOT' is added to the pom descriptor's `<version>` but the artifact is installed as without it under `.m2/repository/rubygems/asciidoctor/2.1.0.alpha.0`. That's why gradele mentions finding `2.0.18.pre-SNAPSHOT`.

I found 3 solutions 
1. Append the -SNAPSHOP tail before installation, [here](https://github.com/asciidoctor/asciidoctorj/blob/2a57067e4ab64cc149d64d6a0c98913f1eba914f/ci/test-asciidoctor-upstream.sh#L23).
2. Append the -SNAPSHOP tail during test, [here](https://github.com/asciidoctor/asciidoctorj/blob/2a57067e4ab64cc149d64d6a0c98913f1eba914f/ci/test-asciidoctor-upstream.sh#L33).
3. Extracting the version from the installed pom and passing it during tests.

I'd say the third the secure one, unless we understand why the snapshot is added.

PS: tested bumping gem-maven-plugin, and it complains about some encoding issue in the pom I could not correct. Some misterious characters :shrug:

abelsromero:
Rebased the fix for upstream tests, we should be seeing :green_circle: now

robertpanzer:
Thank you!

PR #1124, opened 2022-10-12
Replace use of deprecated 'numbered' attribute by 'sectnums'

### LINKED ISSUE #1120: Replace internal use of `numbered` attribute by `sectnums`
`numbered` is considered legacy and while still supported we should probably just change it.

ref: https://github.com/asciidoctor/asciidoctorj/blob/2a57067e4ab64cc149d64d6a0c98913f1eba914f/asciidoctorj-api/src/main/java/org/asciidoctor/Attributes.java#L41

abelsromero:

## Kind of change

- [ ] Bug fix
- [ ] New breaking feature
- [ ] Documentation update
- [ ] Build improvement
- [x] Other :point_right: maintenance chore

## Description

Replace use of deprecated attribute `numbered` by recommended value `sectnums`.

How does it achieve that?

* Replaces constant https://github.com/asciidoctor/asciidoctorj/blob/2a57067e4ab64cc149d64d6a0c98913f1eba914f/asciidoctorj-api/src/main/java/org/asciidoctor/Attributes.java#L41
* Updates tests and JavaDocs when necessary
* No extra tests have been added. There are already valid ones present.

No

Only for users comparing the attibute with the internal value will notice. I personally don't consider that a breaking change since constants should be used to replace manually set values and the internal value should not be checked.
* Theres no impact for users using the constant or AttributesBuilder normally.
* Conversions setting "numbered" manually will still work since is still supported in the Ruby core.

## Issue

Fixes #1120

## Release notes

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -33,6 +33,10 @@ Build Improvement::
 * Upgrade build to Gradle 7.5.1 (#1109)
 * Fix upstream tests forcing SNAPSHOT on Asciidoctor gem installation (#1123) (@abelsromero)
 
+Build / Infrastructure::
+
+* Replace use of deprecated 'numbered' attribute by 'sectnums' (#1123) (@abelsromero)
+
 == 2.5.4 (2022-06-30)
 
 Improvement::
--- a/asciidoctorj-api/src/main/java/org/asciidoctor/Attributes.java
+++ b/asciidoctorj-api/src/main/java/org/asciidoctor/Attributes.java
@@ -38,7 +38,7 @@ public class Attributes {
     public static final String ICONFONT_NAME = "iconfont-name";
     public static final String ICONS_DIR = "iconsdir";
     public static final String DATA_URI = "data-uri";
-    public static final String SECTION_NUMBERS = "numbered";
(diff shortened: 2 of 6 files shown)


robertpanzer:
Thank you!

I wasn't aware that the numbered attribute is deprecated, I found the corresponding comment here:
https://docs.asciidoctor.org/asciidoctor/latest/migrate/asciidoc-py/#updated-and-deprecated-asciidoc-syntax

Since we're changing the value of a constant users should be aware though that they need to recompile their code to use the new value of the constant. In my experience constants that are public static final are usually copied into the class of the referencing site.

abelsromero:
> Since we're changing the value of a constant users should be aware though that they need to recompile their code to use the new value of the constant. In my experience constants that are public static final are usually copied into the class of the referencing site.

I have no issue if this needs to go into a major release. To me it's not technically a breaking change, but it's true the compiler embeds the values and that can bring suprises.

robertpanzer:
I checked that even when setting the attribute `numbered` a converter or extension would see the attribute `sectnums`.
I guess it's indeed safe then to consider it as a minor change and pull it into the v2.5.x branch.
I'll take care of that.

PR #1145, opened 2023-03-19
Expose Document's source and source_lines properties

### LINKED ISSUE #1143: Expose Document::source and Document::source_lines in Document interface
The properties are not exposed and while it can be hacked with `((DocumentImpl)document).getProperty("source")` it'd be nice to have them properly exposed.

https://github.com/asciidoctor/asciidoctor/blob/093eed14225bc9ef4664d03fdc965fa7e2acb7a1/lib/asciidoctor/document.rb#L669-L676

For context, this is necessary to access the result of using [asciidoctor-reducer](https://github.com/asciidoctor/asciidoctor-reducer).

abelsromero:

## Kind of change

- [ ] Bug fix
- [ ] New breaking feature
- [ ] Documentation update
- [ ] Build improvement

## Description

Expose Document's `source` and `source_lines` properties.
This is required to be able to use other asciidoctor components like [asciidoctor-reducer](https://github.com/asciidoctor/asciidoctor-reducer).

How does it achieve that?
Using already present AsciidoctorJ APIs and practices, nothing special.

Not that I know.

No

## Issue

Fixes #1143 

## Release notes

### DIFF
--- a/asciidoctorj-core/src/test/java/org/asciidoctor/jruby/ast/impl/WhenDocumentIsLoaded.java
+++ b/asciidoctorj-core/src/test/java/org/asciidoctor/jruby/ast/impl/WhenDocumentIsLoaded.java
@@ -1,0 +1,119 @@
+package org.asciidoctor.jruby.ast.impl;
+
+import org.asciidoctor.Asciidoctor;
+import org.asciidoctor.Attributes;
+import org.asciidoctor.Options;
+import org.asciidoctor.ast.Document;
+import org.junit.Test;
+
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class WhenDocumentIsLoaded {
+
+    private Asciidoctor asciidoctor = Asciidoctor.Factory.create();
+
+    @Test
+    public void should_return_empty_when_document_is_empty() {
+        assertEmptySources(loadDocument(""));
+    }
+
+    private static void assertEmptySources(Document document) {
+        String source = document.getSource();
+        assertThat(source).isEmpty();
+        List<String> sourceLines = document.getSourceLines();
+        assertThat(sourceLines).isEmpty();
+    }
+
+    @Test
+    public void should_return_source_and_source_lines() {
+        final String asciidoc = asciidocSample();
+
+        Document document = loadDocument(asciidoc);
+
+        String source = document.getSource();
+        assertThat(source).isEqualTo(asciidoc.trim());
+        List<String> sourceLines = document.getSourceLines();
+        assertThat(sourceLines)
+                .containsExactly("= Document Title",
+                        "",
+                        "== Section A",
+                        "",
+                        "Section A paragraph.",
+                        "",
+                        "=== Section A Subsection",
+                        "",
+                        "Section A 'subsection' paragraph.");
+    }
+
+    @Test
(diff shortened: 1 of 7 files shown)


asciidoctorj-core/src/test/java/org/asciidoctor/jruby/ast/impl/WhenDocumentIsLoaded.java
```diff
@@ -0,0 +1,124 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class WhenDocumentIsLoaded {
+
+    private Asciidoctor asciidoctor = Asciidoctor.Factory.create();
+
+    @Test
+    public void should_return_empty_when_document_is_null() {
+        assertEmptySources(loadDocument(null));
```
robertpanzer, line 19: This test seems to fail for me with a NullPointerException.
It seems like this could be fixed by changing `JRubyAsciidoctor.load()` like so:
```java
    @Override
    public Document load(String content, Options options) {
        RubyHash rubyHash = RubyHashUtil.convertMapToRubyHashWithSymbols(rubyRuntime, options.map());
        return (Document) NodeConverter.createASTNode(getAsciidoctorModule().callMethod("load",
                Optional.ofNullable(content).map(rubyRuntime::newString).orElse(null), rubyHash));
    }
```

abelsromero: I don't know how I run it locally, but I'd rather remove the test and deal with supporting `null` consistently in all cases in a separated PR. wdyt?

robertpanzer: Sure, makes sense.

abelsromero: Being purist, if null was an error case, changing it now would be a breaking change.

robertpanzer: Since main is supposed to be the next 3.0.0 it would be allowed to introduce breaking changes :D
But of course we could keep it like this and change the test case.

abelsromero:
Btw, I assume this is fine going to 3.0.0? Or a 2.5.8 is planned?

robertpanzer:
Yes, main would be 3.0.0. 
We can also do a 2.5.8, that contains this change.

asciidoctorj-core/src/test/java/org/asciidoctor/jruby/ast/impl/WhenDocumentIsLoaded.java
abelsromero: Created a story https://github.com/asciidoctor/asciidoctorj/issues/1146 and looked a but into the required changes. If it's OK, I can spend time tomorrow.

abelsromero:
> We can also do a 2.5.8, that contains this change.

If people do not ask for it, I would not do it.

abelsromero:
Changes done, anything else needs improving? :eyes:

robertpanzer:
Thank you! Looks good!

PR #1148, opened 2023-03-21
Accept null as empty content for load and convert

### LINKED ISSUE #1146: Support `null`  as an empty source
`convert` and `load` fail with a java assert exception when input is `null` String.
For satefy we should consider `null` as an empty value that returns an empty document, for both `convert` and `load`.

In the case of convert with `Reader` I'd keep the current exception, at most, add `Objects.requireNonNull(reader, "reader");` to https://github.com/asciidoctor/asciidoctorj/blob/b1902506e70db29bdd2fd5496f1f531e5fbccd13/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/internal/IOUtils.java#L22 so the users gets a hint of where the issue is without adding too much complexity.

```
java.lang.NullPointerException: reader
	at java.base/java.util.Objects.requireNonNull(Objects.java:233)
	at org.asciidoctor.jruby.internal.IOUtils.readFull(IOUtils.java:24)
	at org.asciidoctor.jruby.internal.JRubyAsciidoctor.convert(JRubyAsciidoctor.java:351)
	at org.asciidoctor.jruby.internal.JRubyAsciidoctor.convert(JRubyAsciidoctor.java:358)
	at org.asciidoctor.jruby.ast.impl.SectionImplTest.temporal(SectionImplTest.java:28)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at 

```

abelsromero:
Both empty and null String are accepted as valid for load/convert methods for safety.
* Minor Javadoc fixed
* Integrate WhenDocumentIsLoaded into WhenAsciiDocIsLoadedToDocument

Fixes #1146

## Kind of change

- [ ] Bug fix
- [x] New breaking feature
- [ ] Documentation update
- [ ] Build improvement

## Description

Simply adds some extra consistency and safety to load and convert methods.
There are also some javadoc fixes and test refactors.

How does it achieve that?
Adds additional validations in case of nulls.

We could consider null an invalid value as now and try to fail with a user friendly message.
But semantically null and empty string are the same here imho.

No, I think it should be unnoticed by most users.

## Issue

Fixes #Issue 
Fixes #1146 

## Release notes


robertpanzer:
Thank you!

PR #1153, opened 2023-03-30
Remove invalid destinationDir option from API for v3.0.x

### LINKED ISSUE #853: Destination directory ignored in Asciidoctorj
In Java:

```
Options options = new Options();
options.setDestinationDir("../foo/");
asciidoctor.convertFile(new File("./docs/userguide.adoc"), options);
```

Does not place the output in ../foo. It produces it in-place.

If I add `options.setInPlace(false);` it has no effect.

I'm using Asciidoctorj 2.1.0.

### LINKED ISSUE #941: Deprecation of Options destinationDir
I am confused about the use of `destinationDir` vs `toDir`, I suspect they are the same and maybe the first one could be marked as deprecated to avoid confusion to users.
Seeing this code
https://github.com/asciidoctor/asciidoctor/blob/1d61da9ee0aba63711c415fa629a80f275036aa4/lib/asciidoctor/cli/invoker.rb#L59 and the user manual https://asciidoctor.org/docs/user-manual/#ruby-api-options, seems pretty obvious but I wanted to contrast and maybe check if this has some history reason.

abelsromero:
## Kind of change

- [x] Bug fix
- [x] New breaking feature
- [ ] Documentation update
- [ ] Build improvement

## Description

Fix the issue where using the option `destinationDir` from API had no effect.
The option only is used for the CLI, in that case the value is redirected to 'to_dir'.

How does it achieve that?

1. Completely removes the option from `OptionsBuilder` and `Options` class.
2. Moves the constant with the name from `Options` to `AsciidoctorUtils`: this is the second time I struggle with this class. It's out of place because it contains CLI elements to build the asciidoctor command for logging purposes in the core. So to prevent cyclic dependencies some constants are copied both here and in the `asciidoctor-cli` sub-module -> Solution = create new sub-module "shared" or remove the logging. Third time will be the charm :thinking: 

Initial conversations are about deprecations, but I think it's the moment to do some clean up and breaking changes for v3.0.0.

## Issue

Fixes #853
Fixes #941

:two: in :one: :grin: 

## Release notes

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -35,6 +35,7 @@ Bug Fixes::
 * BREAKING: Fix CLI target file location for source files relative to source dir (#1135) (@AlexCzar)
 * Cell nodes do not inherit from StructuralNode (#1086) (@rahmanusta)
 * Avoid throwing an exception when using AsciidoctorJ CLI and reading input from stdin (#1105) (@AlexCzar)
+* Remove destinationDir Option from API (use toDir instead) (#853, #941) (@abelsromero)
 
 Build Improvement::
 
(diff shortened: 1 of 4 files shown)


robertpanzer:
> 2️⃣ in 1️⃣ 😁

Yep, killed 2 birds with 1 stone :D

PR #1157, opened 2023-04-03
Test Section.isNumbered to ensure expected behaviour

### LINKED ISSUE #1122: Section.isNumbered() returns a boolean while it can also be the symbol :chapter in Asciidoctor
While looking through the changes in #1121 
While looking through the code in Asciidoctor I stumbled over the fact that we have a property Section.numbered that is a boolean.
However, looking at the code in Asciidoctor core, this property can seemingly have the values `true`, `false`, or the symbol `:chapter`: 
https://github.com/asciidoctor/asciidoctor/blob/4bab183b9f3fad538f86071b2106f3b8185d3832/lib/asciidoctor/parser.rb#L1599

Currently, AsciidoctorJ returns true in such case since the symbol seems to coerce to true as the following test indicates.
This behavior is shown by this test:
```java
    @Test
    public void isNumbered() {
        String content = "= Sample Document\n" +
                ":doctype: book\n" +
                ":sectnums: all\n" +
                "\n" +
                "[colophon]\n" +
                "== Test\n" +
                "\n" +
                "Test";

        Document document = asciidoctor.load(content, Options.builder()
                .build());

        List<StructuralNode> blocks = document.getBlocks();
        SectionImpl section = (SectionImpl) blocks.get(0);
        IRubyObject isNumbered = section.getRubyProperty("numbered");
        RubySymbol symbol = JRubyRuntimeContext.get(asciidoctor).newSymbol("chapter");
        assertSame(symbol, isNumbered);
        assertTrue(section.isNumbered());
    }
```

@mojavelinux Can you share some light on the meaning of the value `:chapter`? And should AsciidoctorJ also expose it?

abelsromero:
## Kind of change

- [ ] Bug fix
- [ ] New breaking feature
- [x] Documentation update
- [ ] Build improvement

## Description

Mainly provide some closure to issue #1122.

How does it achieve that?

Test Section.isNumbered to ensure expected behavior
* Core 'numbered' property is not always boolean. In those cases, the boolean can be coerced from non-null value.
See https://github.com/asciidoctor/asciidoctorj/issues/1122
* Minor improvements to RubyObjectWrapper.

We could make further tests and check if the property has the ':chapter' object. But the current implementation already meets the tests, so I see no point in complicating it.
Also, it does not tests doctype `manpage` and `inline` since in those the sections are not taken into consideration or cannot be numbered (manpage).
For reference, I left a comment pointing to the issue in the code.

Includes a couple of minor improvements to `RubyObjectWrapper` but all the tests pass :ok: 

## Issue

Closes #1122

## Release notes

I don't think is worth mentioning.

### DIFF
--- a/asciidoctorj-api/src/main/java/org/asciidoctor/ast/Section.java
+++ b/asciidoctorj-api/src/main/java/org/asciidoctor/ast/Section.java
@@ -74,7 +74,6 @@ public interface Section extends StructuralNode {
      */
     boolean isNumbered();
 
-
     /**
      * Get the section number for the current Section.
      * <p>
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/ast/impl/SectionImpl.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/ast/impl/SectionImpl.java
@@ -67,6 +67,8 @@ public class SectionImpl extends StructuralNodeImpl implements Section {
 
     @Override
     public boolean isNumbered() {
+        // Not always an actual boolean, but coercing when object and non-null is enough:
+        // https://github.com/asciidoctor/asciidoctorj/issues/1122
         return getBoolean("numbered");
     }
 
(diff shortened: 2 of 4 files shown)


robertpanzer:
Thank you!

PR #1159, opened 2023-04-04
(v2.5.x) Remove open access warnings from cli

### LINKED ISSUE #1155: Avoid 'WARN FilenoUtil : Native subprocess control requires open access to the JDK IO subsystem' message
Running the CLI and also from builder [tools](https://github.com/asciidoctor/asciidoctor-maven-plugin/issues/553) displays the following Java warning:

```
023-04-01T22:03:19.247+02:00 [main] WARN FilenoUtil : Native subprocess control requires open access to the JDK IO subsystem
Pass '--add-opens java.base/sun.nio.ch=ALL-UNNAMED --add-opens java.base/java.io=ALL-UNNAMED' to enable.
```

Afaik, there's nothing we can and this comes from jRuby (which [requires access to JDK internals](https://github.com/jruby/jruby/issues/6721)). It's up to calling libraries to add the JVM flags to make explicit that we need access to internals, making the warning dissapear.

For further reference, another interesting conversation I found https://github.com/FasterXML/jackson-databind/issues/3168.

abelsromero:
## Kind of change

- [x] Bug fix
- [ ] New breaking feature
- [ ] Documentation update
- [ ] Build improvement

## Description

Remove `Native subprocess...` warning when running the CLI in Java > 8.
```
023-04-01T22:03:19.247+02:00 [main] WARN FilenoUtil : Native subprocess control requires open access to the JDK IO subsystem
Pass '--add-opens java.base/sun.nio.ch=ALL-UNNAMED --add-opens java.base/java.io=ALL-UNNAMED' to enable.
```

How does it achieve that?

Modifies CLI script to check the Java version running, in case of Java >8 adds required arguments to remove warning.

Leave the warning and document it. However, given the majority of users should be on Java > 8, leaving the warning because we still support it, seems unfair and lazy.

The script changes will add some delay, but it's not appreciable, especially when one takes into account the startup time for the jvm and Ruby scripts.
This does not fix the other warning that appear in newer versions.
```
OpenJDK 64-Bit Server VM warning: Options -Xverify:none and -noverify were deprecated in JDK 13 and will likely be removed in a future release
```
## Issue

Fixes #1155

## Release notes

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -307,6 +307,7 @@ Improvements::
   * Upgrade to Asciidoctor 2.0.8
   * Upgrade to asciidoctor-pdf 1.5.0-alpha.17 (#809)
   * Add Rouge source highlighter to asciidoctor.jar (#806)
+  * Add required `--add-opens` to cli launch script to remove Jdk warnings when running on Java8 (#1155) (@abelsromero)
 
 Bug Fixes::
 
--- a/asciidoctorj-distribution/src/main/resources/customUnixStartScript.txt
+++ b/asciidoctorj-distribution/src/main/resources/customUnixStartScript.txt
@@ -29,6 +29,10 @@ APP_BASE_NAME=`basename "\$0"`
 
 # Add default JVM options here. You can also use JAVA_OPTS and ${optsEnvironmentVar} to pass JVM options to this script.
 DEFAULT_JVM_OPTS=${defaultJvmOpts}
+readonly java_version=\$(java -version 2>&1 | sed -n ';s/.* version "\\(.*\\)\\.\\(.*\\)\\..*".*/\\1\\.\\2/p;')
+if [ "\$java_version" != "1.8" ]; then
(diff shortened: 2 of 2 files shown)


robertpanzer:
I wonder if this really hurts so much that it is worth starting the Java command twice?
Wouldn't it suffice to document that users that want to avoid this warning can run asciidoctorj like this?
```
ASCIIDOCTORJ_OPTS="--add-opens java.base/sun.nio.ch=ALL-UNNAMED --add-opens java.base/java.io=ALL-UNNAMED" asciidoctorj test.adoc
```

abelsromero:
> Wouldn't it suffice to document that users that want to avoid this warning can run asciidoctorj like this?

To me, it's a matter of user experience. Since it doesn't seem [jRuby](https://github.com/jruby/jruby/issues/6721#issuecomment-1424157214) is going to get a fix soon (if it can even be resolved), and we know most users are going to be affected (no one should be in Java8 by now). To me, it doesn't make much sense we force users to a) endure warnings or b) do additional configurations in their machines*.
If we can fix it, even with some bash hacking, let's do it and prevent any possible doubt, especially for new users who might wonder if they are doing something wrong or don't know about/care about Java.

* However, that's going to be my approach for the maven plugin 😢

robertpanzer:
OK, let's go for it :)

PR #1161, opened 2023-04-05
Replace deprecated 'headerFooter' by 'standalone'

### LINKED ISSUE #1160: Replace 'headerFooter' by 'standalone'
I see in the docs that Ruby code :header_hooter` is not the recommended option, but instead should be `:standalone`.

So, as in other cases, I think we should:
- [ ] Replace it in main branch: since we can do breaking changes :fireworks: 
- [ ] Deprecate in v2.5.8

The only other discrepancy with Asciidoctor Ruby I see are:
* `:header_only` (in `parseHeaderOnly` in AsciidoctorJ), not sure about this, should we rename it?
* timings, which is missing and I'd also like to add, the [cli already has some work done](https://github.com/asciidoctor/asciidoctorj/blob/229a0f84205bfed2754e52fd52cdbdb67d584ae1/asciidoctorj-cli/src/main/java/org/asciidoctor/cli/jruby/AsciidoctorInvoker.java#L82-L85). Will create another issue for that.

abelsromero:
## Kind of change

- [ ] Bug fix
- [x] New breaking feature
- [x] Documentation update
- [ ] Build improvement

## Description

Replace deprecated 'headerFooter' by 'standalone'
* Replace headerFooter methods in OptionsBuild & Options
* Update referenes to the old methods
* Added 'standalone' description to docs
Other changes included:
* Re-ordered options in docs to follow alphabetical order
* Remove some unnecessary throws from tests
* Add missing '@deprecated' to templateDir method

How does it achieve that?

Simply rename method and change the use of the `header_footer` option in the internal map by `standalone`.

We could have deprecated the `headerFooter` method but since this is for v3.0.0...breaking changes are ok.
I'll create a PR to deprecate in v2.5.x

No.

## Issue

Fixes #1160

## Release notes

### DIFF
--- a/asciidoctorj-api/src/main/java/org/asciidoctor/Options.java
+++ b/asciidoctorj-api/src/main/java/org/asciidoctor/Options.java
@@ -7,11 +7,15 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+/**
+ * AsciidoctorJ conversion options. Each one maps to an option in Asciidoctor.
+ * See https://docs.asciidoctor.org/asciidoctor/latest/api/options/ for further
+ * details.
+ */
 public class Options {
 
     public static final String IN_PLACE = "in_place";
     public static final String ATTRIBUTES = "attributes";
-    public static final String HEADER_FOOTER = "header_footer";
     public static final String TEMPLATE_DIRS = "template_dirs";
     public static final String TEMPLATE_ENGINE = "template_engine";
     public static final String TO_FILE = "to_file";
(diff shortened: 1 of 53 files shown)


asciidoctorj-api/src/main/java/org/asciidoctor/Options.java
```diff
@@ -7,11 +7,15 @@
...
+/**
+ * AsciidoctorJ conversion options. Each one maps to an option in Asciidoctor.
+ * See https://docs.asciidoctor.org/asciidoctor/latest/api/options/ for further
+ * details.
+ */
 public class Options {
 
     public static final String IN_PLACE = "in_place";
     public static final String ATTRIBUTES = "attributes";
-    public static final String HEADER_FOOTER = "header_footer";
```
robertpanzer, line 14: I wonder if we want to at least keep the constant?
Even if the option is not recommended, Asciidoctor itself still handles it.

robertpanzer:
Thank you!

asciidoctorj-api/src/main/java/org/asciidoctor/Options.java
abelsromero: I thought about it, my rationale was that if this is not the recommended way, we should not make it easy to use. Users can still use it as a String if they want, but since we had no references in the current code we could remove it.

We could add it back, I admit I am being very strict with the "let's clean thing now that we can to breaking changes" :smile:

PR #1162, opened 2023-04-05
(v2.5.x) Deprecate 'headerFooter' in favour of 'standalone'

### LINKED ISSUE #1160: Replace 'headerFooter' by 'standalone'

abelsromero:
## Kind of change

- [ ] Bug fix
- [ ] New breaking feature
- [X] Documentation update
- [ ] Build improvement

## Description

Deprecate 'headerFooter' in favour of 'standalone'
* Replace headerFooter methods in OptionsBuild & Options
* Added 'standalone' description to docs
Other changes included:
* Re-ordered options in docs to follow alphabetical order
* Remove some unnecessary throws from tests
* Add missing '@deprecated' to templateDir method

Related to https://github.com/asciidoctor/asciidoctorj/pull/1161

How does it achieve that?

Simply add new method and change the use of the header_footer option in the internal map by standalone.
The old option is still accepted in the map, hence the check in `AsciidoctorUtils`.

No

No, no breaking change introduced.

## Issue

Fixes #1160

## Release notes

### DIFF
--- a/asciidoctorj-api/src/main/java/org/asciidoctor/Options.java
+++ b/asciidoctorj-api/src/main/java/org/asciidoctor/Options.java
@@ -1,6 +1,5 @@
 package org.asciidoctor;
 
-import java.io.File;
 import java.io.OutputStream;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -74,6 +73,19 @@ public class Options {
     /**
      * Toggle including header and footer into the output.
      *
+     * @param standalone <code>true</code> to generate a standalone output document
+     *                   (which includes the shell around the body content, such
+     *                   as the header and footer).
+     *                   Defaults to <code>true</code> when converting a file only,
+     *                   otherwise is <code>false</code>.
+     */
+    public void setStandalone(boolean standalone) {
+        this.options.put(STANDALONE, standalone);
+    }
+
+    /**
+     * Toggle including header and footer into the output.
+     *
      * @param headerFooter If <code>true</true>, include header and footer into the output,
      *                     otherwise exclude them. This overrides any output-specific defaults.
      *
@@ -85,7 +97,7 @@ public class Options {
     public void setTemplateDirs(String... templateDirs) {
 
         if (!this.options.containsKey(TEMPLATE_DIRS)) {
-            this.options.put(TEMPLATE_DIRS, new ArrayList<Object>());
+            this.options.put(TEMPLATE_DIRS, new ArrayList<>());
         }
 
         List<Object> allTemplateDirs = (List<Object>) this.options.get(TEMPLATE_DIRS);
--- a/asciidoctorj-api/src/main/java/org/asciidoctor/OptionsBuilder.java
+++ b/asciidoctorj-api/src/main/java/org/asciidoctor/OptionsBuilder.java
@@ -60,12 +60,27 @@ public class OptionsBuilder {
     }
 
     /**
-     * Sets header footer attribute.
-     * 
-     * @param headerFooter
-     *            value.
-     * @return this instance.
+     * Toggle including header and footer into the output.
+     *
+     * @param standalone <code>true</code> to generate a standalone output document
+     *                   (which includes the shell around the body content, such
+     *                   as the header and footer).
+     *                   Defaults to <code>true</code> when converting a file only,
+     *                   otherwise is <code>false</code>.
      */
+    public OptionsBuilder standalone(boolean standalone) {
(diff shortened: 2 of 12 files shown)


abelsromero:
Any idea why CI is not running for this branch? :thinking: It's set in the config

robertpanzer:
I have no idea as well unfortunately. 
I wanted to spend some more time digging into this, but recently I barely find any time. 
Do you have any idea?

abelsromero:
> I have no idea as well unfortunately.
> I wanted to spend some more time digging into this, but recently I barely find any time.
> Do you have any idea?

Nop, but should be easy 🤞 I'll check in my fork later.

asciidoctorj-core/src/main/java/org/asciidoctor/jruby/internal/AsciidoctorUtils.java
```diff
@@ -86,7 +86,8 @@ private static List<String> getOptions(Map<String, Object> options) {
             optionsAndAttributes.add(options.get(Options.ERUBY).toString());
         }
 
-        if (options.containsKey(Options.HEADER_FOOTER)) {
```
robertpanzer, line 89: I wonder if we can remove this entire method.
It seems like it's only called to produce some debug log message and in one test. 
But apart from that AsciidoctorUtils.toAsciidoctorCommand() and AsciidoctorUtils.getOptions() does not seem to provide much value.

asciidoctorj-api/src/main/java/org/asciidoctor/Options.java
```diff
@@ -79,7 +91,7 @@ public void setAttributes(Map<String, Object> attributes) {
      *
      */
     public void setHeaderFooter(boolean headerFooter) {
-        this.options.put(HEADER_FOOTER, headerFooter);
+        this.options.put(STANDALONE, headerFooter);
```
robertpanzer, line 94: Is it ok to change this already?
There might for example be postprocessors relying on this option being present when the client code calls .setHeaderFooter().

asciidoctorj-core/src/main/java/org/asciidoctor/jruby/internal/AsciidoctorUtils.java
abelsromero: If you mean `toAsciidoctorCommand` I think the same. Whenever I wanted to reproduce something in asciidoctor I could build the command directly, no need to check this debug. And it's a source of modularity issues.

Do you want to create an issue for that? It helps me to keep track of changes in multiple branches.

asciidoctorj-api/src/main/java/org/asciidoctor/Options.java
abelsromero: You are right, I restored it, thanks!

asciidoctorj-api/src/main/java/org/asciidoctor/OptionsBuilder.java
```diff
@@ -59,15 +59,30 @@ public OptionsBuilder inPlace(boolean inPlace) {
...
-     *            value.
+     *
+     * @param headerFooter value.
      * @return this instance.
+     * @deprecated Use {@link #standalone(boolean)} instead.
      */
+    @Deprecated
     public OptionsBuilder headerFooter(boolean headerFooter) {
-        this.options.setHeaderFooter(headerFooter);
+        this.options.setStandalone(headerFooter);
```
robertpanzer, line 85: Shouldn't we also keep this to `.setHeaderFooter()`?

abelsromero: Indeed

robertpanzer:
Thank you!

PR #1171, opened 2023-04-08
Retrieve all images in catalog during conversion

### LINKED ISSUE #1166: Asset catalog does not include inlined images
First of all, thanks for Asciidoctor and its Java bindings!

I am struggling with using the AST, as I would like to obtain all references to images in my document. I have understood that `findBy` is not able to find any inlined images, however from the issues and discussions around the Asset Catalog I got to understand (probably wrongly) that it would contain also inlined images when processing the source with the appropriate option `catalogAssets` set to `true`.

Is it possible using AsciidoctorJ to obtain the inlined images or do I need to process the individual blocks myself?

Thank you for your help.

abelsromero:
## Kind of change

- [x] Bug fix
- [ ] New breaking feature
- [ ] Documentation update
- [ ] Build improvement

## Description

**DRAFT PR, WIP**

Offer users the option to obtain images from the assets catalog.
**However**, it only works when calling `load()+getContent()`, and unlike in Asciidoctor Ruby, we cannot call `convert` because it fails (see [comment](https://github.com/asciidoctor/asciidoctorj/issues/1166#issuecomment-1500921789)).

How does it achieve that?

Exposes the images properties from the catalog.
However, this PR is not done yet untill:
- [ ] Fix using `convert` to obtain the `Document` instance
- [ ] Fix catalog not being available when using custom converters (see failing test `WhenReadingImagesFromCatalogAssetFromConverter`) :point_right: I think that fixing the first issue should fix this.
- [ ] Docs

WIP

Not

## Issue

Fixes #1166 

## Release notes

### DIFF
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/internal/JRubyAsciidoctor.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/internal/JRubyAsciidoctor.java
@@ -10,7 +10,6 @@ import org.asciidoctor.extension.ExtensionGroup;
 import org.asciidoctor.extension.JavaExtensionRegistry;
 import org.asciidoctor.extension.RubyExtensionRegistry;
 import org.asciidoctor.jruby.AsciidoctorJRuby;
-import org.asciidoctor.jruby.DirectoryWalker;
 import org.asciidoctor.jruby.ast.impl.DocumentHeaderImpl;
 import org.asciidoctor.jruby.ast.impl.NodeConverter;
 import org.asciidoctor.jruby.converter.internal.ConverterRegistryExecutor;
@@ -216,10 +215,6 @@ public class JRubyAsciidoctor implements AsciidoctorJRuby, LogHandler {
         return asciidoctorContent;
     }
 
-    private List<File> scanForAsciiDocFiles(DirectoryWalker directoryWalker) {
-        return directoryWalker.scan();
-    }
-
     @Override
     public void requireLibrary(String... library) {
         requireLibraries(Arrays.asList(library));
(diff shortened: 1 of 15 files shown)


abelsromero:
> Is that what you mean with "it fails"?

Yes, haven't had time to check yet. I assume there's some error in the bindings that should not be very complicated :crossed_fingers:

abelsromero:
I see we explicitly check the return type and just insert null https://github.com/asciidoctor/asciidoctorj/blob/0240f02d54268bc484d6da210d0803c0156578e0/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/internal/JRubyAsciidoctor.java#L310.
And going back into the git history I could not find a real reason for it, seems to come from initial versions and it has been preserved. Could be that we considered returning Document odd when the Java AST was not well defined.
I think the best is to align with Asciidoctor Ruby and we should not take decisions that limit functionality.

I have changed that, and furthermore seeing the code now I see why `convert` may return the dynamic Interface instead of a proper String. I'll open a separated issue for that.

robertpanzer:
I can't recall if this has always been the case, or whether I changed it to this to avoid the problems with the wrappers generated by JRuby.
At least for AsciidoctorJ 3.0 we could check if the expected class is an AST class, and in that case try to apply the NodeConverter.

abelsromero:
> At least for AsciidoctorJ 3.0 we could check if the expected class is an AST class, and in that case try to apply the NodeConverter.

That's my idea, [adaptReturn](https://github.com/abelsromero/asciidoctorj/blob/0395c8a4640c732f0a2b9fc84bbeae8a09f6392d/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/internal/JRubyAsciidoctor.java#L403) takes care of it for now in a simple way. In another PR we can improve and create some fancy Java class [ReturnTypeAdapterVisitorPattern](https://projects.haykranen.nl/java/).
But for now, it gets the job done.

I just updated the changelog, nothing else to add or change from me.

asciidoctorj-core/src/main/java/org/asciidoctor/jruby/internal/JRubyAsciidoctor.java
```diff
@@ -407,6 +400,17 @@ public <T> T convertFile(File file, Map<String, Object> options, Class<T> expect
         }
     }
 
+    private <T> T adaptReturn(IRubyObject object, Class<T> expectedResult) {
+        if (NodeConverter.NodeType.DOCUMENT_CLASS.isInstance(object)) {
+            if (Document.class.isAssignableFrom(expectedResult)) {
+                return (T) new DocumentImpl(object);
```
robertpanzer, line 406: Maybe it would be better to use `NodeConverter.createNode()` if `expectedResult` is Document.class?
Then we would keep NodeConverter as the only class that creates AST nodes.

abelsromero: Absolutely!

abelsromero:
anything else to improve? 👀

robertpanzer:
Sorry for being behind a bit, I'll check it over the weekend.

robertpanzer:
Thanks! Looks good to me.

abelsromero:
Thanks!

PR #1172, opened 2023-04-09
Fix 1158 concurrent conversion to stream

### LINKED ISSUE #1158: Concurrent use of AsciidoctorJ `Asciidoctor` interface results in `ConcurrentModificationException`
I'm testing using the `Asciidoctor` interface in concurrent code. I am able to reproduce this error reasonably consistently (though not every time) simply by starting 100 concurrent threads that all convert asciidoc to PDF. My conversion code looks like this:

```
      asciidoctor.convert(
        adoc,
        Options.builder()
          .backend("pdf")
          .headerFooter(true)
          .attributes(
            Attributes.builder()
              .attribute(
                "pdf-theme",
                "uri:classloader:/my/package/render/pdf/theme.yml"
              )
              .attribute(
                "pdf-fontsdir",
                "uri:classloader:/my/package/render/pdf/fonts;GEM_FONTS_DIR"
              )
              .attribute("imagesdir", "uri:classloader:/my/package/render/pdf/")
              .build()
          )
          .safe(SafeMode.UNSAFE)
          .toStream(outputStream)
          .build()
      )
```

The exception is:

```
java.util.ConcurrentModificationException
	at java.base/java.util.HashMap.computeIfAbsent(HashMap.java:1221)
	at org.jruby.Ruby.addBoundMethod(Ruby.java:2955)
	at org.jruby.anno.TypePopulator$ReflectiveTypePopulator.addBoundMethodsUnlessOmitted(TypePopulator.java:140)
	at org.jruby.anno.TypePopulator$ReflectiveTypePopulator.populate(TypePopulator.java:127)
	at org.jruby.RubyModule.defineAnnotatedMethodsIndividually(RubyModule.java:1296)
	at org.jruby.RubyModule.defineAnnotatedMethods(RubyModule.java:1203)
	at org.asciidoctor.jruby.internal.RubyOutputStreamWrapper.getOrCreateOutputStreamWrapperClass(RubyOutputStreamWrapper.java:63)
	at org.asciidoctor.jruby.internal.RubyOutputStreamWrapper.wrap(RubyOutputStreamWrapper.java:28)
	at org.asciidoctor.jruby.internal.JRubyAsciidoctor.convert(JRubyAsciidoctor.java:304)
	at org.asciidoctor.jruby.internal.JRubyAsciidoctor.convert(JRubyAsciidoctor.java:336)
	at org.asciidoctor.jruby.internal.JRubyAsciidoctor.convert(JRubyAsciidoctor.java:331)
	at my.package.MyClass.generatePdf(MyClass.kt:139)
	at my.package.MyClass$generatePdf$1.invokeSuspend(MyClass.kt)
```

I also see the following warnings:

```
(unknown):0: warning: already initialized constant AsciidoctorJ::OutputStreamWrapper
(unknown):0: warning: already initialized constant AsciidoctorJ::OutputStreamWrapper
(unknown):0: warning: already initialized constant AsciidoctorJ::OutputStreamWrapper
(unknown):0: warning: already initialized constant AsciidoctorJ::OutputStreamWrapper
(unknown):0: warning: already initialized constant AsciidoctorJ::OutputStreamWrapper
(unknown):0: warning: already initialized constant AsciidoctorJ::OutputStreamWrapper
(unknown):0: warning: already initialized constant AsciidoctorJ::OutputStreamWrapper
(unknown):0: warning: already initialized constant AsciidoctorJ::OutputStreamWrapper
(unknown):0: warning: already initialized constant AsciidoctorJ::OutputStreamWrapper
(unknown):0: warning: already initialized constant AsciidoctorJ::OutputStreamWrapper
(unknown):0: warning: already initialized constant AsciidoctorJ::OutputStreamWrapper
(unknown):0: warning: already initialized constant AsciidoctorJ::OutputStreamWrapper
(unknown):0: warning: already initialized constant AsciidoctorJ::OutputStreamWrapper
(unknown):0: warning: already initialized constant AsciidoctorJ::OutputStreamWrapper
(unknown):0: warning: already initialized constant AsciidoctorJ::OutputStreamWrapper
WARNING: while creating new bindings for class org.asciidoctor.jruby.internal.RubyOutputStreamWrapper,
found an existing binding; you may want to run a clean build.
WARNING: while creating new bindings for class org.asciidoctor.jruby.internal.RubyOutputStreamWrapper,
found an existing binding; you may want to run a clean build.
WARNING: while creating new bindings for class org.asciidoctor.jruby.internal.RubyOutputStreamWrapper,
found an existing binding; you may want to run a clean build.
WARNING: while creating new bindings for class org.asciidoctor.jruby.internal.RubyOutputStreamWrapper,
found an existing binding; you may want to run a clean build.
```

robertpanzer:

## Kind of change

- [x] Bug fix
- [ ] New breaking feature
- [ ] Documentation update
- [ ] Build improvement

## Description

Fix the issue reported in #1158.
When converting to a stream a new RubyClass is created lazily.
This step is not protected with a mutex, so that JRuby can fail.
This PR wants to fix this so that this exception does not occur when converting documents in parallel with one Asciidoctor instance.

How does it achieve that?

Eagerly create the class when initializing the Asciidoctor instance.

Yes, the class could be created lazily and a mutex could guarantee mutual exclusion so that the class is created only once.

## Issue

Fixes #1158 on the main branch 

## Release notes

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -40,6 +40,7 @@ Bug Fixes::
 * Cell nodes do not inherit from StructuralNode (#1086) (@rahmanusta)
 * Avoid throwing an exception when using AsciidoctorJ CLI and reading input from stdin (#1105) (@AlexCzar)
 * Remove destinationDir Option from API (use toDir instead) (#853, #941) (@abelsromero)
+* Fix ConcurrentModificationException when converting to stream concurrently (#1158) (@rocketraman)
 
 Build Improvement::
 
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/internal/JRubyAsciidoctor.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/internal/JRubyAsciidoctor.java
@@ -10,7 +10,6 @@ import org.asciidoctor.extension.ExtensionGroup;
 import org.asciidoctor.extension.JavaExtensionRegistry;
 import org.asciidoctor.extension.RubyExtensionRegistry;
 import org.asciidoctor.jruby.AsciidoctorJRuby;
-import org.asciidoctor.jruby.DirectoryWalker;
 import org.asciidoctor.jruby.ast.impl.DocumentHeaderImpl;
 import org.asciidoctor.jruby.ast.impl.NodeConverter;
 import org.asciidoctor.jruby.converter.internal.ConverterRegistryExecutor;
@@ -22,13 +21,30 @@ import org.asciidoctor.jruby.syntaxhighlighter.internal.SyntaxHighlighterRegistr
 import org.asciidoctor.log.LogHandler;
 import org.asciidoctor.log.LogRecord;
 import org.asciidoctor.syntaxhighlighter.SyntaxHighlighterRegistry;
-import org.jruby.*;
+import org.jruby.Ruby;
+import org.jruby.RubyClass;
+import org.jruby.RubyHash;
+import org.jruby.RubyInstanceConfig;
+import org.jruby.RubyModule;
 import org.jruby.exceptions.RaiseException;
 import org.jruby.javasupport.JavaEmbedUtils;
 import org.jruby.runtime.builtin.IRubyObject;
 
-import java.io.*;
-import java.util.*;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.io.Reader;
+import java.io.Writer;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.UUID;
 import java.util.logging.Logger;
 
 public class JRubyAsciidoctor implements AsciidoctorJRuby, LogHandler {
@@ -59,6 +75,7 @@ public class JRubyAsciidoctor implements AsciidoctorJRuby, LogHandler {
 
         this.rubyGemsPreloader = new RubyGemsPreloader(this.rubyRuntime);
         this.logHandlers.add(new JULLogHandler());
+        RubyOutputStreamWrapper.createOutputStreamWrapperClass(this.rubyRuntime);
     }
 
     public static JRubyAsciidoctor create() {
@@ -183,7 +200,6 @@ public class JRubyAsciidoctor implements AsciidoctorJRuby, LogHandler {
         return toDocumentHeader(document);
     }
 
-    @SuppressWarnings("unchecked")
     @Override
     public DocumentHeader readDocumentHeader(String content) {
 
@@ -214,10 +230,6 @@ public class JRubyAsciidoctor implements AsciidoctorJRuby, LogHandler {
             }
         }
         return asciidoctorContent;
-    }
-
-    private List<File> scanForAsciiDocFiles(DirectoryWalker directoryWalker) {
-        return directoryWalker.scan();
     }
 
     @Override
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/internal/RubyOutputStreamWrapper.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/internal/RubyOutputStreamWrapper.java
@@ -22,7 +22,7 @@ public class RubyOutputStreamWrapper extends RubyObject {
 
   private int bytesWritten = 0;
 
-  public static IRubyObject wrap(final Ruby rubyRuntime, final OutputStream out) {
+  static IRubyObject wrap(final Ruby rubyRuntime, final OutputStream out) {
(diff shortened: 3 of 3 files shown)


abelsromero:
> Yes, the class could be created lazily and a mutex could guarantee mutual exclusion so that the class is created only once.

I like the idea of pre-emptive initialization, even when not needed it makes things a lot easier and won't impact conversion performance. But I have a question, ¿are we 100% sure it's not possible to fail during init? I understand not but I am not 100% familiar with all the code. It seems to me that even if someone calls `JRubyAsciidoctor.create` methods directly, each one ends up calling `JRubyAsciidoctor` and each one creates and independent jRuby Runtime.

PS: haven't been able to reproduce the original issue (200 threads), I cannot test the fix, but makes sense based on the original issue logs :disappointed:

robertpanzer:
I could reproduce it locally with 50 threads without this PR on my 10 core Mac.

abelsromero:
> I could reproduce it locally with 50 threads without this PR on my 10 core Mac.

Could this be added as a test to avoid any regression?

Mine doesn't seem to work on Linux, Java17, Ryzen 7 3700X 8-Core(16 threads) :shrug: : Nor converting html or PDF.

```java
package org.asciidoctor.demos;

import org.asciidoctor.Options;
import org.asciidoctor.SafeMode;
import org.asciidoctor.demos.utils.FileUtils;

import java.io.ByteArrayOutputStream;
import java.io.IOException;
import java.nio.file.Files;
import java.util.ArrayList;
import java.util.List;
import java.util.concurrent.*;

public class ConcurrentRunner {

    private static final int N_THREADS = 150;
    private static final int TIMEOUT_SECONDS = 120;

    static final ExecutorService executor = Executors.newFixedThreadPool(N_THREADS);
    static final Asciidoctor asciidoctor = Asciidoctor.Factory.create();

    public static void convert() throws IOException {
        var outputStream = new ByteArrayOutputStream();
        String adoc = Files.readString(FileUtils.file("sample-big.adoc").toPath());

        asciidoctor.convert(
            adoc,
            Options.builder()
                .backend("html5")
                .baseDir(FileUtils.file("asciidoc"))
                .headerFooter(true)
                .safe(SafeMode.UNSAFE)
                .toStream(outputStream)
                .build());

        assert outputStream.size() != 0;
    }

    static class ConvertCallable implements Callable<Boolean> {

        @Override
        public Boolean call() {
            try {
                ConcurrentRunner.convert();
            } catch (Exception e) {
                return Boolean.FALSE;
            }
            return Boolean.TRUE;
        }
    }

    public static void main(String[] args) throws IOException, InterruptedException {
        var runner = new ConcurrentRunner();
        runner.convert();

        final List<Callable<Boolean>> callables = new ArrayList<>();
        for (int i = 0; i < N_THREADS * 5; i++) {
            callables.add(new ConvertCallable());
        }

        List<Future<Boolean>> futures = executor.invokeAll(callables, TIMEOUT_SECONDS, TimeUnit.SECONDS);
        executor.shutdown();
        executor.awaitTermination(TIMEOUT_SECONDS, TimeUnit.SECONDS);

        long count = futures.stream()
            .map(future -> {
                try {
                    return future.get(TIMEOUT_SECONDS, TimeUnit.SECONDS);
                } catch (InterruptedException | ExecutionException | TimeoutException e) {
                    throw new RuntimeException(e);
                }
            })
            .filter(result -> result == Boolean.TRUE)
            .count();

        System.out.println("Successful results: " + count);
    }
}

```

robertpanzer:
> It seems to me that even if someone calls JRubyAsciidoctor.create methods directly, each one ends up calling JRubyAsciidoctor and each one creates and independent jRuby Runtime.

Sorry, did not identify this question while reading initially.
Yes, Every JRubyAsciidoctor instance uses its own Ruby instance, which should have its own classes. For example each Asciidoctor instance also has its own Symbols.

I reproduced it with this test:
```java
public class TestParallelToStream {


    @Test
    public void test() throws InterruptedException {

        List<Thread> threads = new ArrayList<>();
        for (int i = 0; i < 50; i++) {

            threads.add(new Thread(() -> {
                String doc = "= Test\n" +
                        "\n" +
                        "== Test\n" +
                        "\n" +
                        "Test";
                ByteArrayOutputStream out = new ByteArrayOutputStream();
                asciidoctor.convert(doc, Options.builder()
                        .backend("pdf")
                        .toStream(out)
                        .build());
                System.out.println("Converted to " + out.toByteArray().length + " bytes");

            }));
        }
        threads.forEach(Thread::start);
        for (Thread thread : threads) {
            thread.join();
        }
    }
}
```
Maybe the key is using the PDF backend? (For whatever reason.)

abelsromero:
> Maybe the key is using the PDF backend? (For whatever reason.)

Already tried that. But with your version, I reproduce immediately with 3 threads :open_mouth: It must be something about getting the right call the right time.
Nevertheless, thanks a lot, let's merge!

mojavelinux:
The PDF converter loads a lot of dependencies when used, so that makes sense.

PR #1185, opened 2023-04-21
Fix invalid return type when converting to file but expecting 'Document' class

### LINKED ISSUE #1174: 'convert' returns dynamic Ruby Interface instead of String when converting to String
The following example converts an AsciiDoc source to a String, but has 2 problems:

```java
  var options = Options.builder()
      .catalogAssets(true)
      .safe(SafeMode.UNSAFE)
      .build();
  var doc = asciidoctor.convert(Files.readString(sourceDocument.toPath()), options, Document.class);
```

1. The returned internal type is a dynamic coerced interface (in the example `org.jruby.gen.InterfaceImpl682315553`)

image

2. The returned object wraps in fact an String but we asked for `Document.class`.

Both happen because we don't validate the types, but just delegate in jRuby https://github.com/asciidoctor/asciidoctorj/blob/0240f02d54268bc484d6da210d0803c0156578e0/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/internal/JRubyAsciidoctor.java#L314.

I think we should:
* Validate that the expected type and the returned one are compatible. Not sure if we should return null or return exception :thinking: What do ppl think? Or should we just cover known cases and leave the current behavior as lat resort?
* Never return a dynamic JRuby interface.

abelsromero:
## Kind of change

- [x] Bug fix
- [x] New breaking feature
- [ ] Documentation update
- [ ] Build improvement

## Description

Based on previous conversation and [related work](https://github.com/asciidoctor/asciidoctorj/issues/1174) we should not return types that do not match when using `convert(source,options,Class<T>). And I found a case recently.

*How does it achieve that?*

Improves the detection login in `JRubyAsciidoctor::adaptReturn` to correctly treat Document, String and custom Java converter cases.

Other changes:
* WhenConvertingWithExpectedType
* Improved convert method javadoc
* Remove the use of deprecated methods in WhenWritingConverterIsRegistered.groovy
* Moved test 'ObjectConverter' to java to be reusable in Groovy and Java tests

Not that I can think of, but open to suggestions.
But regarding tests, I think we could improve things with Junit5, the new test could use some  `@Nested` and more advanced parametrized tests :disappointed: 

From v2.5.x standpoint is a breaking change. Someone could be casting the returned interface to String to read the content. Now they'll need to actually set `String.class` as the return type.

## Issue

It's related to https://github.com/asciidoctor/asciidoctorj/issues/1174, but I consider the issue a space to discuss more deep changes. For now this PR is just a fix to keep things as similar as they are to the current behavior.

## Release notes

Haven't added anything, I don't think is relevant enough to mention until we close #1174 .


robertpanzer:
Looks good!

Issue #1186, opened 2023-04-22
Migrate to JUnit5


abelsromero:
Trying to bump JUnit to v4 to JUnimake use of newer features I am finding road-blockers.

Normal JUnit4 and some Spock tests can make use of the [vintage compatibility library](https://junit.org/junit5/docs/current/user-guide/#migrating-from-junit4-running), but those with Spock and Arquillian can't.
I see Arquillians hasn't made a proper release in a long time, and the latest Spock integration (`arquillian-spock-core`) is `1.0.0.CR1` from Sept 2019, and is not compatible with newer Spock versions.

Is there any reason no to start rewriting tests to remove Arquillian dependency? I am thinking of re-writing the features as a JUnit5 extension that should work.

I am planning to do this in 2 PRs:
* [x] Set Junit5 as default engine
* [ ] ~Reimplement Arquillian features (Asciidoctor and ClasspathResource injection) as Junit 5 (anything else?), and rewrite tests~ We could keep at the end, see https://github.com/asciidoctor/asciidoctorj/pull/1192

robertpanzer:
Good idea!
I also think that it would be good to "slim" down our dependencies.
Although I hate to say it I would be fine with dropping Arquillian.

The only reason should be the amount of work :)
But if we can migrate gradually then I think the task would be far less intimidating.

abelsromero:
> Although I hate to say it I would be fine with dropping Arquillian.

Me too, it's not only part of the project but of the contributors. But in the end, even if we could keep it using the RC versions, using a JUnit extension means fewer dependencies and a framework more supported. And it's easier for contributors too. I see really no advantages in Arquillian.
So far I have been able to remove Arquillian from some tests adding a few lines to init Asciidoctor and the classpath tools.
That's the part we can refactor to reduce boilerplate, but it's not a big complexity.

> But if we can migrate gradually then I think the task would be far less intimidating.

The current PR aims for that, but there's the issue of Spock compatibility, newer versions have removed `org.spockframework.runtime.Sputnik` which `org.jboss.arquillian.spock.ArquillianSputnik` uses, so I've been forced to remove Arquillian wherever we had `@RunWith(ArquillianSputnik)` :disappointed: .
That's maybe the bigges re-write so far.

PR #1187, opened 2023-04-22
Set Junit5 as default test engine

abelsromero:
## Kind of change

- [ ] Bug fix
- [X] New non-breaking feature
- [ ] New breaking feature
- [ ] Documentation update
- [ ] Build improvement

## Description

Make JUnit5 de default testing framework.
Not possible in Wilddly integration test which requires heavy Arquillian changes and the use of an Arquiliar RC version (https://github.com/arquillian/arquillian-core/issues/137#issuecomment-726596414).

*How does it achieve that?*

* Sets JUnit5 as the default test engine.
* Rewrites polluted tests "category" into a JUnit 5 "tag". The Tag is applied via new annotation `@Polluted` which I personally thing it's nicer that plain using the tag.
* Only minimal changes are done, test that can still work in "junit 4 mode" using `junit-vintage-engine` still use. But at the end only core project requires it, some minimal rewrites allowed to remove it from documentation, cli and distribution modules.
* Removed use of ClasspathResource Test Rule which is not compatible, currently just replaced with use of ClasspathHelper + initializations in each test, I may refactor better when I get some extra time. I want to reduce boilerplate with a junit5 extension in another PR ([see checklist](https://github.com/asciidoctor/asciidoctorj/issues/1186)).
* Restrict codenarc configuration to only required modules

* Other changes
  * Bump codenarc to latest -> Using the `codenarc` configuration did not override the Groovy version in use, I applied it via `constraint`. It can be validated by listing dependencies for codenarc configuration (`./gradlew :module:dependencies`).
  * Bump Groovy to v3.0.x
  * Add documentation reports to GH update upload
  * Fix build and upstream tests reports overriding each other when uploading

Maybe go for "big bang" but I'd rather not do that. We can get something working, merge, and then migrate over time, next way we are removing the use of deprecated methods from tests when we need to review them.

No

## Issue

Related to #1186, we'll see how far we can get to completion in this one PR. :crossed_fingers: 

## Release notes

Added a note to the change just in case someone follows the repo and wants to see what's going on, I will update with the next PR when all is done and we can consider the migration completed.


abelsromero:
Windows failures...

abelsromero:
It seems Windows has some issues cleaning test directories, it's related to https://github.com/junit-team/junit5/issues/2811. Right now I am considering not deleting temp directories :thinking: 
The OS will eventually do that.

abelsromero:
Most of the changes are boilerplate, but there are still a lot. No hurry.

abelsromero:
> P.S.: Sorry for that stupid question

Not at all, in fact imo the location should be improved. Just open an execution detail sand scroll to botom https://github.com/asciidoctor/asciidoctorj/actions/runs/4788558210 to the "Artifacts" section.

We could have a custom job that aggregates them all to have a single zip with some bash, but for now, I am going for a quick one. Downside is yaml duplication.
The env mapping in the config is necessary because you cannot use matrix variables directly :shrug:

PR #1200, opened 2023-05-09
Remove deprecated methods in Options, OptionsBuilder, Attributes & AttributesBuilder

### LINKED ISSUE #1199: Remove deprecated methods from Options and Attributes
New Builder methods have been available for several versions and I think it's time to remove them to simplify code.
The goals are:
* Reduce code
* Stream line the API: no more having multiple ways to create Options or Attributes.
The only exception are the `get() Map` methods maintained for [convinience](https://github.com/asciidoctor/asciidoctorj/issues/1080). 

This does not suppress https://github.com/asciidoctor/asciidoctorj/issues/977, that was a good conversation and there's room to improve, this is just to remove what we already agree to clean code.

abelsromero:
## Kind of change

- [ ] Bug fix
- [x] New breaking feature
- [ ] Documentation update
- [ ] Build improvement

## Description

See https://github.com/asciidoctor/asciidoctorj/issues/1199

*How does it achieve that?*

Simply remove methods already marked as deprecated and fix where necessary (mostly tests).
* Some extra work removing deprecated use (not the method) of `Document::blocks`, using now `Document::getBlocks`

No :thinking: 
Docs feed from test code, so it will be updated. Still pending a manual review though.

It's a  breaking change for v3.0.0, nothing to do for v2.5.x branch.

## Issue

Fixes #1199

### DIFF
--- a/docs/modules/guides/nav.adoc
+++ b/docs/modules/guides/nav.adoc
@@ -1,5 +1,6 @@
 * Help & Guides
 ** Updating to New Releases
+***xref:api-migration-guide-v25x-to-v30.adoc[]
 *** xref:extension-migration-guide-16-to-20.adoc[]
 *** xref:extension-migration-guide-15-to-16.adoc[]
 ** Running in Frameworks
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -36,6 +36,7 @@ Improvement::
 * Expose ImageReferences in the catalog (#1166) (@abelsromero)
 * Return Document AST when using convert or convertFile with appropriate options (#1171) (@abelsromero)
 * Expose Links in the catalog (#1183) (@abelsromero)
+* BREAKING: Remove deprecated methods in Options, OptionsBuilder, Attributes & AttributesBuilder (#1199) (@abelsromero)
 
 Bug Fixes::
 
--- a/asciidoctorj-api/src/main/java/org/asciidoctor/Attributes.java
+++ b/asciidoctorj-api/src/main/java/org/asciidoctor/Attributes.java
@@ -10,13 +10,12 @@ import java.util.Map;
 public class Attributes {
 
     private static final char ATTRIBUTE_SEPARATOR = '=';
+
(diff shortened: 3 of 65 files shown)


abelsromero:
> I'll be off for the next 2 weeks

Enjoy! I need to review the docs and I want to start a "migration guide". We owe to the users to make the migration simple.

abelsromero:
I decided to also remove `Attributes::setAttributes(Map<String, Object>)`. I was NOT marked as deprecated, but it's not consistent with the approach of Java-like builders. 
Users can still use `AttributesBuilder::attribute(String)` `AttributesBuilder::attribute(String, Object)` and even ones `Attributes` instance is created the previous String and Map setters, so no feature is lost for advanced users willing to do more (imho risky) mutable code.

abelsromero:
Ready for review :tada: 
* Squashed changes to keep code and docs changes separated
* Created adjacent issue to track other deprecated methods to remove for v3.0.0. That's all I have in mind that we need to v3.0.0 :thinking: (I don't know how many times I thought that already though)
image

docs/modules/guides/nav.adoc
```diff
@@ -1,5 +1,6 @@
 * Help & Guides
 ** Updating to New Releases
+***xref:api-migration-guide-v25x-to-v30.adoc[]
```
robertpanzer, line 3: There seems to be a blank missing here.
Without that the navigation item "Updating to New Releases" is missing, or has the title "v3.0.x migration guide" instead.

I just realized I did not write anything about the changes to the extensions yet. I'll add that once this is merged.

robertpanzer: To clarify what I mean, the navigation bar looks like this atm:
Bildschirmfoto 2023-05-29 um 11 24 23

robertpanzer (approved):
This is great! Thank you so much.
I can fix the doc issue in a follow up documenting the changes to the extensions.

docs/modules/guides/nav.adoc
abelsromero: Thanks for the fix in the other PR :bow:

PR #1223, opened 2023-06-18
(v2.5.x) Fix antora metadata

abelsromero:
## Kind of change

- [ ] Bug fix
- [ ] New breaking feature
- [x] Documentation update
- [ ] Build improvement

## Description

Update Antora metadata for branch v2.5.x

*How does it achieve that?*
Simply update `antora.yml` file.

No

No

## Issue

*If this PR fixes an open issue, please add a line of the form:*

No issue.

## Release notes

*Please add a corresponding entry to the file CHANGELOG.adoc.*
I don't think it makes sense, this metadata will be updated in the docs as soon as they are updated (nightly iirc).
It's not something that depends on a release.

### DIFF
--- a/docs/antora.yml
+++ b/docs/antora.yml
@@ -3,16 +3,16 @@ title: AsciidoctorJ
 version: '2.5'
 asciidoc:
   attributes:
-    release-version: 2.5.7
-    artifact-version: 2.5.7
+    release-version: 2.5.10
+    artifact-version: 2.5.10
     asciidoctorj-epub3-version: 1.5.1
     asciidoctorj-pdf-version: 1.6.2
-    jruby-version: 9.3.8.0
-    url-jruby: http://jruby.org
+    jruby-version: 9.4.2.0
+    url-jruby: https://www.jruby.org/
     url-gradle: https://gradle.org
     url-repo: https://github.com/asciidoctor/asciidoctorj
     url-tilt: https://github.com/rtomayko/tilt
-    url-font-awesome: http://fortawesome.github.io/Font-Awesome
+    url-font-awesome: https://fontawesome.com
     url-highlightjs: https://highlightjs.org
     url-prismjs: https://prismjs.com
 nav:


docs/antora.yml
```diff
@@ -3,16 +3,16 @@ title: AsciidoctorJ
 version: '2.5'
 asciidoc:
   attributes:
-    release-version: 2.5.7
-    artifact-version: 2.5.7
+    release-version: 2.5.10
+    artifact-version: 2.5.10
     asciidoctorj-epub3-version: 1.5.1
-    asciidoctorj-pdf-version: 1.6.2
+    asciidoctorj-pdf-version: 2.3.6
```
robertpanzer, line 9: We currently don't include asciidoctorj-pdf 2.x.x in the asciidoctorj 2.5.x distribution since this should be a major breaking change (at least according to semver rules).
Should we instead keep the last 1.x version here and reference 2.3.6 in the docs for asciidoctorj 3.x?

abelsromero: This is more of a compatibility matrix to me, and since I have been using them in the [maven-examples](https://github.com/asciidoctor/asciidoctor-maven-examples/blob/61735bd07dc52d883f509c960c79922d1017a018/asciidoctor-pdf-example/pom.xml#L16C9-L18) without issues I assumed the latest of both "j" and "j-pdf" were fine together.
I personally don't see a problem with that. If anything, we should add an/multiple pdf integration test to make sure the versions are compatible.

abelsromero: > pdf integration test

If we decide on that, I am definitely going to ask for a dependant to manage the version. We'll get an automatic PC that will test the compatibility as soon as j-pdf is out.

robertpanzer: Then we should likely also add asciidoctorj-pdf 2.3.6 to the distribution 2.5.x, right?
I am a bit hesitant to pull in a major version change in a bugfix release.

robertpanzer: Generally, testing the compatibility can be quite difficult, to me it means more than the API compiles, and writing integration tests that cover every possible use case can be challenging.
I'd like to use whatever the dependency claims through semantic versioning, if it claims to have a major breaking change, we should also only expose it in a major release.
I think that is also the way that other libraries like Spring Boot follow.

PR #1243, opened 2023-11-20
Fixes #1230. Add test to make sure Asciidoctor version is returned wh…

### LINKED ISSUE #1230: Ensure AsciidoctorJ version is returned always in CLI
I see this null check https://github.com/asciidoctor/asciidoctorj/blob/6088e5b143646c8cd291542d9f95b55cbaaec0fb/asciidoctorj-cli/src/main/java/org/asciidoctor/cli/jruby/AsciidoctorInvoker.java#L93 and I wonder if we should remove it and ensure the version is always obtainable.

We could keep the check for resiliency, but log a message indicating something went wrong.

robertpanzer:
…en using --version

## Kind of change

- [x] Bug fix
- [ ] New breaking feature
- [ ] Documentation update
- [ ] Build improvement

## Description

This PR fixes #1230.
#1230 mentioned that we have a fallback when returning the version of the packaged Asciidoctor version.
This PR removes that fallback and would fail hard with a NullPointerException if the generated file containing the version numbers is not available.
Additionally a test makes sure that the generated file is available.

How does it achieve that?

Remove the fallback, and enhance the tests.

We could search the class path for the asciidoctor gemspec and parse that, or maybe call some API exposed by JRuby to query the gems (or a Ruby-native API).

No, everything should work as before.

## Issue

Fixes #1230 

## Release notes

### DIFF
--- a/asciidoctorj-cli/src/test/java/org/asciidoctor/cli/WhenAsciidoctorIsCalledUsingCli.java
+++ b/asciidoctorj-cli/src/test/java/org/asciidoctor/cli/WhenAsciidoctorIsCalledUsingCli.java
@@ -283,9 +283,14 @@ public class WhenAsciidoctorIsCalledUsingCli {
             System.setOut(sysout);
             new AsciidoctorInvoker().invoke("--version");
             String result = out.toString(StandardCharsets.UTF_8);
+            String expected = System.getProperty("org.asciidoctor.asciidoctorj-test.version.asciidoctorj");
+            Assertions.assertThat(expected)
+                    .as("Bad test setup, could not determine expected Asciidoctor version")
+                    .isNotEmpty();
+            expected = expected.replace("-SNAPSHOT", "");
             Assertions.assertThat(result)
                     // Needs to be updated when version changes
-                    .contains("2.0.20");
+                    .contains(expected);
         } finally {
             System.setOut(previousSystemOut);
         }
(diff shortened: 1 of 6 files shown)


robertpanzer:
The upstream build obviously won't work with that.
I'll try to figure out a way to adapt the test accordingly.
Maybe I can make the version available to the test via a system property.

asciidoctorj-cli/src/test/java/org/asciidoctor/cli/WhenAsciidoctorIsCalledUsingCli.java
```diff
@@ -274,6 +275,23 @@ void should_convert_to_subdirectories(@ClasspathResource("relative/sub/test.adoc
...
+    @Test
+    void should_print_version() throws IOException {
+        PrintStream previousSystemOut = System.out;
+        try (ByteArrayOutputStream out = new ByteArrayOutputStream();
+            PrintStream sysout = new PrintStream(out)){
+            System.setOut(sysout);
+            new AsciidoctorInvoker().invoke("--version");
+            String result = out.toString(StandardCharsets.UTF_8);
+            Assertions.assertThat(result)
+                    // Needs to be updated when version changes
```
abelsromero, line 287: Upstream tests fail on line 288, probabley related to the hardcoded assert. Maybe validate simply with a regex or read `asciidoctorj-version.properties` from classpath to assert it's the same?

PS1: Btw, since `createVersionFile` is run during `jar` I don't understand what file is being read during tests but tests work 🤔

abelsromero (changes requested):

abelsromero (approved):

PR #1244, opened 2023-11-20
CLI should set :mkdirs option by default (#1241)

### LINKED ISSUE #1241: CLI should set :mkdirs option by default

robertpanzer:

## Kind of change

- [x] Bug fix
- [ ] New breaking feature
- [ ] Documentation update
- [ ] Build improvement

## Description

Make the CLI set the :mkdirs option.

How does it achieve that?

Similar to the implementation of Asciidoctor it sets the option :mkdirs if the input is not read from stdin.

## Issue

Fixes #1241

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -17,6 +17,10 @@ Improvement::
 
 * Upgrade to JRuby 9.4.3.0 (#1235) (@headius)
 
+Bug Fixes::
+
+* CLI should set :mkdirs option by default (#1241) (@mojavelinux)
+
 == 2.5.10 (2023-06-04)
 
 Improvement::
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/cli/AsciidoctorInvoker.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/cli/AsciidoctorInvoker.java
@@ -160,6 +160,8 @@ public class AsciidoctorInvoker {
             return;
         }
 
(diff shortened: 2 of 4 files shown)


robertpanzer:
@mojavelinux This PR should be a minimal solution for adding the mkdirs option as reported in #1241 
It would generate all results directly into the target directory though, so it would at most create that target directory, but not any structure below that which would reflect the path of the documents relative to the source directory set with `-R`.

mojavelinux:
That's fine because this is orthogonal to supporting `**/*.adoc`. Most uses of the CLI are to convert a single document to a specified output file or directory. That is what is needed at a minimum in order for AsciidoctorJ to be used with the Antora PDF extension.

mojavelinux:
:tada:

PR #1250, opened 2023-12-09
Moved context functions to ContentNode

RayOffiah:

## Kind of change

- [ ] Bug fix
- [ ] New breaking feature
- [ ] Documentation update
- [ ] Build improvement

## Description

Adding a function to set the Context of a node.

How does it achieve that?
Added a setContext function to the ContentNode

Possibly.

Not that I know of.

## Issue

Fixes #Issue 

## Release notes

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -35,6 +35,8 @@ Bug Fixes::
 
 Improvement::
 
+* Add `setContext` function to StructuralNode.
+
 * Add command line option --failure-level to force non-zero exit code from AsciidoctorJ CLI if specified logging level is reached. (#1114)
 * Upgrade to asciidoctorj 2.0.20 (#1208)
 * Upgrade to asciidoctorj-pdf 2.3.7 (#1182)
--- a/asciidoctorj-api/src/main/java/org/asciidoctor/ast/ContentNode.java
+++ b/asciidoctorj-api/src/main/java/org/asciidoctor/ast/ContentNode.java
@@ -18,6 +18,8 @@ public interface ContentNode {
 
     String getContext();
 
+    void setContext(String context);
+
     Document getDocument();
 
     boolean isInline();
--- a/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/ast/impl/ContentNodeImpl.java
+++ b/asciidoctorj-core/src/main/java/org/asciidoctor/jruby/ast/impl/ContentNodeImpl.java
@@ -34,6 +34,11 @@ public abstract class ContentNodeImpl extends RubyObjectWrapper implements Conte
         return getString("context");
     }
 
+    @Override
+    public void setContext(String context) {
+        setString("context", context);
+
+    }
 
     @Override
     public ContentNode getParent() {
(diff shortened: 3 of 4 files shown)


abelsromero:
I imagine this this related to this conversation https://asciidoctor.zulipchat.com/#narrow/stream/279644-users.2Fasciidoctorj/topic/Setting.20block.20context.20in.20AsciidoctorJ/near/406515443.

I have 2 comments:
1. At first glance, it's weird to move the methods to `StructuralNode` because it is under `ContentNode`. If we consider the property to be "writable", why not directly in `ContentNode`?

2. It'd be great to better understand the use case if you could add a test with the use case you mentioned in Zulip?  I do wonder if a new node could just be built some other way?

asciidoctorj-api/src/main/java/org/asciidoctor/ast/ContentNode.java
```diff
@@ -16,7 +16,7 @@ public interface ContentNode {
 
     ContentNode getParent();
 
-    String getContext();
```
robertpanzer, line 19: If I am not missing anything this would also remove access to the context of a PhraseNode.
But a PhraseNode even needs to get a context given when created by a Processor.

RayOffiah: Yes, I think I was wrong with my first assessment. It's fine to just add the function to the ContentNode.

asciidoctorj-core/src/main/java/org/asciidoctor/jruby/ast/impl/ContentNodeImpl.java
```diff
@@ -34,6 +34,8 @@ public String getContext() {
         return getString("context");
     }
 
+    @Override
+    public void setContext(String context) { setString("context", context); }
```
robertpanzer, line 38: Could you please format this in the same way as the rest of the file?
We usually have line breaks after curly braces.

robertpanzer (commented):
Would you mind adding a test case to make sure that the method really works?

asciidoctorj-core/src/main/java/org/asciidoctor/jruby/ast/impl/ContentNodeImpl.java
RayOffiah: Yep, can do that.

CHANGELOG.adoc
```diff
@@ -16,7 +16,7 @@ For a detailed view of what has changed, refer to the {url-repo}/commits/main[co
 === Breaking changes
 
 Improvement::
-
+* Add `setContext` function to StructuralNode. Move `getContext`
```
robertpanzer, line 19: Your PR doesn't seem to be moving getContext() anymore.
Additionally, I don't think it is a breaking change.
Could you please fix this in the changelog?

RayOffiah: Yup.

robertpanzer (approved):
LGTM

robertpanzer:
Thanks!
Feel free to also open a PR against the v2.5.x branch if you want to have this in a minor release before 3.0.0 is released.

RayOffiah:
Yes, I can do that. Might have to wait until later in the week though.

Sent from Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #1270, opened 2024-03-16
Fixes #1265. Support setting column widths

### LINKED ISSUE #1265: `Column#setWidth` is ignored
When using `Column#setWidth` on the AST API, this is completely ignored.

And not only when modifying an existing table that was originally constructed from parsed adoc, but also when you build a new table from ground up, for example in a tree processor.

I expected when I set the width of the first column to 1 and the second to 1 too, that it behaves like if I would have that in an adoc table columnspec.

But actually, not matter what you set the width to via AST API, the columns (just tried with two columns) are always treated like first column set to 1 and second set to 0.
To get what you want currently, you instead have to set the attribute `colpcwidth` manually like you want it.

I would expect using `setWidth` to be effective properly, at least when constructing the table from ground up.

> mojavelinux:
> It might be worth mentioning in the issue the reason the value needs to be mapped. It's because width is an evaluated property and the API needs to perform that work since the assignment comes after the evaluation that's done internally by the parser. In other words, setting width alone has no impact.

So if the parser is doing also other calculations, it might be the same issue.

robertpanzer:

## Kind of change

- [x] Bug fix
- [ ] New breaking feature
- [ ] Documentation update
- [ ] Build improvement

## Description

Add support for setting the width of table columns.
While columns have a method `setWidth()` that sets the attribute `width` of a column, this does not become effective.
In addition it is required to set the attribute `colpcwidth`, which Asciidoctor core only sets internally when calling Table#assign_column_widths.

How does it achieve that?

This PR adds a method Table.assignColumnWidths which calls the Ruby method mentioned above.

Very likely.

There should be no change of existing behavior.

## Issue

Fixes #1265

This PR is a draft until more tests, in particular for the auto width_cols are added.

## Release notes

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -32,6 +32,7 @@ Improvement::
 Bug Fixes::
 
 * -s CLI option should be changed to -e to align with Asciidoctor (#1237) (@mojavelinux)
+* Column#setWidth is ignored (#1265) (@Vampire)
 
 === Compatible changes
 
--- a/asciidoctorj-api/src/main/java/org/asciidoctor/ast/Table.java
+++ b/asciidoctorj-api/src/main/java/org/asciidoctor/ast/Table.java
@@ -57,4 +57,5 @@ public interface Table extends StructuralNode {
      */
     void setGrid(String grid);
 
+    void assignColumnWidths();
 }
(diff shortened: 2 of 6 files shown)


robertpanzer:
I added a test that uses autowidth columns, where the width attribute has a negative value.
Together with that I was considering adding an overloaded implementation for assignColumnWidths that allows to pass the widthBase.
Testing this I realized that I do not understand properly how this works, since the assigned widths were just the absolute widths set on the columns.
I.e., if I set the widths to 2, 3 and -1 for 3 columns, even with a width base of 10 the widths in the resulting html were 2, 3 and not set.
Therefore, I kept it like it is now, with only one implementation for assignColumnWidths.

robertpanzer:
I can backport this to 2.5.x.

PR #1293, opened 2024-12-28
Upgrade Gradle, move to org.ysb33r.jruby plugins

### LINKED ISSUE #1284: Remove dependency on jruby-gradle plugin
It seems like the jruby-gradle plugin is no longer maintained.
It doesn't work with the latest version of Gradle, it doesn't resolve anymore with the default gradle plugin syntax, and there hasn't been any activity on the repository for more than a year.
Corresponding issues in https://github.com/jruby-gradle/jruby-gradle-plugin/issues/452 and https://github.com/jruby-gradle/jruby-gradle-plugin/issues/451 are also unanswered.

I think the only functionality that we use today from this plugin is the option to install Gems.
It should be possible to invoke JRuby ourselves to install Gems in our build script.

I think the only dependencies that AsciidoctorJ should have is JRuby and of course the ruby gems.

robertpanzer:
## Kind of change

- [ ] Bug fix
- [ ] New breaking feature
- [ ] Documentation update
- [x] Build improvement

## Description

Remove the use of the jruby gradle plugin. This seems to be unsupported and currently prevents running the build with newer versions of Java.

How does it achieve that?

Migrate to the new plugin suite for managing ruby gems proposed by @ysb33r 

Yes, the build could use the Maven plugin. It is better though to stay within the gradle ecosystem.

## Issue

Fixes #1284


build.gradle
```diff
@@ -18,16 +18,14 @@ buildscript {
...
-          exclude module: 'okhttp-digest'
-        }
-        classpath 'org.ysb33r.gradle:grolifant:0.17.0'
     }
 }
 
 // modern plugins config
 plugins {
+  id 'org.ysb33r.jruby.resolver'    version '2.0.0-alpha.4'
+  id 'org.ysb33r.jruby.base' version '2.0.0-alpha.4'
```
ysb33r, line 27: The final the `2.0.0` was released today.

robertpanzer: Thanks! Updated to 2.0.0 and the tests seem to pass.

